PR #22, opened 2020-03-17
Address x rotation for GLTF

### LINKED ISSUE #12: B3DM : Apply 90 degree GLTF fix about X
Inline:
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/ThreeTilesRenderer.js#L317

Spec reference:
https://github.com/CesiumGS/3d-tiles/tree/master/specification#y-up-to-z-up

gkjohnson:
Fix #12 

Apply a 90 degree rotation about X to accommodate +Y -> +Z conversion.

Frustum checks seem to be incorrect which this exposed more apparently, which should be fixed.

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -37,7 +37,7 @@ let params = {
 	'maxDepth': 15,
 	'loadSiblings': true,
 
-	'up': '+Y',
+	'up': '+Z',
 	'displayBounds': false,
 	'showThirdPerson': true,
 	'reload': reinstantiateTiles,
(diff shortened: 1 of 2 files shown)


gkjohnson:
This seems really ill documented and there's no example data to show how the behavior is supposed to work because applying the rotation causes issues in both data sets available. There also seems to be a `gltfUpAxis` field in the asset object of tileset.json that's no longer in the spec?

PR #61, opened 2020-05-18
Add loaders for other tiles formats

gkjohnson:
A quick effort to get familiar with the other tiles file formats.

What needs to happen:

- Understand the processing of feature and batch tables, which are used in PNT and I3DM files. References to Cesium [Feature Table Source](https://github.com/CesiumGS/cesium/blob/master/Source/Scene/Cesium3DTileFeatureTable.js) and [Batch Table Source](https://github.com/CesiumGS/cesium/blob/master/Source/Scene/BatchTable.js).

- Share feature and batch table implementations across loaders.

- Test against [example tileset types](https://github.com/CesiumGS/3d-tiles-samples/tree/master/tilesets).

- Implement support in TilesRenderer parseTile.

- Implement support for debug colors in DebugTilesRenderer.


gkjohnson:
More information on feature and batch tables laid out in #63.

Issue #78, opened 2020-06-22
won't load properly on 3DTilesRendererJS


yoyomule:
Hello, I converted the bim model to 3dtile, which can be loaded normally in cesium. But it won't load properly on 3DTilesRendererJS. The error is as follows pic, it seems to be missing a file with root.b3dm
![image](https://user-images.githubusercontent.com/10000818/85267344-bc4e2f00-b4a7-11ea-81ed-78b6885d0bd6.png)
![image](https://user-images.githubusercontent.com/10000818/85267364-c1ab7980-b4a7-11ea-8ba6-cd141703c4f4.png)
 I provided the test file in the attachment.
[bim3dtile.zip](https://github.com/NASA-AMMOS/3DTilesRendererJS/files/4811490/bim3dtile.zip)

gkjohnson:
Hi @yoyomule! Thanks for the report.

I downloaded the zip you provided and I'm able to load the tileset without errors. Can you provide more information on your set up? Would you also be able to provide a screenshot of what you expect it to look like?

Please also keep in mind that there are still some unimplemented features such as CMPT file format that can't currently be loaded in this tileset, as well. See #15 for more details.

gkjohnson:
Ping @yoyomule. Is this still an issue? I'm not seeing the problems you've reported.

yoyomule:
Sorry, this question is here
https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/78#issuecomment-647668347

Garrett Johnson <notifications@github.com> 于2020年6月30日周二 下午12:11写道：

gkjohnson:
> Sorry, this question is here

Yes and I've responded here: https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/78#issuecomment-647668347.

I cannot reproduce your problem.

yoyomule:
Hello, we loaded it, but it is not displayed correctly. I provided two pictures, one is loaded in cesium, and the other is loaded in 3DTilesRendererJS. As you can see, the situation in 3DTilesRendererJS is wrong, and related error messages are also prompted. 
![image](https://user-images.githubusercontent.com/10000818/86088606-3ea6a680-bad9-11ea-8641-0c3bfddc4afa.png)
![image](https://user-images.githubusercontent.com/10000818/86088614-41a19700-bad9-11ea-96b8-94cbf8883f37.png)
[...]

PR #104, opened 2020-07-20
Add Draco, DDS, and KTX2 Loader support

### LINKED ISSUE #28: Add Draco Compression Support
Should this be baked into the loader? It's a big file to require people to include. Besides how does loading the draco assembly work with general build processes?

gkjohnson:
Fix #28 

**TODO**

- [x] Add loaders to CMPTLoader
- [x] Add loaders to I3DMLoader
- [x] Add loaders to TilesRenderer
- [ ] Linting
- [x] d.ts updates
- [ ] Test using data from #78 

NOTE: It may be simpler to just have the i3dm and b3dm loaders resolve a gltf loader from a manager using something like a ".gltf" path which would allow for creation and use of a custom GLTF loader which could include draco etc loaders. Otherwise there's lots of redundant information here.


gkjohnson:
Closing in favor of #105

Issue #110, opened 2020-08-07
b3dm batch attributes are not available in three.js


liberostelios:
While B3DMLoaderBase parses the batch attributes of a `b3dm` tile, there is no way to access those data from the 3D scene (e.g., by using a raycaster and trying to find the attributes of a triangle).

I think that could be easily solved by assigning those information to the scene object created by the three.js `TilesRenderer`. I did something like this [here](https://github.com/liberostelios/3DTilesRendererJS/commit/b3e55e2b3acd5f0a6276042444d4e024fe29613b), so then I am able to pick those information when necessary ([here](https://github.com/liberostelios/3DTilesRendererJS/blob/06600aec2424a3f5ddb464f2b3bdafae85e1f78e/example/viewer.js#L336-L348) is an example how). Not sure if that's the most elegant way to solve this, but it worked for me.

gkjohnson:
I think this is a good addition. I'm torn between putting it in the loaders themselves and putting it into the 3DTilesRenderer. Why don't we set `batchTable` and `featureTable` on the scene itself in `B3DMLoader`. The same should be done for `I3DMLoader` and `PNTSLoader`.

Would you like to make a PR?

Thanks for contributing back improvements!

liberostelios:
Sure, I will make a PR! But I'll have to clean up the code first, because I've hacked a lot of things and this is still based on an older version of the code.

gkjohnson:
Fixed in #113

PR #113, opened 2020-08-14
Export batch attributes in three.js

liberostelios:
This should allow batch attributes to be exposed in three.js, as discussed in #110.

### DIFF
--- a/src/three/B3DMLoader.js
+++ b/src/three/B3DMLoader.js
@@ -23,6 +23,10 @@ export class B3DMLoader extends B3DMLoaderBase {
 
 				model.batchTable = b3dm.batchTable;
 				model.featureTable = b3dm.featureTable;
+
+				model.scene.batchTable = b3dm.batchTable;
+				model.scene.featureTable = b3dm.featureTable;
+
 				resolve( model );
 
 			}, reject );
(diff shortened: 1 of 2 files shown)


liberostelios:
Haven't tested this, tbh. But it should work. Will have to write a test.

There are also other ways to do this, as you mentioned in #110, probably as part of the `B3DMLoader`. And maybe there could be an interface for querying the information somehow, instead of having the user to do a lot of manual work.

gkjohnson:
Thanks! I do feel this should go in the `B3DMLoader` class, though, so that it's available when a b3dm is loaded through the [CMPTLoader](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/992d5397239504bff38df897b7e098d165fa3694/src/three/CMPTLoader.js#L31-L41), as well. We should also expose the batch table itself because there are convenience functions for retrieving data for cases when the batch info isn't just an array:

```js
model.batchTable = result.batchTable;

// ...

model.batchTable.getData( 'identificatie' )[ batch_id ];
```

> Haven't tested this, tbh. But it should work.

This is simple enough that I'm confident. I think at another the [b3dm loader example](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/b3dmExample.html) can be updated to include reading batch table data to serve as a test.

gkjohnson:
Excellent, thank you! Both the the I3DMLoader and PNTSLoader provide feature and batch tables too, I think, if you want to expose those in the same way, as well.

Issue #118, opened 2020-09-25
"Render on change" example enhanchement


liberostelios:
The current [example](https://github.com/NASA-AMMOS/3DTilesRendererJS#render-on-change) for not using a loop works, but it won't start loading the tiles unless something forces a second `tilesRenderer.update();` to be called.

This can be improved by making `needsRerender` an integer with the number of times to re-render and starting with 2. So the example would be:

```JavaScript
let needsRerender = 2;
const tilesRenderer = new TilesRenderer( './path/to/tileset.json' );
tilesRenderer.onLoadTileSet = () => needsRerender = 1;
tilesRenderer.onLoadModel = () => needsRerender = 1;

function renderLoop() {

	requestAnimationFrame( renderLoop );
	if ( needsRerender > 0) {
	
		needsRerender--;
		tilesRenderer.update();
		renderer.render( camera, scene );
	
	}

}
renderLoop();
```

There is also a typo on `upadate()`.

gkjohnson:
> There is also a typo on `upadate()`.

Heh, thanks :) fixed!

> The current example for not using a loop works, but it won't start loading the tiles unless something forces a second `tilesRenderer.update();` to be called.

Hmm I thought that's what #112 addressed? If this is true then I think it's a bug. Have you tracked down why? Is `onLoadTileSet` being fired? Here's what should be happening:

1. Initial update is called which  triggers load of tileset.json
1. tileset.json finishes loading which fires `onLoadTileSet` setting the render flag
1. update is called again with triggers load of tile geometry.
1. tile geometry finishes loading which fires `onLoadModel` setting the render flag
1. repeat

liberostelios:
>     2. tileset.json finishes loading which fires `onLoadTileSet` setting the render flag
> 
>     3. update is called again with triggers load of tile geometry.
> 
>     4. tile geometry finishes loading which fires `onLoadModel` setting the render flag

Actually, the example above was incorrect. This is the part that breaks. So the right code would be:

```JavaScript
let needsRerender = 1;
const tilesRenderer = new TilesRenderer( './path/to/tileset.json' );
tilesRenderer.onLoadTileSet = () => needsRerender = 2;
tilesRenderer.onLoadModel = () => needsRerender = 1;

function renderLoop() {

	requestAnimationFrame( renderLoop );
	if ( needsRerender > 0) {
	
		needsRerender--;
		tilesRenderer.update();
		renderer.render( camera, scene );
	
	}

[...]

PR #119, opened 2020-09-28
Add option to show empty tiles

### LINKED ISSUE #111: Should the error be calculated for empty tiles?
Currently, there is a check for whether a tile has content before the error calculation is done. I guess that works in case where tiles are used to represent different LoDs.

In our case, as discussed in #77, we decided to use empty tiles to allow for a "flatter" area to be hidden when the error was sufficient enough. That required some change in the renderer (see [here](https://github.com/liberostelios/3DTilesRendererJS/commit/a4c0065ec137529a859ae7aacafdb966c753ea5e#diff-25d902c24283ab8cfbac54dfa101ad31)).

I am not sure if there are any negative side-effects for normal case, but it works good enough for us. We built a tileset where, basically, we have one layer of empty tiles with an error and each of that has as child its respective tile with the content. That way, the renderer "picks" the empty tile, if it's far enough, or the content tile, if it's close.

Would that benefit others? What's the rationale behind not computing the error of empty tiles? Just for performance gains? I believe maybe there could be an option to switch between computing the error for empty tiles or not, so that it can serve any case. I would be willing to implement that and make a PR, if you will.

liberostelios:
This is, basically, to resolve #111. It allows for the calculation of errors on empty tiles to occur, which then can "show" empty tiles.

This is useful for a use case with a more GIS-oriented tilesets. In that case, the tileset can have a complicated structure just so that it optimises which tiles are shown, but the content are basically just a simple flat tileset (like a WMTS, for instance).


liberostelios:
As mentioned in #118, this causes a side effect where after the tileset is downloaded the content start showing up only after the second `tiles.update()` function is called.

src/base/TilesRendererBase.js
```diff
@@ -70,6 +70,7 @@ export class TilesRendererBase {
 		this.loadSiblings = true;
 		this.displayActiveTiles = false;
 		this.maxDepth = Infinity;
+		this.showEmptyTiles = false;
```
gkjohnson, line 73: Lets default this to `true` because it's the right way according to the spec. I may want to remove the option at some point but we can leave it for now.

gkjohnson:
Just a note for myself but the [recursivelyMarkUsed](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/base/traverseFunctions.js#L36-L53) function continues traversing until a non empty tile is found. If we remove the `showEmptyTiles` options and always stop at empty tiles we can remove that function.

liberostelios:
Great news! Glad to see this PR is actually useful to you.

> Just did a quick test to reproduce the the issue you mentioned in #118 but I'm still unable to see it. Have you reproduced it in the example file in the repo?

I haven't tested this, to be honest, so it might have to do with our specific dataset. I'll check and come back.

> If you have time would you mind helping to update the manual test cases, as well, to reflect this correct behavior? The last three deal with empty tiles: [this one](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/TESTCASES.md#verify-maxdepth-limit-does-not-stop-at-empty-tiles), [this one](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/TESTCASES.md#verify-tileset-with-missing-mid-tile-content-loads-and-renders-correctly), and [this one](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/TESTCASES.md#verify-tileset-with-missing-mid-tile-content-loads-and-renders-correctly-1).
Sure! I'll try to spend some time in the next couple of days to test the behaviour mentioned above and check the tests.

src/base/TilesRendererBase.js
liberostelios: Done with b3c03c8eea6c4ef37c25ad534c2cdb8c4415dfd5

gkjohnson: Thanks! And sorry I thought it yesterday but never typed it.\: `showEmptyTiles` is a little confusing I think. Maybe we can rename the flag `stopAtEmptyTiles`? Or maybe invert the flag and call it `ignoreEmptyTiles`? I'm not sure if you have any preferences

liberostelios: Good point! I'll rename it.

gkjohnson:
I just took a deeper look and I think it's good! There's one more spot [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/base/traverseFunctions.js#L170-L179) that will skip empty tiles when traversing children specifically when `loadSiblings` is true but that should just affect what tiles get loaded and are marked as active. I think it can stay as is for now and we can handle it when removing `recursivelyMarkUsed` (https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/119#issuecomment-701509766).

Once you get a chance to take a look at the tests we can get this merged!

liberostelios:
I've update the two test cases to reflect the new change. Also, added a parameter to enable/disable `stopAtEmptyTiles` in the main example.

I am not sure about the third testcase, though. I am a bit confused, as it seems that the title is wrong on that one. Probably that should be just "Verify tileset with additive refinement renders correctly" and it shouldn't be affected by the `stopAtEmptyTiles` option, should it? At least in practice, I see no difference when I enable/disable it.

gkjohnson:
> I am not sure about the third testcase, though. I am a bit confused, as it seems that the title is wrong on that one.

You're right I don't think the third one is relevant here. The other two test updates look good! I'll merge after the bundle conflicts are resolved somehow -- they don't need to be committed in PRs but if you just want to take your changes or overwrite with the build changes in master. I'll update them once everything is merged.

This all looks ready to go otherwise, though!

liberostelios:
I can revert the bundle files to their original state to resolve the conflict, so you take care of building the example with the new change when it's all merged.

gkjohnson:
> I can revert the bundle files to their original state to resolve the conflict, so you take care of building the example with the new change when it's all merged.

Sounds perfect!

liberostelios:
ecc39ad did the trick! Should be fine to merge, now. 👍

gkjohnson:
Great thanks for putting the time into this! I'll run through the tests this week and publish to npm when I get a chance.

gkjohnson:
Just published in [v0.1.6](https://github.com/NASA-AMMOS/3DTilesRendererJS/releases/tag/v0.1.6).

PR #121, opened 2020-10-04
Add batch table example use

### LINKED ISSUE #68: Add example to demonstrate batch table use
For all model loader types

gkjohnson:
Fix #68

Adds some documentation to the readme on how to use and read batch table data.

@liberostelios It would be great to get your feedback on the README update to make sure it's useful. I've updated the [b3dmExample](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/b3dmExample.html) page to display the batch table data, as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,41 @@ function renderLoop() {
 renderLoop();
 ```
 
+## Read Batch Id and Batch Table Data
+
+How to find the batch id and batch table associated with a mesh and read the data.
+
+```js
+const tilesRenderer = new TilesRenderer( './path/to/tileset.json' );
+
+// ...checking intersections...
+
+const intersects = raycaster.intersectObject( scene, true );
+if ( intersects.length ) {
+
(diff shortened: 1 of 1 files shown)


liberostelios:
I think the example is pretty clear to me.

Just a minor suggestion, maybe. Since you are being defensive regarding the existence of `_batchid` in the geometry's attributes, you might as well check for the existence of the `'BatchTableKey'` key in `batchTable`. This will also demonstrate the use of the `getKeys()` function. Something like:

```JavaScript
if ( batchTable.getKeys().includes( 'BatchTableKey' ) ) {

	console.log( batchTable.getData( 'BatchTableKey' )[ hoveredBatchid ] ); 

}
```

gkjohnson:
You're right that probably is worth guarding against -- I added the following check:

```js
		const batchData = batchTable.getData( 'BatchTableKey' );
		if ( batchData ) {

			console.log( batchData[ hoveredBatchid ] );

		}
```

PR #127, opened 2020-10-20
Fix linting issues

liberostelios:
Just fixes a few linting issues and thought you might want to incorporate that.


gkjohnson:
Thanks! I think the lint command in the package.json is malformed (or at least system dependent) which is why these weren't caught -- I'll give it a look and try to get that fixed in master.

PR #141, opened 2020-12-23
loading from Ion

PetterGs:
I needed to load tiles from Cesium Ion for a project, So I decided that perhaps this feature could be added here too. I understand that Ion itself isn't open source. 

This branch adds the ability to load Cesium Ion assets in the following ways:
1. By adding an Ion asset id instead of url, accompanied to the Ion access token. [link to example](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/141/files#r548169359)
2. By adding an ION endpoint url with a temporary 1 hour bearer token.
3. By adding just the Ion asset id, thus using the Ion default token automatically.

Also if would perhaps be better to add this feature as a url parser module for other rest endpoints instead. I don't know of any other 3dTiles endpoints except Ion however.

This isn't a pull request, It's a draft made for discussion mainly. I can make all needed changes before actually making this a pull request, if the maintainers agree.

This branch, if approved, would make it easier to validate the loading of 3dTiles, as well as comparing own 3dTiles to the ones of Cesium. This would greatly speed up development of this project, by making it easier to discover bugs or lacking features.

~~There is also a bug that I am unsure of how to fix properly. I had to comment out line 599 in _TilesRenderer.js_. If I let the updateMatrix run, it rotated the tiles 90 degreess and the resulting bounds for frustum culling would not align to the actual model presented in the scene. I'm not sure why that is.~~

I'm not affiliated with Cesium in any way. I just needed this feature for a project.

Here are two Ion asset ids that can show this branch in action:
[40866 - Cesium office](https://pettergs.github.io/3DTilesRendererJS-ION/example/bundle/ionExample.html#40866l)
[69380 - Melbourne](https://pettergs.github.io/3DTilesRendererJS-ION/example/bundle/ionExample.html#69380) 

image


PetterGs:
[I had to set draco loading to use javascript instead of WASM because using WASM caused the sandboxed memory of my browser to run out. This could be solved by either calling the loading of gltf models via a bottlenecking function or doing it in workers.](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/141/files#diff-3762573ed5374d6ca2fe23b228ee6a0338195b5fcb5849c5ec48f7c7acaea525R121)

example/ionExample.js
```diff
@@ -0,0 +1,731 @@
...
+	'showThirdPerson': false,
+	'showSecondView': false,
+	'reload': reinstantiateTiles,
+
+};
+
+init();
+animate();
+
+function loadIonJson( assetId, accessToken ) {
```
PetterGs, line 80: Loading of bearer token in advance

example/ionExample.js
```diff
@@ -0,0 +1,731 @@
...
+
+	tiles.manager.addHandler( /\.gltf$/, {
+
+		parse( ...args ) {
+
+			// Note the DRACO compression files need to be supplied via an explicit source.
+			// We use unpkg here but in practice should be provided by the application.
+			const dracoLoader = new DRACOLoader();
+			dracoLoader.setDecoderPath( 'https://unpkg.com/three@0.123.0/examples/js/libs/draco/gltf/' );
+			dracoLoader.setDecoderConfig( { type: "js" } ); // WASM overloads memory without some sort of bottleneck on loading
```
PetterGs, line 121: I had to set draco loading to use javascript instead of WASM in the ionExample.js script. Because using WASM caused the sandboxed memory of my browser to run out. This could be solved by either calling the loading of gltf models via a bottlenecking function or doing it in workers.

src/base/TilesRendererBase.js
```diff
@@ -32,13 +34,18 @@ export class TilesRendererBase {
 
 	}
 
-	constructor( url ) {
+	//The url can be:
+	//A plain url to the tileset json file
+	//A Cesium Ion asset number with access token
+	//A url to the Cesium Ion json file with bearer token as the ionAccessToken. This needs to be fetched from the Ion endpoint separately
+	constructor( url, ionAccessToken = defaultIonToken ) {
```
PetterGs, line 41: It would naturally be better to load assets via an options object instead. This is mostly a quick example.

gkjohnson:
Hey @PetterGs! Thanks for the PR -- I think making the renderer work more cleanly with official Cesium Ion data or at least providing an example of how is a great addition so I'm happy to see this. It'll take me a bit to get through the code and I haven't used Ion myself so bear with me as I intermittently add questions when I have time. In the mean time a few questions:


1.

> Here are two Ion assets ids that can be used "for free":
> 40866 - Cesium office
> 69380 - Melbourne

Thanks for the assets -- can you elaborate on what "for free" here means in quotes? Is there a trial token or something that we can include in the repo within Cesiums terms for the sake of demos? 

2. Is there some documentation on what this "Ion Asset JSON" is and how to use it so I can understand the changes a bit more?

3.

> There is also a bug that I am unsure of how to fix properly. I had to comment out line 599 in TilesRenderer.js. If I let the updateMatrix run, it rotated the tiles 90 degreess and the resulting bounds for frustum culling would not align to the actual model presented in the scene. I'm not sure why that is.

That is definitely a bug -- looks like I accidentally committed it the other day. I'll fix it and make a new release later today. Thanks for the catch!

example/ionExample.js
gkjohnson: Interesting -- there are a couple things happening that might be leading to extra memory being created / kept around:

- We have to create a [duplicate buffer of just the GLTF bytes from the b3dm buffer](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/B3DMLoader.js#L17) because three.js' GLTFLoader does not account for a typed arrays byte offset or byte length. I suggested respecting the offset and length [here](https://github.com/mrdoob/three.js/pull/19616#issuecomment-642281533) but it hasn't been added, yet.

- Both the featureTable and batchTable [retain a reference to the full original B3DM buffer](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/base/B3DMLoaderBase.js#L70-L75) meaning that after draco decoding has unpacked the byte buffers we're still keeping around the originally loaded compressed buffer, as well as the new decompressed buffer. This can be fixed by copying the relevant feature and batch table data to new buffers which will cut down on the retained memory there.

I'm not sure if those two things are the source of this issue, though, particularly if using the js version of the decoder is working just fine. If you can track down an issue a bit more it might be worth opening an issue in the three.js repo.

src/three/TilesRenderer.js
```diff
@@ -247,12 +262,55 @@ export class TilesRenderer extends TilesRendererBase {
...
+		var me12 = me[ 12 ], me13 = me[ 13 ], me14 = me[ 14 ], me15 = me[ 15 ];
+
+		planes[ 0 ].setComponents( me3 - me0, me7 - me4, me11 - me8, me15 - me12 ).normalize();
+		planes[ 1 ].setComponents( me3 + me0, me7 + me4, me11 + me8, me15 + me12 ).normalize();
+		planes[ 2 ].setComponents( me3 + me1, me7 + me5, me11 + me9, me15 + me13 ).normalize();
+		planes[ 3 ].setComponents( me3 - me1, me7 - me5, me11 - me9, me15 - me13 ).normalize();
+		planes[ 4 ].setComponents( me3 - me2, me7 - me6, me11 - me10, me15 - me14 ).normalize();
+		planes[ 5 ].setComponents( me3 + me2, me7 + me6, me11 + me10, me15 + me14 ).normalize();
+
+	}
```
gkjohnson, line 312: Which version of three.js are you using that doesn't include `setFromProjectionMatrix`?

PetterGs: I had to use 0.111.0 for a project. I can remove this function if you wish.

PetterGs:
> Thanks for the assets -- can you elaborate on what "for free" here means in quotes? Is there a trial token or something that we can include in the repo within Cesiums terms for the sake of demos?
> 

Melbourne is provided by the [City of Melbourne Open Data team](https://data.melbourne.vic.gov.au/Property/City-of-Melbourne-3D-Textured-Mesh-Photomesh-2018/d5tb-r7a6). [Here are the terms.](https://data.melbourne.vic.gov.au/stories/s/data-principles/9f8u-v2fn?src=hdr)

The Cesium office only has the Cesium website as attribution data, I can ask about their license.

> Is there some documentation on what this "Ion Asset JSON" 

I haven't found any documentation yet. It's simply a small data file that among other things, gives you a one hour token to see ion 3dTiles with. Just so you don't have to share your real token with anyone. It's art of the [ION Rest API](https://cesium.com/docs/rest-api/)

>Hey @PetterGs! Thanks for the PR

Happy to help.

PetterGs:
I can make all the necessary changes.

gkjohnson:
> It's simply a small data file that among other things, gives you a one hour token to see ion 3dTiles with. Just so you don't have to share your real token with anyone. It's part of the ION Rest API.

Got it -- I think I understand the API a bit now. But it doesn't seem like we need to really make any changes to the core files in order to work with Cesium Ion, right? I think an example that instantiates a tiles renderer like so would be enough:

```js
const ionJsonUrl = `https://api.cesium.com/v1/assets/${ionAssetId}/endpoint?access_token=${accessToken}`;
fetch( ionJsonUrl ).then( json => {

    tiles = new TilesRenderer( json.url );
    tiles.fetchOptions.headers = {
        Authorization : `Bearer ${json.accessToken}`,
    };
    tiles.update();

} );
```

Am I understanding right? Is the issue with fetching child tilesets? If so would it make sense to provide an overrideable function that transforms URLs before fetching?

Regarding the default token -- from this issue (https://github.com/CesiumGS/cesium/issues/8433) looks like their default access token will expire with every cesium release so it might not be great for a long term demo but can be used for validation. A demo like you shown that lets you type in an asset id (or provides a dropdown of a couple asset ids) and paste in an access token would be good, though.

src/three/TilesRenderer.js
gkjohnson: I see -- I'm a bit torn but I don't know if I want to start supporting every old version of three.js by copying code from the newer versions. Otherwise we'd have to do it for every function that gets added that we want to use. If you're using a version of three that doesn't support this function and you can't upgrade you should be able to patch it in like so:

```js
import { Frustum } from 'three';
Frustum.prototype.setFromProjectionMatrix = function () {
  /* ... */
};
```

example/ionExample.js
gkjohnson: I think I found the source when I was glancing over the code -- the issue was that a new `DracoLoader` was being created for every tile being loaded which loaded a new instance of the WASM / JS code which also slowed down parsing a lot. Creating a single common DracoLoader prevents the draco code from being loaded multiple times and in turn greatly speeds up the loading and parsing of tiles. Here's a diff of the change I made to the use case example in the README: https://github.com/NASA-AMMOS/3DTilesRendererJS/commit/cb3328c9af176415ecb0620c05cb855573ccadca

src/three/TilesRenderer.js
PetterGs: You are right. I'll remove it.

PetterGs:
> Am I understanding right? Is the issue with fetching child tilesets? If so would it make sense to provide an overrideable function that transforms URLs before fetching?

Not quite, since fetching through this API needs to use a URL object with query parameters.  [Like so - line 554.](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/141/files#diff-073e29fd6323d0defdcc69b38d6d61b0f1aaa6a486d29285d087139975afe5a9R554)

I also fetch the json inside TilesRendererBase, for people who do not mind sharing their token. The prefetch is for people who are doing the fetching on a separate server for instance.

I think the "correct" way to do this is to create a Ion url parser class that you can choose via options. In this case the Rest calls would be abstracted away. This would help anyone to easily and cleanly create other such classes for other rest api's.

example/ionExample.js
PetterGs: fixed it in my example.

src/base/TilesRendererBase.js
```diff
@@ -456,9 +549,16 @@ export class TilesRendererBase {
 
 					return Promise.resolve();
 
+				}
+				let uri = tile.content.uri;
+				if ( this.ionAssetId ) {
```
PetterGs, line 554: Setting search params to endpoint

PetterGs:
Wow. Melbourne is loading beautifully now with wasm enabled.

gkjohnson:
> Wow. Melbourne is loading beautifully now with wasm enabled.

I know I was so confused when I saw it was loading so slowly at first 😁 

> Not quite, since fetching through this API needs to use a URL object with query parameters. Like so

I see so the "version" search parameter is required, then? How about providing a `preprocessURL` function that gets called before the root, external tileset, and geometry urls are fetched would let you add any search parameters needed:

```js
tiles.prepreocessURL = url => {

    // called before loading a tileset or tile geometry url
    const ionVersion = new URL( tiles.rootURL ).searchParams.get( 'v' );
    const newURL = new URL( url );
    newURL.searchParams.set( 'v', ionVersion );
    return newURL;

};
```

Then loading an ion tileset for a user would look like so. If the temporary token is not needed then they can skip the initial fetch:

```js
const ionJsonUrl = `https://api.cesium.com/v1/assets/${ionAssetId}/endpoint?access_token=${accessToken}`;
fetch( ionJsonUrl ).then( json => {

    tiles = new TilesRenderer( json.url );
    tiles.fetchOptions.headers = {

        Authorization : `Bearer ${json.accessToken}`,

    };
    tiles.preprecessURL = url => {

        // prcess url, add url params, etc
        return url;

    };
    tiles.update();

} );
```

This seems a bit more flexible to me and avoids the need for any ion-specific code to be added to the core files -- and if anyone has any other server-specific need to add search params needs they can be handled similarly. What do you think?

PetterGs:
> This seems a bit more flexible to me and avoids the need for any ion-specific code to be added to the core files -- and if anyone has any other server-specific need to add search params they can be handled similarly. What do you think?

Sounds good. If I understand you correctly the ion specific code will only exist in the ionExample page. This sounds very clean. I'll make a revision soon.

gkjohnson:
> If I understand you correctly the ion specific code will only exist in the ionExample page. This sounds very clean. I'll make a revision soon.

Thats right -- thank you very much! I appreciate it.

Happy holidays!

PetterGs:
I made the necessary changes.

I wonder if I could ask you a favor? Could add example/bundle files to the gitignore list and add some sort of github build action to run `npm run start` instead. That would be great.

PetterGs:
One bug still exists. If I try to load the assetId from url hash. The model is flipped. I'm not sure why that is. Some sort of timing error with matrix update perhaps?

gkjohnson:
Great thanks! I'll take a closer look soon.

> I wonder if I could ask you a favor? Could add example/bundle files to the gitignore list and add some sort of github build action to run npm run start instead. That would be great.

Do you have any write ups or recommendations on how to do this? As far as I can tell once a file is committed in the project the .gitignore file does not affect it. I agree it would be nice if there was an easy way to do that, though. For the moment you can take the master changes on the bundled files and only commit the files you intend to change.

>

src/three/TilesRenderer.js
```diff
@@ -247,6 +262,32 @@ export class TilesRenderer extends TilesRendererBase {
...
+					this.group.quaternion.x = rotationToNorthPole.x;
+					this.group.quaternion.y = rotationToNorthPole.y;
+					this.group.quaternion.z = rotationToNorthPole.z;
+					this.group.quaternion.w = rotationToNorthPole.w;
+
+					this.group.position.y = - distanceToEarthCenter;
+
+				}
+
+			} );
```
gkjohnson, line 290: I don't think repositioning or rotating the tile set based on assumptions about whether a tileset is geopositioned or not should be a part of the core behavior. This should be the responsibility of the application. Perhaps we can move this to the example page, as well?

gkjohnson:
> I wonder if I could ask you a favor? Could add example/bundle files to the gitignore list and add some sort of github build action to run npm run start instead. That would be great.

I've looked into it a bit more and made issue #145 for building and commiting the example pages to a separate branch which I might have time to get to in the coming weeks.

> One bug still exists. If I try to load the assetId from url hash. The model is flipped. I'm not sure why that is. Some sort of timing error with matrix update perhaps?

This is because when a tileset is loaded from the has it defaults to +Z (which is the convention for cesium tilesets) instead of +Y which is what our example tileset defaults to. You can fix it by adjusting this line (in `ionExample.js`):

https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/4fac2ac830d0fafdd948473eceb101ce6260ab2d/example/index.js#L68

src/three/TilesRenderer.js
PetterGs: moved

gkjohnson: Great thanks!

example/ionExample.js
```diff
@@ -0,0 +1,769 @@
...
+const hashUrl = window.location.hash.replace( /^#/, '' );
+let camera, controls, scene, renderer, tiles, cameraHelper;
+let thirdPersonCamera, thirdPersonRenderer, thirdPersonControls;
+let secondRenderer, secondCameraHelper, secondControls, secondCamera;
+let orthoCamera, orthoCameraHelper;
+let raycaster, mouse, rayIntersect, lastHoveredElement;
+let offsetParent;
+let statsContainer, stats;
+
+const defaultIonToken = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJqdGkiOiIwY2Q2MzQ1OS1kNjI4LTRiZDEtOWVkZC1kMWI4YzAyODU3OGMiLCJpZCI6MjU5LCJpYXQiOjE2MDY4NzMyMTh9.8EwC6vilVHM2yizt8nG6VmbNu66QiCrk3O-1lEDPI9I';
```
gkjohnson, line 51: Can we add a link to where this token came from? It sounds like it will break eventually so it would be good to track that.

src/three/TilesRenderer.js
```diff
@@ -87,6 +87,12 @@ export class TilesRenderer extends TilesRendererBase {
 	}
 
 	/* Public API */
+	getRootMatrix() {
+
+		return this.root.cached.boxTransform;
+
+	}
+
```
gkjohnson, line 95: Can we change this to use a similar pattern to `getBounds` and pass in a matrix4 to avoid accidental modification of the root box transform instance and return true / false depending on if the root is available. Lets also rename it `getBoundsTransform` so it's clear this and `getBounds` are related:

```js
getBoundsTransform( target : Matrix4 ) : Boolean
```

gkjohnson: We should also add it to the typescript definitions.

gkjohnson:
Thanks for the updates! Looks like it's just about there.

I've update the project so changed files are no longer present when running `npm start` so if you can merge in master and remove the bundle file changes by just taking the master versions that would be great.

And lastly updating the docs would be nice. We should add documentation for `addBoundsTransform` and `preprocessURL` and if you'd like to add a snippet to the [Use section](https://github.com/NASA-AMMOS/3DTilesRendererJS#use) of the README to demonstrate users can use the renderer with Cesium Ion assets that would be appreciated, too!

PetterGs:
Done

gkjohnson:
Great, thanks! I can take care of the last typescript definition. I asked above but where did the default Cesium token come from so we can document it? I'll add the appropriate comment.

PetterGs:
done
> Great, thanks! I can take care of the last typescript definition. I asked above but where did the default Cesium token come from so we can document it? I'll add the appropriate comment.

Done

gkjohnson:
Great, thank you for all the changes! There are a couple more things I want to do before a new release but I should have it done in the next week or so.

gkjohnson:
Just published the new functions from this PR in release v0.2.3 with a couple small changes. Changelog [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/releases/tag/v0.2.3). Thanks again!

PR #146, opened 2020-12-30
Update main example

PetterGs:
A small fix to draco loading. I assume that the GLTFLoader doesn't need to be constructed in the parse function.

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -93,15 +93,14 @@ function reinstantiateTiles() {
 	const dracoLoader = new DRACOLoader();
 	dracoLoader.setDecoderPath( 'https://unpkg.com/three@0.123.0/examples/js/libs/draco/gltf/' );
 
-	const loader = new GLTFLoader( tiles.manager );
-	loader.setDRACOLoader( dracoLoader );
-
 	tiles = new TilesRenderer( url );
 	tiles.fetchOptions.mode = 'cors';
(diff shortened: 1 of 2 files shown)


example/index.js
```diff
@@ -91,16 +91,17 @@ function reinstantiateTiles() {
 	// Note the DRACO compression files need to be supplied via an explicit source.
 	// We use unpkg here but in practice should be provided by the application.
 	const dracoLoader = new DRACOLoader();
-	dracoLoader.setDecoderPath( 'https://unpkg.com/three@0.116.1/examples/js/libs/draco/gltf/' );
+	dracoLoader.setDecoderPath( 'https://unpkg.com/three@0.123.0/examples/js/libs/draco/gltf/' );
+
+	const loader = new GLTFLoader( tiles.manager );
+	loader.setDRACOLoader( dracoLoader );
```
gkjohnson, line 97: I prefer not to reuse a single instance of GLTFLoader because in practice once a load has begun the options cannot be changed so if multiple loads happen simultaneously that require different settings it could cause issues. It won't happen here but I think it's best to create a new one every time. DRACOLoader is unique because of how it loads the WASM multiple times otherwise which I plan to submit a PR to three to update the docs for.

Updating the decoder path is good, though. Thanks!

PetterGs: very well

gkjohnson:
Thanks!

PR #154, opened 2021-01-21
Fix I3DM files not correctly loading external assets

### LINKED ISSUE #153: Some models of bim still cannot be displayed and an error occurs
Some models of bim still cannot be displayed and an error occurs. I provide the address of cesium ion for your analysis. It is normal in cesium. 
token：eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJqdGkiOiIwM2RiZjA2Zi04ZmVkLTQ3ZDEtODcyMi1iZmRiNThmOGIyZjAiLCJpZCI6MTk4MiwiaWF0IjoxNTMwOTE1MjcwfQ.NrYvOPolAxoq7ueAJUNDjVY9BzM1kBTjn1FX3J0nNWc

asset id：257841
image
The other is a model that can be displayed normally, the token is the same, and the asset id is 257692.
image

gkjohnson:
Fix #153 

Before nothing would display now i3dm models are showing:

image

### DIFF
--- a/src/base/CMPTLoaderBase.d.ts
+++ b/src/base/CMPTLoaderBase.d.ts
@@ -15,6 +15,7 @@ export interface CMPTBaseResult {
 
 export class CMPTLoaderBase {
 
+	workingPath : string;
 	load( url : String ) : Promise< CMPTBaseResult >;
 	parse( buffer : ArrayBuffer ) : CMPTBaseResult;
 
--- a/src/base/CMPTLoaderBase.js
+++ b/src/base/CMPTLoaderBase.js
@@ -6,6 +6,7 @@ export class CMPTLoaderBase {
 	constructor() {
 
 		this.fetchOptions = {};
+		this.workingPath = '';
 
 	}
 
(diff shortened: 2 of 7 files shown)


yoyomule:
image
Hello, I found that 0.26 fixes some i3dm and can open the model, but the model has a lot of dislocations.
Token:
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJqdGkiOiIwM2RiZjA2Zi04ZmVkLTQ3ZDEtODcyMi1iZmRiNThmOGIyZjAiLCJpZCI6MTk4MiwiaWF0IjoxNTMwOTE1MjcwfQ.NrYvOPolAxoq7ueAJUNDjVY9BzM1kBTjn1FX3J0nNWc
ID:267389

gkjohnson:
Thanks for the continued issue reports! I've made issue #158 to track it and I'll take a look when I have some time maybe over the next couple weeks.

gkjohnson:
@yoyomule fixed the issue in #159 and added support for some other feature semantics, as well. The missing geometry was addressed by the latest version of three.js. And the rotated geometry was caused by ignoring original transformation information in the GLTF when converting to instanced meshes.

I'll publish a new version in a bit!

PR #163, opened 2021-03-01
Improve the performance of attribute parsing for .b3dm

liberostelios:
The previous implementation would use a lot of string allocations which would cause the garbage collector to kick in often and that was a major fit for performance.

This patch fixes the issue by using the built-in TextDecoder class, but it might break compatibility with old browsers (e.g. Internet Explorer).

We noticed that this was causing major hick ups before. I don't think it's worth to keep compatibility with old browsers in this scenario. Or we can just use the old function as a fallback in case TextDecoder is not present.

### DIFF
--- a/src/utilities/arrayToString.js
+++ b/src/utilities/arrayToString.js
@@ -1,12 +1,6 @@
 export function arrayToString( array ) {
 
-	let str = '';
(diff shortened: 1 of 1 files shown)


gkjohnson:
Sounds like a good change. Thanks! I'll publish a new version later this week.

PR #165, opened 2021-03-03
fix typescript decalre

Aierklk:


src/three/TilesRenderer.d.ts
```diff
@@ -22,8 +22,9 @@ export class TilesRenderer extends TilesRendererBase {
 	setResolution( camera : Camera, resolution : Vector2 ) : Boolean;
 	setResolutionFromRenderer( camera : Camera, renderer : WebGLRenderer ) : Boolean;
 
+	onLoadTileSet : ( ( tileSet : Object ) => void ) | null;
 	onLoadModel : ( ( scene : Object3D, tile : object ) => void ) | null;
 	onDisposeModel : ( ( scene : Object3D, tile : object ) => void ) | null;
-	forEachLoadedModel( callback : ( scene : Object3D, tile : object ) => void );
+	forEachLoadedModel : ( ( callback : ( scene : Object3D, tile : object ) => void ) => void ) | null;
```
gkjohnson, line 28: Can you explain this change and why it was wrong before? `forEachLoadedModel` is a function is not a settable field so it cannot be null.

Aierklk: sorry, That is my mistake. I push a new commit.

gkjohnson: Can we adjust it to be consistent with the other function definitions like `setResolution` and just add `: void` at the end of the existing definition for the return type?

```js
forEachLoadedModel( callback : ( scene : Object3D, tile : object ) => void ) : void;
```

Thanks again for the contribution!

gkjohnson:
I'll fix the `forEachLoadedModel` definition -- thanks again for the contribution!

Issue #168, opened 2021-03-19
Raycasting of individual meshes is broken


liberostelios:
The current implementation of raycasting for tiles breaks the traditional functionality of `Mesh.raycast()`. I believe this is a risky move because the user of the library doesn't know that and if they call `.intersectObjects()` directly on the meshes they get no return.

The current solution is very efficient and nicely thought, but this is a very important side-effect. I admit that I couldn't find a better solution to this. Maybe there is a way that `Mesh.raycast()` is only set to `emptyRaycast()` during the traversal of the tree?

Otherwise, I think it should be at least mention it in the docs as a warning. And maybe also mention that a workaround to the current situation is to loop through the meshes calling the prototype function instead:

```
let result = [];

Object.getPrototypeOf( mesh ).raycast.call( mesh, raycaster, result );
```

gkjohnson:
Hey @liberostelios! Yes there have been a few trades related to tiles made for the sake of performance including disabling raycasting on the them so the bounding volume hierarchy can be traversed in the group.raycast function. It's also not possible to reposition the tiles with `position`, `rotation`, etc without calling `updateMatrixWorld` manually because [automatic matrix updates are disabled unless the group moves](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/7ad0e30d2c9eaf352842a502678835b720452bb7/src/three/TilesGroup.js#L22-L80) since the child tiles are not expected to move.

I agree it's worth documenting both of these things. I'm not exactly sure where, though? Maybe start in the TilesRenderer class section? Or the `TilesGroup` object could be documented explicitly with some of these caveats.

> maybe also mention that a workaround to the current situation is to loop through the meshes calling the prototype function instead

What's your use case for raycasting against the individual tiles? My expectation was that users wouldn't / shouldn't need to know about individual tiles because the renderer needs explicit control over updating them and adding and removing them from the scene. The terrain should be treated as a whole group. If there's a compelling need, though, then maybe a toggle could be added to enable or disable the optimized raycast functionality.

> Maybe there is a way that Mesh.raycast() is only set to emptyRaycast() during the traversal of the tree?

This is an interesting idea but unfortunately there's no way to know when a raycast is set to `traverse = true` or when the children have finished being iterated over, as far as I know. But I'm open to ideas!

liberostelios:
Hey @gkjohnson! Thanks for the quick response.

[...]

PR #169, opened 2021-03-23
Add option to enable/disable override of raycast

liberostelios:
A potential solution for #168

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -13,6 +13,7 @@ import {
 	Math as MathUtils,
 	Frustum,
 	LoadingManager,
+	Group,
 } from 'three';
 import { raycastTraverse, raycastTraverseFirstHit } from './raycastTraverse.js';
 
@@ -67,6 +68,67 @@ export class TilesRenderer extends TilesRendererBase {
 
 	}
 
+	get optimizeRaycast() {
+
+		return this._optimizeRaycast;
+
+	}
+
+	set optimizeRaycast( value ) {
+
+		if ( this._optimizeRaycast != value ) {
+
+			this._optimizeRaycast = value;
+
+			if ( this._optimizeRaycast ) {
+
+				this.traverse( tile => {
+
+					if ( tile instanceof TilesGroup ) {
+
+						this.raycast = TilesGroup.prototype.raycast;
+
+					} else {
+
+						this.raycast = emptyRaycast;
+
+					}
+
+				} );
+
+			} else {
+
+				this.traverse( tile => {
+
+					if ( tile instanceof TilesGroup ) {
+
+						console.log( "Replace to group's raycast" );
+						tile.raycast = Group.prototype.raycast;
+
+					} else {
+
+						tile.raycast = Object.getPrototypeOf( tile ).raycast;
+
+					}
+
+				} );
+
+			}
+
+			this.traverse( tile => {
+
+				if ( tile.scene ) {
+
+					updateFrustumCulled( tile.scene, value );
+
+				}
+
+			} );
+
+		}
+
+	}
+
 	constructor( ...args ) {
 
 		super( ...args );
@@ -77,6 +139,7 @@ export class TilesRenderer extends TilesRendererBase {
 		this.activeTiles = new Set();
 		this.visibleTiles = new Set();
 		this._autoDisableRendererCulling = true;
+		this._optimizeRaycast = true;
 
 		this.onLoadTileSet = null;
 		this.onLoadModel = null;
@@ -171,6 +234,12 @@ export class TilesRenderer extends TilesRendererBase {
 	}
 
 	raycast( raycaster, intersects ) {
+
+		if ( ! this._optimizeRaycast ) {
+
+			return;
+
+		}
 
 		if ( ! this.root ) {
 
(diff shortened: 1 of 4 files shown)


gkjohnson:
Thanks for making the PR!

I think I'd like to avoid traversing over all tiles when modifying the `overrideRaycast` field so there's no surprise performance overhead when toggling the field. It can also be a bit tricky to follow because new tiles are often added and removed from the hierarchy as the camera moves around.

I was imagining keeping `TilesRenderer.raycast` unmodified and instead adjusting [TileGroup.raycast](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/7ad0e30d2c9eaf352842a502678835b720452bb7/src/three/TilesGroup.js#L16-L20) to check if `tilesRenderer.optimizeRaycast` was true before calling `TilesRenderer.raycast`:

```js
raycast( ... ) {

  if ( this.tilesRenderer.optimizeRaycast ) {
    
    this.tilesRenderer.raycast( ... )

  }

}
```

And instead of `emptyRaycast` we can replace the built raycast on all loaded tiles meshes [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/7ad0e30d2c9eaf352842a502678835b720452bb7/src/three/TilesRenderer.js#L617-L622) with one that checks whether optimizedRaycast is false:

```js
function overriddenRaycast( ... ) {

  if ( ! tilesRenderer.optimizeRaycast ) {

    Object.getPrototypeOf( this ).raycast.call( this, ... );

  }

}
```

That seems like it would have fewer moving parts to me. What do you think?

liberostelios:
That's right. I haven't thought of using `emptyRaycast` as a proxy to call the original `raycast` function. I'll do that.

liberostelios:
I've implemented the aforementioned idea. The only thing missing was that `overridenRaycast()` didn't have access to the tilesRenderer so I had to add this to each object's `userData`.

I also realized that the previous option in the example wasn't work.

I tried to verify that this works by some custom console log. I don't know if you have a better way to verify that it works.

src/three/TilesRenderer.js
```diff
@@ -28,7 +28,15 @@ const vecZ = new Vector3();
...
-function emptyRaycast() {}
+function overridenRaycast( raycaster, intersects ) {
+
+	if ( ! this.userData.tilesRenderer.optimizeRaycast ) {
+
+		Object.getPrototypeOf( this ).raycast.call( this, raycaster, intersects );
+
+	}
+
+}
```
gkjohnson, line 39: Great! This is what I was thinking. However instead of putting the tiles renderer in userdata lets create a new function per tiles renderer instead that can be used for the objects (from what I can tell userdata is really intended for serializable data). So the function can be created in the TilesRenderer constructor:

```js
constructor( ... ) {

  // ...

  const tilesRenderer = this;
  this._overridenRaycast = function( raycaster, intersects ) {

    if ( tilesRenderer.optimizeRaycast ) {

      Object.getPrototypeOf( this ).raycast.call( this, raycaster, intersects );

    }

  };

}
```

And then we can use that function on each of the loaded objects:

```js
c.raycast = this._overridenRaycast;
```

src/three/TilesRenderer.js
```diff
@@ -172,24 +181,28 @@ export class TilesRenderer extends TilesRendererBase {
...
+		if ( this.optimizeRaycast ) {
 
-			return;
+			if ( ! this.root ) {
 
-		}
+				return;
 
-		if ( raycaster.firstHitOnly ) {
+			}
```
gkjohnson, line 190: I don't think we should have to adjust the `TilesRenderer.raycast` function. Instead we should adjust [TilesGroup.raycast](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/TilesGroup.js#L16-L20) to check if the tiles renderer (which it already has a reference to) has the tiles override flag is set to true. That way `TilesRenderer.raycast` will always perform the optimized version. Perhaps we can rename this function and expose it in a future release so it can be usable independent of three.js' Raycaster mechanism.

gkjohnson (changes requested):
Awesome, thank you! Made a couple comments above. This is more along the lines of what I was thinking.

> I tried to verify that this works by some custom console log. I don't know if you have a better way to verify that it works.

I think adjusting the kitchen sink example (`example/index.js`) to get an array of leaf nodes of the tiles to raycast against, which is what it sounded like you were doing in your app, would be a good way to enable testing of the feature. That way it's clear if it breaks in the future.

Last thing would be to update the docs to document the new option!

src/three/TilesRenderer.js
liberostelios: I think this is resolved now.

src/three/TilesRenderer.js
liberostelios: Also resolved.

liberostelios:
I think the latest commit must have improved the way things work. I'll write the docs as well.

I, also, locally changed the example to do an `intersectObjects()` on the meshes themselves, instead of a recursive raycast on the main group. I chose not to push this, though. Do you think it's worth having it there? We could have an extra option (like "PURE_THREEJS" in the params of the example and in this case use the different logic to make sure we just intersect the meshes. Not sure if I see a benefit on that, but it's an easy change and I have zero problems doing this.

gkjohnson:
@liberostelios 

The code looks great now thanks for the changes! I'll wait until the docs are updated to merge.

> Do you think it's worth having it there? We could have an extra option (like "PURE_THREEJS" in the params of the example and in this case use the different logic to make sure we just intersect the meshes.

I think the code is easy enough to follow and validate. If this breaks some time in the future and we need a better way to test we can reconsider. Automated tests would be good for a case like this but setting that up for three.js and the tiles renderer is a task for another day, I think.

gkjohnson:
@liberostelios Looks like the docs have been updated. Is this all ready to go? If so I'll give it all another look and we can get it merged!

liberostelios:
Sorry, my bad. Forgot to mention it here. Yes, I am finished with the docs. I hope that's good enough.

gkjohnson:
Great! Thanks again for adding this feature and getting everything updated. It all looks good. I'll publish a new version in the next week or so.

gkjohnson:
Just published in v0.2.10. Thanks again!

Issue #182, opened 2021-05-26
Points Cloud 3D Tiles are not displayed correctly. 


RomTreg:
<img width="1224" alt="Screenshot 2021-05-26 at 23 21 41" src="https://user-images.githubusercontent.com/46277669/119725939-3685f100-be79-11eb-8f7a-3893f64ca3ec.png">
[tileset.json.zip](https://github.com/NASA-AMMOS/3DTilesRendererJS/files/6549427/tileset.json.zip)

I've attached tiles.json and screenshot just in case . 
All tiles are displayed at the center and change their orientation apparently.
Any ideas on how to fix this?

gkjohnson:
Hi @RomTreg! It's very possible it's a bug in the way point clouds are handled considering we've mostly use this with mesh data. We don't have the bandwidth to fix every issue associated with the project right now but I might be able to point to where the issue might be if you'd like to help make a PR and get it fixed (if it is an issue with the library and not generated tileset).

Are you able to provide a full tileset with pnts files instead of just the tileset.json? It's not possible to see the issue otherwise. Some more information that might help as well:

- Can you provide an image what the tileset is supposed to look like?
- Do the tile bounding boxes look correct when visualized?
- How was the tileset generated?
- Have you tested the tileset in Cesium?

Thanks for the report!

RomTreg:
Thank you for a quick response.

Answering to your question:
1. Yes, I will provide image of the Tileset and Tileset itself
2. Tileboxes are visualised aside from the model. Actually we experienced the same issue with textured models, but changing coordinates in root to 0,0,0 solve the issue
3. We generate tile sets with Metashape (agisoft)
4. We tested it with cesium and it works, but we work with threejs and do not plan to switch to cesium at the moment

RomTreg:
@gkjohnson , everything has been sent to your email. thanks!

gkjohnson:
Thanks for sending it over -- when I load one of the pnts files in the [pntsExample.html page](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/pntsExample.html) I see the following warning in the console:

> ```
> PNTSLoader.js:37 PNTSLoader: Unsupported FeatureTable feature "RTC_CENTER" detected.
> ```

[...]

PR #183, opened 2021-05-30
Fix RTC_CENTER for pnts model

alex-lancer:
Fix for [Issue 182](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/182);
- add RTC_CENTER correction in PNTSLoader
- wrap local up-axis correction rotation into condition to avoid its apply to pnts model

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -617,10 +617,12 @@ export class TilesRenderer extends TilesRendererBase {
 			// rotation fix which is why "multiply" happens here.
 			if ( extension !== 'pnts' ) {
 
-				scene.matrix.multiply( tempMat ).premultiply( cachedTransform );
-				scene.matrix.decompose( scene.position, scene.quaternion, scene.scale );
+				scene.matrix.multiply( tempMat );
 
 			}
+
+			scene.matrix.premultiply( cachedTransform );
+			scene.matrix.decompose( scene.position, scene.quaternion, scene.scale );
 			scene.traverse( c => {
 
 				c[ INITIAL_FRUSTUM_CULLED ] = c.frustumCulled;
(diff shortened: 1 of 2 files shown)


src/three/TilesRenderer.js
```diff
@@ -615,8 +615,12 @@ export class TilesRenderer extends TilesRendererBase {
...
 			// (such as applying RTC_CENTER) meaning they should happen _after_ the z-up
 			// rotation fix which is why "multiply" happens here.
-			scene.matrix.multiply( tempMat ).premultiply( cachedTransform );
-			scene.matrix.decompose( scene.position, scene.quaternion, scene.scale );
+			if ( extension !== 'pnts' ) {
+
+				scene.matrix.multiply( tempMat ).premultiply( cachedTransform );
+				scene.matrix.decompose( scene.position, scene.quaternion, scene.scale );
+
+			}
```
gkjohnson, line 623: The only matrix we don't want to apply is the `tempMat` matrix when a model is made from points -- the `cachedTransform` matrix still needs to be applied because it represents and offset transformation from the tileset tiles. Perhaps the cachedTransform multiplication and decompose can be applied after this condition?

```js

// ...

if ( extension !== 'pnts' ) {

	scene.matrix.multiply( tempMat );

}

scene.matrix.premultiply( cachedTransform );
scene.matrix.decompose( scene.position, scene.quaternion, scene.scale );

// ...
```

gkjohnson (changes requested):
Thanks for a PR! There's just a small change that needs to be made but overall it looks good!

src/three/TilesRenderer.js
alex-lancer: Done

gkjohnson (approved):

Issue #143, opened 2020-12-24
Prioritize loads of closer tiles


gkjohnson:
We want tiles that are nearby and noticeable to be loaded more quickly. This could be achieved by taking into account the apparent error or distance to the camera when sorting the tiles to load. What's in view of the frustum should be prioritized, as well.

**Options**

```js
function priorityFunc( tile ) {

    return tile.__error;

}
```

```js
function priorityFunc( tile ) {

    return tile.__distance;

}
```

dbuck:
I did a little looking at this, as I was seeing more delay while loading the higher detail tiles than I was expecting to see.

In order to do some prioritizing by __distance or __error, I had to adjust TilesRenderer.calculateError to still run for tiles which have geometricError of 0. Doing that required setting a min(Epsilon, tile.geometricError), because, div0 ;)

Once full detail tiles also included the error calc, then a test sorting function which considered both tree depth and screen space error started to work a bit better, though I'm sure it will adjust further as we get more test tilesets to work with:

```
// A priority callback that includes some screen space error consideration in addition to default tree depth
function priorityFunc( tile ) {

  const defaultPriority = 1 / ( tile.__depthFromRenderedParent + 1 );
  let errorPriority = 1 / (tile.__error + 1);
  
  // if error priority is 0, then we're *VERY* close to that tile, and should bump it up a lod level?
  // TBD if this is horribly bad in practice? It may work as long as it doesn't jump multiple levels? TBD
  if (errorPriority === 0) {
    errorPriority = -0.5;
  }

  return defaultPriority - errorPriority;
}

tiles.downloadQueue.priorityCallback = priorityFunc;
tiles.parseQueue.priorityCallback = priorityFunc;
```

gkjohnson:
Hey @dbuck! Thanks for sharing your findings. I'm feeling more and more like somehow using the screen space error to determine drive order would be a pretty good improvement.

> In order to do some prioritizing by __distance or __error, I had to adjust TilesRenderer.calculateError to still run for tiles which have geometricError of 0.

I believe if `geometricError` is `0` then `tile.__error` should always be zero so I don't think any adjustments should be needed to use error in the function. `tile.cached.distance` will only be set when fully running `computeError`, though.

[...]

PR #189, opened 2021-06-23
[143] - Prioritize load of closer tiles

dbuck:
Proposal for #143 

Adjust the priority queue callback sig to allow direct comparing of tiles, in order to have tiles compare differently for different render states. Ex: Prefer tiles in view, if tiles are in view, prefer tiles with lower better error values, and if error values are the same, prefer by distance to the camera(s)

Adjust TilesRenderer.calculateError so that it will calculate distance/error even for tiles at 0 geometricError, to allow using those values to prioritize the parsing in addition to downloading tiles.

While thinking about it a little more, I want to make sure that the values needed to make a good default priority sort are documented in some way. (Because I want to make a slightly more MP specific sort in my usage !!). That said, I wasn't sure how to best document that, and wanted to try and capture it in the typedefs.

So there is a second changeset that starts to add some typscript type checking in the js files linked from #190 which I wasn't sure if you were interested enough in to include in this PR, but they could go into a followup pretty quickly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -622,7 +622,7 @@ The maximum percentage of [minSize](#minSize) to unload during a given frame.
 ### .unloadPriorityCallback
 
 ```js
-unloadPriorityCallback = null : ( itemA, itemB ) => Number
+unloadPriorityCallback = null : ( item ) => Number
 ```
 
 Function to derive the unload priority of the given item. Higher priority values get unloaded first.
--- a/README.md
+++ b/README.md
@@ -586,7 +586,7 @@ The maximum number of jobs to be processing at once.
 ### .priorityCallback
 
 ```js
-priorityCallback = null : ( item ) => Number
+priorityCallback = null : ( itemA, itemB ) => Number
 ```
 
 Function to derive the job priority of the given item. Higher priority values get processed first.
@@ -622,7 +622,7 @@ The maximum percentage of [minSize](#minSize) to unload during a given frame.
 ### .unloadPriorityCallback
 
 ```js
-unloadPriorityCallback = null : ( item ) => Number
+unloadPriorityCallback = null : ( itemA, itemB ) => Number
 ```
 
 Function to derive the unload priority of the given item. Higher priority values get unloaded first.
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -24,13 +24,15 @@ const priorityCallback = ( a, b ) => {
 		// tiles which have greater error next
 		return a.__error - b.__error;
 
-	} else {
+	} else if ( a.__distanceFromCamera !== b.__distanceFromCamera ) {
 
 		// and finally visible tiles which have equal error (ex: if geometricError === 0)
 		// should prioritize based on distance.
 		return a.__distanceFromCamera - b.__distanceFromCamera;
 
 	}
+
+	return 0;
 
 };
 
@@ -39,14 +41,7 @@ const priorityCallback = ( a, b ) => {
  * @param {Tile} tile
  * @returns number
  */
-const lruPriorityCallback = ( tile ) => {
-
-	const defaultPriority = 1 / ( tile.__depthFromRenderedParent + 1 );
-	const errorPriority = 1 / ( tile.__error + 1 );
-	const priority = defaultPriority - errorPriority;
-	return priority;
-
-};
+const lruPriorityCallback = ( tile ) => 1 / ( tile.__depthFromRenderedParent + 1 );
 
 export class TilesRendererBase {
 
@@ -227,7 +222,7 @@ export class TilesRendererBase {
 
 		// Expected to be set during calculateError()
 		tile.__distanceFromCamera = Infinity;
-		tile.__error = 0.0;
+		tile.__error = Infinity;
 
 		tile.__inFrustum = false;
 		tile.__isLeaf = false;
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -471,7 +471,6 @@ export class TilesRendererBase {
 				return this.fetchTileSet( uri, Object.assign( { signal }, this.fetchOptions ), tileCb );
 
 			} )
-				.then( res => res.json() )
 				.then( json => {
 
 					// if it has been unloaded then the tile has been disposed
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -5,8 +5,48 @@ import { PriorityQueue } from '../utilities/PriorityQueue.js';
 import { determineFrustumSet, toggleTiles, skipTraversal, markUsedSetLeaves, traverseSet } from './traverseFunctions.js';
 import { UNLOADED, LOADING, PARSING, LOADED, FAILED } from './constants.js';
 
-// Function for sorting the evicted LRU items. We should evict the shallowest depth first.
-const priorityCallback = tile => 1 / ( tile.__depthFromRenderedParent + 1 );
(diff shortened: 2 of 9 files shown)


src/base/TilesRendererBase.js
```diff
@@ -185,7 +225,10 @@ export class TilesRendererBase {
 
 		}
 
+		// Expected to be set during calculateError()
+		tile.__distanceFromCamera = Infinity;
```
dbuck, line 229: Should this also set `tile.__error = Infinity` here?

src/three/TilesRenderer.js
```diff
@@ -807,13 +802,14 @@ export class TilesRenderer extends TilesRendererBase {
 
 					const pixelSize = info.pixelSize;
 					error = tile.geometricError / ( pixelSize * invScale );
+					// TODO: distance not updated while in orthographic views
```
dbuck, line 805: I'm not really sure what to do about this one, not setting a distance here *seems* like it shouldn't introduce different behavior from current builds with the distance based parse priority sorting, since the values will be equal.

If that is still a desired behavior, I'm wondering if we add another level in the default sort function which falls through to __depthFromRenderedParent when the distances are equal to match previous priorityFunction behavior?

Again, only an edge case with sorting the tiles at geometricError: 0, as all other tiles will end up sorted based on the error metric

gkjohnson (commented):
This looks great -- thanks for the addition! I can feel the difference when loading tiles and it should be a pretty good improvement for our tools, as well.

src/base/TilesRendererBase.js
gkjohnson: Yeah I think `tile.__error = Infinity` is fine to initialize to considering it's reset every update.

src/three/TilesRenderer.js
gkjohnson: Not considering the distance when only an orthographic camera is used makes sense to me. I agree I think we should prioritize the shallowest tile if the distances are the same. At least we need to return 0 if the distances are the same because `Infinity - Infinity = NaN`.

src/three/TilesRenderer.js
```diff
@@ -807,13 +802,14 @@ export class TilesRenderer extends TilesRendererBase {
...
 
 				} else {
 
 					const distance = boundingBox.distanceToPoint( tempVector );
 					const scaledDistance = distance * invScale;
 					const sseDenominator = info.sseDenominator;
-					error = tile.geometricError / ( scaledDistance * sseDenominator );
+					error = Math.max( 0.01, tile.geometricError ) / ( scaledDistance * sseDenominator );
```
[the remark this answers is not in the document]
dbuck: Leftover from an old experiment when I was trying to sort purely by error before adding the distance into the mix, removed

src/three/TilesRenderer.js
```diff
@@ -824,6 +820,7 @@ export class TilesRenderer extends TilesRendererBase {
 			}
 
 			tile.cached.distance = minDistance;
+			tile.__distanceFromCamera = minDistance;
```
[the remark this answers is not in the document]
dbuck: I didn't get a strong feeling yet other than it seemed odd that these properties were initialized in Base, but only ever set from the three/TilesRenderer.calculateError calls.

src/base/TilesRendererBase.js
```diff
@@ -5,8 +5,48 @@ import { PriorityQueue } from '../utilities/PriorityQueue.js';
...
+ * @returns number
+ */
+const lruPriorityCallback = ( tile ) => {
+
+	const defaultPriority = 1 / ( tile.__depthFromRenderedParent + 1 );
+	const errorPriority = 1 / ( tile.__error + 1 );
+	const priority = defaultPriority - errorPriority;
+	return priority;
+
+};
```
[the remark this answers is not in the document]
dbuck: Leftover experiment from not-understanding what __depthFromRenderedParent means, restored

README.md
```diff
@@ -622,7 +622,7 @@ The maximum percentage of [minSize](#minSize) to unload during a given frame.
 ### .unloadPriorityCallback
 
 ```js
-unloadPriorityCallback = null : ( item ) => Number
+unloadPriorityCallback = null : ( itemA, itemB ) => Number
```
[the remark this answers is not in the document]
dbuck: It might if you think there's value in evicting based on similar/inverted rules to the download/parse priority? I hadn't looked into it yet, this was a mistake from editing the readme before noting that they were different implementations

src/base/TilesRendererBase.js
```diff
@@ -420,19 +458,20 @@ export class TilesRendererBase {
...
 
 				}
 
-				const uri = this.preprocessURL ? this.preprocessURL( tile.content.uri ) : tile.content.uri;
-				return this.fetchTileSet( uri, Object.assign( { signal }, this.fetchOptions ), tile );
+				const uri = this.preprocessURL ? this.preprocessURL( tileCb.content.uri ) : tileCb.content.uri;
+				return this.fetchTileSet( uri, Object.assign( { signal }, this.fetchOptions ), tileCb );
 
 			} )
+				.then( res => res.json() )
```
gkjohnson, line 474: I missed this previously -- `this.fetchTileSet` should already be returning the fetched json so this line shouldn't be needed.

README.md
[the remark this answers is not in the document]
gkjohnson: I created #192 to track the idea -- I figure maybe a small improvement to unloading tiles could be to sort first by tile depth and then by least recently used and then maybe by distance?

src/three/TilesRenderer.js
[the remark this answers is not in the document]
gkjohnson: The *Base classes are intended to be rendering engine agnostic so in theory they can be used for implementing a renderer in other engines. `TilesRendererBase` defines and initializes any tile variables needed in order to do traversal while `three/TilesRenderer` class implements the TilesRendererBase interface to fill in the values and functionality that requires interfacing with rendering-engine-specific constructs, like three.js' camera classes, etc.

Hopefully that's more clear!

src/base/TilesRendererBase.js
dbuck: Removing, though some separation will be needed if moving forward with something like #190

I was trying to reconcile the return type difference between the `!isExternalTileset` tile fetch `Promise<Response>` and `fetchTileset(): Promise<Tileset>` by adding a generic `<T>` to PriorityQueue.add() and it seemed like `Promise<Response>` was more appropriate vs the previous `Promise<any>`.

Looking at it again I can see that is not correct, as the downloadQueue sorts `<TileInternal | Tileset>` which as a type would just make *everything* more awkward.

gkjohnson: Thanks! Yeah I agree -- I think it's okay to rework it in the long run I just want to keep it from breaking in the mean time.

gkjohnson:
Thanks again for the improvement! I'll try to publish a new version with this adjustment to npm again in the next day or two.

dbuck:
All good! I've got more customizations on top in a branch referenced directly by commit still, we've got a few rounds of iteration before anything gets near production ;)

src/base/TilesRendererBase.js
dbuck: Totally, I'm pretty heavy on types.. let that tooling catch as many errors as I can get it to catch ;)

We'll work it out over the next couple weeks I think as change requirements come up!

gkjohnson:
Sounds good -- I'll hold off on publishing for now, then, until we have some of the other features you're working on in

PR #194, opened 2021-07-07
Adjust calculateError use

### LINKED ISSUE #193: TilesRendererBase.calculateError: Adjust the function to not return anything
As of #189 both the tile distance and error are computed and set on the tile object in `calculateError` meaning it no longer needs to return a value or be [set on the tile object in the traverse functions](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/de17021e819a914e33de72961d45b92a7515dd2c/src/base/traverseFunctions.js#L148-L149)

gkjohnson:
fix #193 

**TODO**
- Depth debug visualization is not working

### DIFF
--- a/src/base/traverseFunctions.js
+++ b/src/base/traverseFunctions.js
@@ -146,8 +146,10 @@ export function determineFrustumSet( tile, renderer ) {
 	// at an external tile set.
 	if ( ( stopAtEmptyTiles || ! tile.__contentEmpty ) && ! tile.__externalTileSet ) {
 
-		const error = renderer.calculateError( tile );
-		tile.__error = error;
+		// compute the _error and __distanceFromCamera fields
+		renderer.calculateError( tile );
+
+		const error = tile.__error;
 		if ( error <= errorTarget ) {
 
 			return true;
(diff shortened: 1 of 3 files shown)


src/base/traverseFunctions.js
```diff
@@ -146,8 +146,10 @@ export function determineFrustumSet( tile, renderer ) {
 	// at an external tile set.
 	if ( ( stopAtEmptyTiles || ! tile.__contentEmpty ) && ! tile.__externalTileSet ) {
 
-		const error = renderer.calculateError( tile );
-		tile.__error = error;
+		// compute the _error and __distanceFromCamera fields
+		renderer.calculateError( tile );
+
+		const error = tile.__error;
```
dbuck, line 152: Ahh! makes sense.. Missed that one! I don't use DEPTH too often, I tend to just use RANDOM_COLOR and gauge the depth/lods based on the layout of the colors.

gkjohnson: Heh looks like I broke the depth-based debug coloring a bit ago 😬

PR #198, opened 2021-07-20
explicitly use path-browserify for better webpack 5+ support

### LINKED ISSUE #197: Remove dependency on path library
Projects that use webpack 5+ will display this error/warning

```
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "path": require.resolve("path-browserify") }'
        - install 'path-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "path": false }
```

Wondering whether it would make sense to remove the dependency in the library by explicitly using `path-browserify`.
Alternatively we could opt to write the code internally, but it would probably be a copy of the `path-browserify` functionality and require added maintenance.
The downside of using `path-browserify` is that is only supports POSIX paths

It would be good to get some thoughts on this.

Gmadges:
Fixes #197 

explicitly use `path-browserify` as a dependency to better support new versions of webpack (5+).

Under the hood older webpack builds would automatically polyfill node libraries with browser suitable alternatives like `path-browserify`

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "jest": "^25.5.4",
     "jest-cli": "^25.5.4",
     "parcel-bundler": "^1.12.4",
+    "path-browserify": "^1.0.1",
     "static-server": "^2.2.1",
     "three": ">=0.126.0",
     "typescript": "^3.9.9"
--- a/package.json
+++ b/package.json
@@ -48,12 +48,14 @@
     "jest": "^25.5.4",
     "jest-cli": "^25.5.4",
     "parcel-bundler": "^1.12.4",
-    "path-browserify": "^1.0.1",
     "static-server": "^2.2.1",
     "three": ">=0.126.0",
     "typescript": "^3.9.9"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -48,6 +48,7 @@
     "jest": "^25.5.4",
     "jest-cli": "^25.5.4",
     "parcel-bundler": "^1.12.4",
+    "path-browserify": "^1.0.1",
```
gkjohnson, line 51: This should be a regular dependency, right? Not a dev dependency?

gkjohnson:
Thanks for the quick PR!

package.json
Gmadges: yep, good shout!

gkjohnson:
Thank you!

PR #199, opened 2021-07-21
Supporting boundingSpheres for GeometricError calculations

Gmadges:
my first pass on supporting boundingSpheres for geometricError calculations, unsure if this fits with the current style of the code.

changes:
- cache inverse transform to be used later on (we currently only cache inverse box transform)
- fall back to sphere data if box data is unavailable (i.e use TransformInverse & boundingSphere)

I am yet to manually test with a tileset

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -786,6 +786,7 @@ export class TilesRenderer extends TilesRendererBase {
 			const boundingBox = cached.box;
 			const boxTransformInverse = cached.boxTransformInverse;
 			const transformInverse = cached.transformInverse;
+			const useBox = boundingBox && boxTransformInverse;
 
 			let maxError = - Infinity;
 			let minDistance = Infinity;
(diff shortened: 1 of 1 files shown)


gkjohnson (commented):
Hey! Generally this change looks good just one comment to address. Thanks for the contribution! Let me know once you've had a chance to test it and I'll look into merging this. If you have any example data sets that only use sphere bounding volumes that would be good to provide, as well. One of the reasons spheres haven't been supported here is I haven't found cases that would benefit, yet.

And depending on how deep into this you want to go I could see the `computeError` function being updated to take the maximum distance between both the sphere and box bounds to compute error so the most accurate value is used but I'd want to do some performance testing to see how that impacts the running of the update function.

Thanks again!

Gmadges:
Awesome thanks for the feedback. 
I'm still working on building a suitable public test dataset.

ulenspy:
See this [tileset](https://simplex3d.co.il/download/Naz_3Dtiles_example.zip) as example

gkjohnson:
Thanks @ulenspy! @Gmadges let me know when your happy with the PR and have tested it and I'll give it a go myself and we can get this merged.

gkjohnson:
Thanks!

PR #200, opened 2021-07-28
TilesRendererBase: Improve default priority behavior

gkjohnson:
Related to #143

- Change the priority queue sort function to prioritize loading tile geometry that is within the frustum. Previously tiles outside the frustum were being loaded first because the error for those tiles is set to infinity because it's never calculated.
- Change the priority function to use `__used` instead of `__lastFrameVisited`. `__lastFrameVisited` could cause issues when multiple tile sets were loaded and shared priority queues in case the number of updates calls (the max possible value of `_lastFrameVisited`) became out of sync between the renderer causing one tileset to always finish loading first.
- Handle case where bounding boxes have a dimension of 0 in one axis (mars data sky tiles)

cc @dbuck

### DIFF
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -19,6 +19,11 @@ const priorityCallback = ( a, b ) => {
 		// the lastFrameVisited tracks the last frame where a tile was used
 		return a.__lastFrameVisited - b.__lastFrameVisited;
 
+	} else if ( a.__inFrustum !== b.__inFrustum ) {
+
+		// prioritize loading whatever is in the frame
+		return a.__inFrustum ? 1 : - 1;
+
 	} else if ( a.__error !== b.__error ) {
 
 		// tiles which have greater error next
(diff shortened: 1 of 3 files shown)


src/base/TilesRendererBase.js
```diff
@@ -14,12 +14,17 @@ import { UNLOADED, LOADING, PARSING, LOADED, FAILED } from './constants.js';
  */
 const priorityCallback = ( a, b ) => {
 
-	if ( a.__lastFrameVisited !== b.__lastFrameVisited ) {
+	if ( a.__inFrustum !== b.__inFrustum ) {
```
dbuck, line 17: In the Tile interface in TilesRendererBase.d.ts: add __inFrustum, __used, and remove __lastFrameVisited

dbuck (commented):
Looks good to me! I think we should add it to the type interface also, if it is useful enough to be referenced by package users who may customize their own priority.

PR #204, opened 2021-08-09
Update the priority function (again)

gkjohnson:
cc @dbuck

After playing around with this a bit more in our internal projects I found that the priority sort function on master would cause the queue to prioritize tiles that the camera was inside the bounds of before anything else because in that case the distance is "0" and the error is "Infinity" for those tiles. This would lead to the closest parent tile displaying at lowest level of detail for a long time until all children loaded before it would "pop" to the target level of detail. This was particularly noticeable because we'd start the camera in the middle of the tileset close to the ground.

This PR adjusts the sort function so tiles at the shallowest depth are loaded first and within that depth the most visible / highest error tiles are loaded first within that set.

Getting a good priority function has turned out to be trickier than I expected 😁 I was trying to figure out how to keep the visual error consistent as data loaded so depth wouldn't have to be considered but I don't have anything concrete that comes to mind, yet.

### DIFF
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -14,26 +14,31 @@ import { UNLOADED, LOADING, PARSING, LOADED, FAILED } from './constants.js';
  */
 const priorityCallback = ( a, b ) => {
 
-	if ( a.__inFrustum !== b.__inFrustum ) {
+	if ( a.__depth !== b.__depth ) {
 
-		// prioritize loading whatever is in the frame
+		// load shallower tiles first
+		return a.__depth > b.__depth ? - 1 : 1;
+
+	} else if ( a.__inFrustum !== b.__inFrustum ) {
+
+		// load tiles that are in the frustum at the current depth
 		return a.__inFrustum ? 1 : - 1;
 
 	} else if ( a.__used !== b.__used ) {
 
-		// prioritize tiles that were used most recently
+		// load tiles that have been used
 		return a.__used ? 1 : - 1;
 
-	} if ( a.__error !== b.__error ) {
+	} else if ( a.__error !== b.__error ) {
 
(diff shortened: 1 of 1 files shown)


dbuck:
> Getting a good priority function has turned out to be trickier than I expected 😁

Truth ! -- I still think the ideal world is something that allows the end user to take into account **actual** occlusion culling.

I've been wanting to look into seeing what might be adaptable to webgl from the HZB culling styles in the nanite presentation with depth from previous frame reprojected + filling in gaps http://advances.realtimerendering.com/s2021/Karis_Nanite_SIGGRAPH_Advances_2021_final.pdf ~page 17

Not exactly the same from the lack of clusters, and requires me personally to level up a couple more times first, but intuitively it seems like the end-game if you have enough overdraw or complex enough shading (I certainly have the overdraw in larger interior scans, but shading is pretty minimal, so debatable for the utility keeps pushing that experiment further away !)

gkjohnson:
I think accounting for occlusion when sorting download priority might require some significant changes to the traversal algorithm considering the way sibling tiles are handled. For the priority queue I think looking into a way / option to map the current tile depth to the deepest depth of its children that meet the screen error criteria to a progress percent (`depth / deepestValidChildDepth`) and sort by that value might bee an interesting solution. It would still require taking sibling tiles into account, though.

Depending on what your bottleneck is you could look into the [WEBGL_multi_draw extension](https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_multi_draw) to reduce draw calls, a depth prepass to reduce expensive overdraw, or [webgl occlusion queries](https://developer.mozilla.org/en-US/docs/Web/API/WebGLQuery) to see what has or hasn't been visible in a previous frame. Compute shaders also might make something like the HZB approach a bit simpler but those won't be available until WebGPU is out 😬

PR #206, opened 2021-08-12
Support tilesets containing only bounding spheres (no bounding boxes)

justinmanley:
`TilesRenderer` methods `getBounds` and `getOrientedBounds` assume that a bounding box is always present. However, the 3D tiles spec [allows](https://docs.opengeospatial.org/cs/18-053r2/18-053r2.html#71) tilesets to specify either a box, sphere, or region as the bounding volume.

This PR converts bounding spheres into boxes during tile preprocessing if the tileset only specifies a bounding sphere, but not a box. Both the original sphere and the calculated box are retained.

This PR also adds some minimal test cases for the bounding volume calculations and moves around some code in the `TilesRenderer` to make it easier to test.


gkjohnson:
Hello! Thanks for the contribution! Generally I'd prefer to avoid creating derivative bounds in the preprocessor because doing something like creating a sphere from a box results in a much larger, looser bounding volume than necessary. Right now a sphere is only being derived from a box to facilitate frustum checks (which use the spheres at the moment) and support for checking frustum intersections with the oriented bounds has not been added (and probably will be slower?)

> Combined with #199, this PR should make it possible to display tilesets with bounding spheres using the DebugTilesRenderer.

Can you elaborate on why #199 will be needed in addition to this PR? This seems like a replacement for that one. 

Generally I think I prefer the approach in #199 that adds support for bounding spheres to error calculations so the original and more optimal bounds are used. I'm just waiting on that PR to be tested to merge it -- if you'd like to try testing #199 to make sure it works for you we can get that merged and revisit what else should change regarding these bounds.

Thanks again!

justinmanley:
@gkjohnson thanks for the quick reply!

I'm not attached to this approach, and I totally agree with what you're saying about the derived bounds being looser than necessary.

The reason I sent this PR in addition to #199 is that [this line ](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/DebugTilesRenderer.js#L395) in the DebugTilesRenderer currently breaks on sphere-only tilesets, even with #199 patched in.

If you'd prefer to keep derived bounds out of the preprocessor, I'm happy to add some guards to the DebugTilesRenderer instead so that the DebugTilesRenderer only generates bounds helpers for bounds which are actually present. All the examples which use [this pattern](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/433da49f5a89fd66a8d1206cb545850517749e43/example/index.js#L495) to move the tileset into view will also need to be updated to handle the case where there is no bounding box, only a sphere. I'm happy to make these changes if you think this approach is better.

Let me know which approach you prefer!

gkjohnson:
> The reason I sent this PR in addition to #199 is that this line in the DebugTilesRenderer currently breaks on sphere-only tilesets, even with #199 patched in.

Oh got it -- thanks for pointing it out. If that's the only thing missing from #199 I'll merge it and we can add anything else. I'll be able to test and merge #199 early next week.

> I'm happy to add some guards to the DebugTilesRenderer instead so that the DebugTilesRenderer only generates bounds helpers for bounds which are actually present

This is the change I'd prefer. Thanks for your help!

justinmanley:
Sounds good!

Just updated with the following changes:

- DebugTilesRenderer checks whether each type of bounds are present before displaying the associated helper
- Examples show how to bring sphere-only tilesets into the camera view
- TilesRenderer gains a `getBoundingSphere` method (analogous to `getBounds`) to support bringing sphere-only tilesets into the camera view

src/three/DebugTilesRenderer.js
```diff
@@ -372,11 +372,19 @@ export class DebugTilesRenderer extends TilesRenderer {
...
-			boxGroup.add( boxHelperGroup );
-			boxHelperGroup.updateMatrixWorld( true );
+			if ( boxHelperGroup ) {
 
-			sphereGroup.add( sphereHelper );
-			sphereHelper.updateMatrixWorld( true );
+				boxGroup.add( boxHelperGroup );
+				boxHelperGroup.updateMatrixWorld( true );
+
+			}
```
gkjohnson, line 380: We should also check above when removing the `boxHelperGroup` and `sphereHelper` to see if the objects are defined.

src/three/DebugTilesRenderer.js
```diff
@@ -392,40 +400,49 @@ export class DebugTilesRenderer extends TilesRenderer {
 				const scene = cached.scene;
 				if ( scene ) {
 
-					const cachedBox = cached.box;
-					const cachedBoxMat = cached.boxTransform;
```
gkjohnson, line 403: Some extra space here

gkjohnson (commented):
Thanks for the changes! I just merged #199 if you want to merge that into this branch to test. Otherwise a couple comments.

src/three/DebugTilesRenderer.js
justinmanley: Thanks for noticing. Done!

src/three/DebugTilesRenderer.js
justinmanley: Removed the extra space.

justinmanley:
Just rebased on master and pushed fixes for the two tweaks you pointed out.

src/three/DebugTilesRenderer.js
gkjohnson: Maybe something got lost in a rebase but I'm not seeing what you changed above -- I'm looking on lines 370 and 371 for a check to see if `boxHelperGroup` and `sphereHelper` got removed from the associated group

gkjohnson (commented):
Everything looks great! Just [this last item](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/206/files#r689151870) that maybe got lost somewhere and we should be ready to merge!

src/three/DebugTilesRenderer.js
justinmanley: ah, sorry! you're right. I added the checks to `disposeTile` but not to the if-branch of `setTileVisible`. Thanks for catching that! Should be good now.

PR #207, opened 2021-08-14
Fix for parsing Cesium Ion paths

PetterGs:
I wasn't able to run the Cesium default Photogrammetry tileset since it appears a url object is returned from the server instead of a string.

This token can be used temporarily for testing: **_eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJqdGkiOiJkN2UwMzFmOC1mY2E3LTQ1ZDQtYjg4NC1mZDhlMzZmZjk4NTkiLCJpZCI6MjU5LCJpYXQiOjE2Mjc5MzM4NTJ9.Rv6Icz_hBcEDbR160A3jPJQfhdZKWo_MJ6KovN7HRpw_**

AssetId: _**40866**_

### DIFF
--- a/example/ionExample.js
+++ b/example/ionExample.js
@@ -157,7 +157,7 @@ function reinstantiateTiles() {
 				url = new URL( json.url );
 				const version = url.searchParams.get( 'v' );
 
-				tiles = new TilesRenderer( url );
+				tiles = new TilesRenderer( url.toString() );
 				tiles.fetchOptions.headers = {};
 				tiles.fetchOptions.headers.Authorization = `Bearer ${json.accessToken}`;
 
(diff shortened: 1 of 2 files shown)


gkjohnson:
Thanks for the fix! I must have changed something and missed that it broke the ion example.

I think I'd like to keep the TilesRenderer class itself simple and continue to only expect a string as a url parameter. Instead I think we should change this line to convert the URL to a string before passing it into the TilesRenderer constructor:

https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/1d6d6881e2567e7fe2cb49d7d39c1abaa7aeb535/example/ionExample.js#L160

PetterGs:
No prob.

gkjohnson:
Great, thanks!

PR #208, opened 2021-08-21
Handle URL parameters and fragments.

### LINKED ISSUE #191: Add URL parameters support
Hi there!

I have an issue with external links, containing parameters in url.
For example, my tileset refers to tile by .../001.b3dm?x=34323.454&y=54322.3301
Looks like, the viewer recognizes its extension not as "b3dm" file, but as "b3dm?x=34323.454&y=54322.3301", and doesnt support it. 

See the attached screenshot:
image_2021_06_22T10_17_52_127Z

Tested on tileset: https://cloud.skylineglobe.com/sg/b3dm/NeveTzedek3.420160/tileset.json

PatrickNausha:
This is intended to fix #191 

I couldn't find any contributing guidelines for this repository, but am happy to conform to any expectations as needed. This solution uses the `URL` API which [will not work in Internet Explorer](https://caniuse.com/url). Is that acceptable?


src/base/TilesRendererBase.js
```diff
@@ -551,7 +553,7 @@ export class TilesRendererBase {
 						}
 
 						const uri = parseTile.content.uri;
-						const extension = uri.split( /\./g ).pop();
+						const extension = getUrlExtension( uri ) || '';
```
gkjohnson, line 556: Is there a specific scenario where this fallback to an empty string is necessary or is it defensive programming? If it's the latter I think I'd prefer to keep it as `null` because I think it will be more clear that something went wrong.

src/utilities/urlExtension.js
```diff
@@ -0,0 +1,41 @@
...
+			parsedUrl = new URL( 'https://example.com/' + url );
+
+		} catch ( _ ) {
+
+			// Ignore invalid URLs
+			return null;
+
+		}
+
+	}
```
gkjohnson, line 27: You can use the second argument of `URL` to specify the base url which will be used as a fallback if needed for the passed in URL stem so we won't have to use two try / catches:

```js
parsedUrl = new URL( url, 'http://fakehost.com/' );
```

src/utilities/urlExtension.js
```diff
@@ -0,0 +1,41 @@
...
+ * @returns {string} null if no extension found
+ */
+export function getUrlExtension( url ) {
+
+	let parsedUrl;
+	try {
+
+		parsedUrl = new URL( url );
+
+	} catch ( _ ) {
```
gkjohnson, line 13: Nit: If we're not using the caught error then we shouldn't have to include it in the catch statement. It should be enough to just use

```js
try {
  // ...
} catch {
  // ...
}
```

gkjohnson (commented):
Excellent! Thank you for the contribution and especially for adding some great tests! I added a few comments but overall it looks good. I assume you've been able to test this with your own tilesets? Do you happen to have one you can share that was breaking without this addition?

src/utilities/urlExtension.js
PatrickNausha: Today I learned: [Optional catch binding](https://github.com/tc39/proposal-optional-catch-binding).

Unfortunately it appears this project's eslint config won't parse that syntax. For the line containing
```js
	} catch {
```
I see
```
/(redacted)/3DTilesRendererJS/src/utilities/urlExtension.js
  13:10  error  Parsing error: Unexpected token {
```

Unless you object, I'll keep the `_` discard value for now and leave the eslint investigation/fix for a later PR.

src/utilities/urlExtension.js
PatrickNausha: Today I learned:

>  If url is an absolute URL, a given base will be ignored.

https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#parameters

src/base/TilesRendererBase.js
PatrickNausha: I had no specific scenario in mind. I'll switch this to `const extension = getUrlExtension( uri );`. My only goal here was to preserve the original behavior where `extension` is assigned a string so long as `uri` is a string.

PatrickNausha:
> Do you happen to have one you can share that was breaking without this addition?

The best thing I have to share is this munged version of `tileset-external.json` where I sprinkled URL params on the `uri` values: https://github.com/PatrickNausha/3DTilesRendererJS/blob/broken-tileset/example/data/tileset-external-with-url-params.json

src/utilities/urlExtension.js
gkjohnson: Oh okay good to know! I'll take a look at that another time. Thanks for trying it out.

gkjohnson:
> The best thing I have to share is this munged version of tileset-external.json where I sprinkled URL params on the uri values: https://github.com/PatrickNausha/3DTilesRendererJS/blob/broken-tileset/example/data/tileset-external-with-url-params.json

Oh great -- do you mind adding that to the example/data folder so it can easily be tested in the future?

PatrickNausha:
>  adding that to the example/data folder

Done. https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/208/commits/e4cce69e93671da741dbe19735ca3781f09a9670

gkjohnson:
Great, thank you again! I'll have this published soon

PR #210, opened 2021-08-24
Add option to use a custom color mode in DebugTilesRenderer

Rennzie:
The PR adds a `CUSTOM_COLOR_MODE` option for `DebugTilesRenderer`. It gives developers flexibility over and above the existing colour modes. 

It was born from a need to debug LODs, by colouring a LOD tile explicitly. The `DEPTH` and `RELATIVE_DEPTH` options are alright but lacking explicitness in colour (grey scale) and accuracy i.e a child tile may not be a LOD. To colour LODs I needed to add a property into the tilesetJSON (i.e `extras.lodLevel`). This introduces an unnecessary coupling between `3d-tiles-renderer` and design of the tiler. Using the more generic custom colour approach with a callback feels more useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -533,16 +533,16 @@ RANDOM_COLOR
 // Render every individual mesh in the scene with a random color.
 RANDOM_NODE_COLOR
 
-// Sets a custom color using the customDebugColor call back. 
+// Sets a custom color using the customColorCallback call back. 
 CUSTOM_COLOR_MODE
 ```
-### .customDebugColor
+### .customColorCallback
 
-```
-customDebugColor: (tile: Tile, child: Object) => void
+```js
+customColorCallback: (tile: Tile, child: Object3D) => void
 ```
 
-The callback used if `debugColor` is set to `CUSTOM_COLOR_MODE`. Value default to `null` and must be explicitly set
+The callback used if `debugColor` is set to `CUSTOM_COLOR_MODE`. Value defaults to `null` and must be set explicitly.
 
 ### .displayBoxBounds
 
--- a/README.md
+++ b/README.md
@@ -532,7 +532,17 @@ RANDOM_COLOR
 
 // Render every individual mesh in the scene with a random color.
 RANDOM_NODE_COLOR
+
+// Sets a custom color using the customDebugColor call back. 
+CUSTOM_COLOR_MODE
 ```
+### .customDebugColor
+
+```
+customDebugColor: (tile: Tile, child: Object) => void
+```
+
+The callback used if `debugColor` is set to `CUSTOM_COLOR_MODE`. Value default to `null` and must be explicitly set
 
 ### .displayBoxBounds
 
--- a/src/three/DebugTilesRenderer.js
+++ b/src/three/DebugTilesRenderer.js
@@ -17,6 +17,7 @@ export const RELATIVE_DEPTH = 5;
 export const IS_LEAF = 6;
 export const RANDOM_COLOR = 7;
 export const RANDOM_NODE_COLOR = 8;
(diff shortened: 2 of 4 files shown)


src/three/DebugTilesRenderer.js
```diff
@@ -344,6 +345,20 @@ export class DebugTilesRenderer extends TilesRenderer {
...
 						}
+						case CUSTOM_COLOR_MODE: {
+
+							if ( this.customDebugColor ) {
+
+								this.customDebugColor( tile, c );
+
+							} else {
+
+								console.error( 'customDebugColor callback not defined' );
```
gkjohnson, line 356: Nit: Lets change this to a warning and prefix it with the class so it's more clear where the error is coming from:

```js
console.error( 'DebugTilesRenderer: customDebugColor callback not defined' );
```

README.md
```diff
@@ -532,7 +532,17 @@ RANDOM_COLOR
...
+// Sets a custom color using the customDebugColor call back. 
+CUSTOM_COLOR_MODE
+```
+### .customDebugColor
+
 ```
+customDebugColor: (tile: Tile, child: Object) => void
+```
+
+The callback used if `debugColor` is set to `CUSTOM_COLOR_MODE`. Value default to `null` and must be explicitly set
```
gkjohnson, line 545: Can we call the field `customColorCallback` so it's more clear that it's a callback to be set. And thanks for updating the documentation!

README.md
```diff
@@ -532,7 +532,17 @@ RANDOM_COLOR
...
 // Render every individual mesh in the scene with a random color.
 RANDOM_NODE_COLOR
+
+// Sets a custom color using the customDebugColor call back. 
+CUSTOM_COLOR_MODE
+```
+### .customDebugColor
+
 ```
+customDebugColor: (tile: Tile, child: Object) => void
```
gkjohnson, line 542: `child` should have type `Object3D` and lets use the `js` extension for coloring the code block:

````
```js
code here
```
````

gkjohnson (commented):
Some small comments but overall it looks good to me!

cc @dbuck if I recall you might have mentioned a similar desire to add custom coloring.

src/three/DebugTilesRenderer.js
Rennzie: Oh yes, that's a lot nicer!

README.md
Rennzie: Ah thank, I wasn't 100% sure on the type here.

Rennzie:
Thanks @gkjohnson. I've made all the changes you requested. Please let me know if you spot anything else.

gkjohnson:
Great, thank you! I think the last thing that might be a nice addition would be updating the main "kitchen sink" index.js example to demonstrate how to use this and make it easier to test in the future.

You can add the new dropdown option [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/example/index.js#L247-L259) and set the custom color callback in the [reinstantiateTiles](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/example/index.js#L81-L106) function. Something simple like coloring every odd mesh red and every even mesh black would be a good enough demo of the new feature. Or if you have a more fun idea feel free to add that!

Let me know if that sounds okay to you and we can get this merged!

Rennzie:
> Let me know if that sounds okay to you and we can get this merged!

Sounds good. I'll get that done ASAP

Rennzie:
All done. Couldn't think of a better example so stuck with yours. 😀

gkjohnson:
Excellent -- thank you so much! I'll plan to make a new release once #208 is merged

Issue #213, opened 2021-09-16
Immersive VR tile loading (WebXR)


phoenixbf:
3D Tiles progressive loading in the VR example seem not to trigger during exploration using WebXR-enabled browsers (e.g. built-in Oculus Quest browser or mobile Chrome). While it works fine on Firefox Reality (Oculus Quest) browser

gkjohnson:
Can you provide more information on the issue? If you're able to provide a fix I'd be happy to take a PR. I can only test with Firefox Reality at the moment and don't have the bandwidth to debug the example more deeply.

phoenixbf:
The issue should be replicable with the example: https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/vr.html

Basically once the VR session starts, tiles do not switch (at all) according to camera location (tested on Oculus Browser, WebXR on android Chrome mobile) - while Firefox Reality seems to correctly manage the virtual camera, thus loading children as we move around.
I also report that it's working fine using WebXR API emulator (Chrome, windows 10) - quite strange.

It is also unclear to me in the example why all tileset cameras are cleared and reset on each render (line 240): shouldnt be sufficient to set/replace it once, when the VR presentation starts?

So far I tried different ways to (re)set virtual camera in XR sessions, but no luck.
I'm testing several options, once I find a solution I'll post here

gkjohnson:
> Basically once the VR session starts, tiles do not switch (at all) according to camera location (tested on Oculus Browser, WebXR on android Chrome mobile) - while Firefox Reality seems to correctly manage the virtual camera, thus loading children as we move around.

You can log transform of the VR camera (used [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/5637a7824d40f64dc10cfdb8f8a34e9f7b4ab08c/example/vr.js#L245)) and if it doesn't update as you move your head then that seems like the core of the issue. It's possible this could be a three.js issue if not a browser issue, then. I had also tested with desktop Chrome and the Chrome emulator previously without issue, as well. Just not other mobile browsers. You may want to try with the latest three.js, as well.

> It is also unclear to me in the example why all tileset cameras are cleared and reset on each render (line 240): shouldnt be sufficient to set/replace it once, when the VR presentation starts?

[...]

PR #216, opened 2021-11-01
custom scheduling callback for PriorityQueue

phoenixbf:
customizable scheduling callback for PriorityQueue allowing correct handling when a WebXR session is active (see #213 ), defaults to requestAnimationFrame

### DIFF
--- a/example/vr.js
+++ b/example/vr.js
@@ -42,6 +42,8 @@ let box, sphere, grid;
 let raycaster, fwdVector, intersectRing;
 let offsetParent;
 let controller, controllerGrip;
+let xrSession = null;
+let tasks = [];
 
 let params = {
 
@@ -52,7 +54,6 @@ let params = {
 };
 
 init();
-animate();
 
 function init() {
 
@@ -68,6 +69,8 @@ function init() {
 
 	document.body.appendChild( renderer.domElement );
 	renderer.domElement.tabIndex = 1;
+
+	renderer.setAnimationLoop( frame );
 
 	// create workspace
 	workspace = new Group();
@@ -102,11 +105,15 @@ function init() {
 	tiles = new TilesRenderer( '../data/tileset.json' );
 	offsetParent.add( tiles.group );
 
+	// We set camera for tileset
+	tiles.setCamera( camera );
+	tiles.setResolutionFromRenderer( camera, renderer );
+
 
 	// We define a custom scheduling callback to handle also active WebXR sessions
 	const tilesSchedulingCB = func => {
 
-		setTimeout( func, 50 );
+		tasks.push( func );
 
 	};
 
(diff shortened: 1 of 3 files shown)


lojjic:
I like the flexibility of this solution, for enabling XR of course but also for integrating with other custom task scheduling systems.

The new hook should also be added to the documentation (README.md and PriorityQueue.d.ts). I'm happy to assist with that if you want.

gkjohnson:
Looks good! Like you and @lojjic mentioned I think we should:

- update the README
- update the VR example to properly demonstrate how to set up the tiles renderer for use in VR (adjust the schedulingCallback member)
- update the d.ts files

Thank you fo the fix!

phoenixbf:
I also updated the VR example. It can be further optimized (and simplified a bit) by setting up tileset-camera only once, when WebXR session starts (see #213 )

example/vr.js
```diff
@@ -102,6 +102,19 @@ function init() {
 	tiles = new TilesRenderer( '../data/tileset.json' );
 	offsetParent.add( tiles.group );
 
+
+	// We define a custom scheduling callback to handle also active WebXR sessions
+	const tilesSchedulingCB = func => {
+
+		setTimeout( func, 50 );
+
+	};
```
gkjohnson, line 111: Thanks for updating the example! In order to best support the XR loop we'd have to do something like so, right?

```js
let _xrHandle = - 1;
let _rafHandle = - 1;
const tilesSchedulingCB = func => {

  if ( renderer.xr.isPresenting ) {

    const session = renderer.xr.getSession();
    session.cancelAnimationFrame( _xrHandle );
    _xrHandle = session.requestAnimationFrame( func );

  }

  cancelAnimationFrame( _rafHandle );
  _rafHandle = requestAnimationFrame( func );

};
```

I think this would be the most ergonomic way to support both rAF implementations. Thoughts @lojjic?

gkjohnson (commented):
I also notice that we're using the global `requestAnimationFrame` in the "animate" function:

https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/9d7e759ee8d17795d6906ba4245841aae4686fff/example/vr.js#L228-L232

I wonder why this seems to work in this case when rAF seemed to not work in the priority queues? Maybe there's special handling in browsers for existing "chained" rAF calls when an xr session is started? Should we change the example to use `renderer.setAnimationLoop` like three.js recommends for vr?

phoenixbf:
> I also notice that we're using the global `requestAnimationFrame` in the "animate" function:
> 
> https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/9d7e759ee8d17795d6906ba4245841aae4686fff/example/vr.js#L228-L232
> 
> I wonder why this seems to work in this case when rAF seemed to not work in the priority queues? Maybe there's special handling in browsers for existing "chained" rAF calls when an xr session is started? Should we change the example to use `renderer.setAnimationLoop` like three.js recommends for vr?

Yes, indeed.
We should refactor and change it to `renderer.setAnimationLoop( );`. This could be also a good opportunity to simplify the camera management for standard and WebXR sessions. I'll try this afternoon to provide a revisited VR example

example/vr.js
lojjic: Hmm it may be tricker than that; I think it may result in an orphaned rAF when the user switches from one mode to the other...? 🤔 

The `renderer.xr` looks like it fires 'sessionstart' and 'sessionend' events which you could possibly use to move any pending callbacks from one rAF over to the other.

Or maybe utilize `setAnimationLoop` in a queue/flush setup:

```js
let callbacks = [];

const tilesSchedulingCB = func => {
  callbacks.push(func);
};

threeRenderer.setAnimationLoop((...args) => {
  if (callbacks.length) {
    const funcs = callbacks;
    callbacks = [];
    funcs.forEach(cb => cb(...args));
  }
});
```

phoenixbf: Ok guys, so after the `setTimeout` setup for the VR example (that works fine), I also tested the **rAFs** approach, but without luck for active WebXR session - also it is a bit tricky like @lojjic said when switching between default and XR.

So I'm proposing something similar to the suggested callbacks (pending tasks consumed in the `setAnimationLoop`) for the VR example - that I think is simpler, and works fine for both standard and WebXR sessions (tested on different devices).

Furthermore if we find something even better in the future, it is an easy change in `tilesSchedulingCB`

gkjohnson: > I think it may result in an orphaned rAF when the user switches from one mode to the other...?

Thanks it occurred to me that this may be an issue a bit after I posted. Presumably there's a good reason why the WebXR rAF works independently from the global rAF this way otherwise this seems like a very awkward API quirk to work around.

The only other option that comes to mind is adding an "autoUpdate" option to the `PriorityQueue` that can be used to disable the auto scheduling requiring the user to trigger it themselves:

```js
tiles.downloadQueue.autoUpdate = false;
tiles.parseQueue.autoUpdate = false;

// ...

function animate() {

  tiles.downloadQueue.tryRunJobs();
  tiles.parseQueue.tryRunJobs();

  // ...

}
```

But I think we can switch to that another time if necessary. Unless you guys have other thoughts on the above I'll go ahead and test the PR over the next couple days and get this merged / published!

PR #217, opened 2021-11-01
Calculate SSE directly from projection matrix rather than specific Camera subclass properties

lojjic:
Our app uses a custom Camera implementation where we control its projection matrix directly, changing its FOV and even transitioning between perspective and orthographic projections on the fly. But the existing SSE calculation uses the logical config properties of `PerspectiveCamera` and `OrthographicCamera` subclasses, which are incorrect for our camera.

This PR tweaks the SSE logic to look directly at each camera's `projectionMatrix` instead of assuming particular Camera subclasses, based on the following characteristics from how Three's cameras construct their projections:

- Orthographic projections have `mat[ 15 ] === 1`
- Perspective FOV and ortho width/height are encoded into `mat[0]` and `mat[5]`

I've been running this for a while with a `oldResult === newResult` assertion in there to validate that the end behavior is equivalent. Let me know if there's a better way to prove correctness for this PR.

Some lesser-used properties like [PerspectiveCamera.zoom](https://threejs.org/docs/?q=perspe#api/en/cameras/PerspectiveCamera.zoom) should now be taken into account too.

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -10,14 +10,12 @@ import {
 	Sphere,
 	Vector3,
 	Vector2,
-	Math as MathUtils,
 	Frustum,
 	LoadingManager
 } from 'three';
 import { raycastTraverse, raycastTraverseFirstHit } from './raycastTraverse.js';
 
 const INITIAL_FRUSTUM_CULLED = Symbol( 'INITIAL_FRUSTUM_CULLED' );
-const DEG2RAD = MathUtils.DEG2RAD;
 const tempMat = new Matrix4();
 const tempMat2 = new Matrix4();
 const tempVector = new Vector3();
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -391,17 +391,24 @@ export class TilesRenderer extends TilesRendererBase {
 
 			}
 
+			// Read the calculated projection matrix directly to support custom Camera implementations
 			const projection = camera.projectionMatrix.elements;
+
+			// The last element of the projection matrix is 1 for orthographic, 0 for perspective
 			info.isOrthographic = projection[ 15 ] === 1;
 
 			if ( info.isOrthographic ) {
 
+				// See OrthographicCamera.updateProjectionMatrix and Matrix4.makeOrthographic:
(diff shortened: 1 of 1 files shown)


gkjohnson:
Great, thank you! So I'm understanding that you're not using an off-axis or otherwise unique projection matrix from what OrthographicCamera and PerspectiveCamera can produce, right? You're just modifying the `projectionMatrix` fields?

This change sounds fine to me I'd just like to make the matrix element use more followable. Can we add a comment or clear variable name indicating which row / column the matrix element is from and what each element means (if possible)? It's not super clear when looking at the elements array right now. And lastly do you have any references you used for understanding the specifics of the right way to do this? It's been awhile since I've dug into the details of projection matrices.

> Some lesser-used properties like PerspectiveCamera.zoom should now be taken into account too.

Didn't realize this was broken! This seems like a more flexible way to handle the error calculations.

lojjic:
Good call, I'll add some comments.

That's correct we're not doing anything weird with the projections, they're still standard perspective/ortho projections like those subclasses would produce, just done without their help.

As for references for deeper understanding, I'll admit I didn't go that deep. This was really more of an algebra exercise, looking at the Three.js source code and tracing where those fov/etc. parameters that _were_ being used end up in the projection matrices, then working backwards from there to produce equivalent SSE values from the matrix.

Relevant Three.js code:
* [PerspectiveCamera.updateProjectionMatrix](https://github.com/mrdoob/three.js/blob/88b6328998d155fa0a7c1f1e5e3bd6bff75268c0/src/cameras/PerspectiveCamera.js#L177)
* [OrthographicCamera.updateProjectionMatrix](https://github.com/mrdoob/three.js/blob/88b6328998d155fa0a7c1f1e5e3bd6bff75268c0/src/cameras/OrthographicCamera.js#L84)
* [Matrix4.makePerspective/makeOrthographic](https://github.com/mrdoob/three.js/blob/88b6328998d155fa0a7c1f1e5e3bd6bff75268c0/src/math/Matrix4.js#L772)

gkjohnson:
Okay thanks for updating the comments! Looks good to me after taking a look at those functions.

I'll plan to publish a new version once the other VR PR is handled.

PR #223, opened 2021-11-16
Examples: Add VR Mars Demo

gkjohnson:
Switch the VR demo to use Mars terrain _or_ create a separate VR Mars demo.

**TODO**
- [ ] Rotate the container group
- [ ] Decide whether a separate demo should be create from the basic VR demo. Or create custom tile set with external data pointing to both
  - Create a tileset that points to both the external sky and ground tilesets instead of creating two tileset renderers manually
- [ ] Test


gkjohnson:
Closing in favor of #228

Issue #185, opened 2021-06-03
Starting discussion on future extension support?


dbuck:
I was checking out some of the recent changes in the [`3d-tiles-next` branch here](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next), and it brought up some thoughts about future extensibility. 

First item I was interested in the ability to just use GLTF as the content within tiles via [3DTILES_content_gltf](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_content_gltf/0.0.0)
-> Replace the B3dm wrapper with gltf.

gkjohnson:
I think being able to register an extension that uses certain predefined hooks to preprocess tiles like seems like a good design in the long run. It looks like [GLTFLoader lets you register a full class that defines the hooks you'd need](https://github.com/mrdoob/three.js/blob/master/examples/jsm/loaders/GLTFLoader.js#L78-L112), but I'm not super familiar with the details of that code. For some of the more basic extensions I think it makes sense to just support them by default, though, which is probably the easier place to start. Were there any other extensions you were looking at using? What kinds of hooks would they wind up needing?

For GLTF and GLB support I'd be okay with just adding those in the `TilesRenderer.parseTile` function. Looking at the Cesium implementation it seems they [already load GLB and GLTF files](https://github.com/CesiumGS/cesium/blob/dacc86c6a32f1ece62d66ac20b6ea5ff6f040ad0/Source/Scene/Cesium3DTileContentFactory.js#L86-L99) for the 3d tiles as well as others.

It probably also makes sense to start logging unsupported extensions so it's more clear what functionality is being skipped or might be broken when loading a tileset with an extension -- though I'm not sure how many of those are out in the wild yet.

Rennzie:
Hey @gkjohnson had you given this much more thought? We're looking at using the upcoming gltf extension `EXT_mesh_features`. Possibly something we could help with next year.

gkjohnson:
Hi @rennzie! #230 by @dbuck adds support for the `3DTILES_content_gltf` extension which will enable use of GLTF and GLB files in the tileset json files. That PR is nearly done and should be ready to merge after a couple more changes.

[...]

PR #225, opened 2021-11-18
[types] - more complete types for the 3d-tiles spec tiles

dbuck:
While considering a couple options for 3d-tiles extensions for #185 I wanted to fix up some types to track the expected tileset schema for where extensions blobs could be.

* Adds Tileset / TileBase interface to track the 3d-tiles spec schema.
* Adds Tile interface which is exposed in callbacks using during loading, includes specific properties used in overridable functions (Ex: Priority functions).

And: Questionable usefulness without actually using it within the repo ->
* Adds TileInternal type for the internally used tile properties.


dbuck:
I also updated the referenced three version in the package.json, this is debatable, it shouldn't affect any usage for people using the package directly/normally.

I found I was having type conflicts for types which extend three types when referencing the repo locally using `yarn link` in order to work on the version used in my project directly. Updating the three version in the package.json to a higher rev, after they had split the typedefs out of the main repo, and into @types/three cleared up my local problems.

I chose 128 simply because that's the version we've been using for some number of weeks without any problems ! It could also update to the 134 without any issue, but I've personally not done as much testing against that version yet !

In testing out local examples, everything works as expected except for the `pntsExample.html`, which is trying to reference an example data set from the cesium repo which does not seem to exist anymore? cc: @gkjohnson

src/base/TileInternal.d.ts
```diff
@@ -0,0 +1,35 @@
+import { Tile } from './Tile';
+
+/**
+ * Internal state used/set by the package.
+ */
+
+export interface TileInternal extends Tile {
```
gkjohnson, line 7: We can keep this for now -- might make a good reference? If it looks like it won't be useful we can remove it later.

gkjohnson (commented):
Looks good! Just a few changes to make.

gkjohnson:
Excellent! Looks great, thank you. I'll wait to publish since it sounded like you had some other stuff in the works if I recall correctly.

Thanks again!

Issue #232, opened 2021-12-08
Add tile visibility change callback for TilesRenderer


ligaofeng0901:
For now, there are some callback functions like `onLoadTileSet`、`onLoadModel`、`onDisposeModel` in TilesRenderer. In some situations, I wanna add a callback when a tile's visiblity changes, I think it can be implemented by adding some codes in function `setTileVisible` of `TilesRenderer`, just right here
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/TilesRenderer.js#L781
add code just like
`
if (this.onTileVisibleChanged) {
        this.onTileVisibleChanged(tile, visible);
}
`

gkjohnson:
Hello @ligaofeng0901! Can you elaborate on your use case? Why do you want a visibility callback?

ligaofeng0901:
Ok @gkjohnson . I load a big scene of city by 3d tiles. In each tile, there are some traffic lights. When a tile is loaded, I put these traffic lights information into a Map object, and create another mesh to present the state of a traffic light. Then in each frame, I will update the state of every traffic light. There are more than ten thousands traffic lights in the whole 3d tiles. I only want to update these visible traffic light. I think it's helpful If there is a visiblity callback. 
For now, I rewrite the function `setTileVisible` of `TilesRenderer`  by `TilesRenderer.prototype.setTileVisible = function (tile, visible) {...}`, it's a littile bit tricky. I think a visiblity callback is useful.

dbuck:
We have a similar use case, we use it narrow down the number of meshes we update some computation on to only the meshes from visible tiles.

So far we've been doing it via monkeypatched wrapper functions tracking active and visible, with an intention to work out a more appropriate way to handle it within the package.

gkjohnson:
Gotcha -- thanks for the explanations!

A couple possible options that come to mind:

1. Making the `tilesRenderer.visibleTiles` public so users can access the set of currently visible tiles like so: `tilesRenderer.visibleTiles.forEach( tile => { ... } )`. Though they'd have to access `tile.cached.scene`, instead, which isn't great.

2. Add a `tilesRenderer.forEachVisibleModel( callback )` that takes a `scene` and `tile` arguments like `forEachLoadedModel`.

[...]

PR #233, opened 2021-12-15
Add tile visibility callback

ligaofeng0901:
Related #232 
Add a property named `onTileVisibilityChange ` to `TilesRenderer `, so we can track the visibility of each tile. If it's ok and if necessary, I can add the api intro into readme file.


gkjohnson:
This looks good to me! Just need to fix the lint error and update documentation.

gkjohnson:
Looks great -- thank you!

PR #246, opened 2022-03-26
Used wrong parameter when getData function call another function.

jonnxie:
componentType parameter should passed first.


gkjohnson:
Thanks!

Issue #255, opened 2022-06-01
Absolute paths in content.uri are not supported


usefulthink:
A tileset hosted at `https://example.tld/tiles/root/tileset.json` that contains content-uris like `/tiles/content/abcdef` will try to load `https://example.tld/tiles/root/tiles/content/abcdef` instead of the correct `https://example.tld/tiles/content/abcdef`. 

This is caused by the `urlJoin()` function not handling absolute paths correctly.

PR #256, opened 2022-06-01
fix: improve content-uri handling

usefulthink:
The old implementation didn't handle absolute paths properly (#255), so
this is an attempt at solving that issue.

It turned out that the `urlJoin` utility-function was only used in a
single place, where it could be replaced with the URL constructor, so I
removed both the utility-function as well as its tests.

That left only a single call to `path.basename` remaining as the reason
to have the `path-browserify` dependency, so I replaced that as well and
removed the dependency.


usefulthink:
should I also commit the `package-lock.json`? That seems to have a lot of unrelated changes, probably caused by different npm-versions.

gkjohnson:
> should I also commit the package-lock.json?

Lets not -- I'll update when I get a chance.

gkjohnson:
Thanks! I'll take a look at your other PR when I have a moment.

PR #257, opened 2022-06-01
fix: detect content type from mime-type and magic bytes

### LINKED ISSUE #254: content loading relies on filename-extensions
We're currently working with a massive hosted tileset we don't control that doesn't use filename-extensions in it's paths.

[According to the specification](https://github.com/CesiumGS/3d-tiles/tree/main/specification#file-extensions-and-mime-types) this should be fine (and in fact, Cesium can use the tileset without problem), but `TilesRenderer.parseTile()` relies of the filename extension to determine the type of content received.

I think this could best be adressed by determining the file type from the content-type header of the response or from the magic-bytes for binary formats and ignoring the filename extension entirely.

usefulthink:
Removes the reliance on file-extensions to determine the content type and instead use the mime-type or magic bytes from the buffer to detect the format.

fixes #254

### DIFF
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -550,12 +550,16 @@ export class TilesRendererBase {
 
 					if ( res.ok ) {
 
-						return {
+						return res.arrayBuffer().then( buffer => {
 
-							mediaType: res.headers.get( 'Content-Type' ),
-							buffer: res.arrayBuffer()
+							return {
 
-						};
+								mediaType: res.headers.get( 'Content-Type' ),
+								buffer
+
+							};
+
+						} );
 
 					} else {
 
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -550,16 +550,7 @@ export class TilesRendererBase {
 
 					if ( res.ok ) {
 
-						return res.arrayBuffer().then( buffer => {
-
-							return {
-
-								mediaType: res.headers.get( 'Content-Type' ),
-								buffer
-
-							};
-
-						} );
+						return res.arrayBuffer();
 
 					} else {
 
@@ -568,7 +559,7 @@ export class TilesRendererBase {
 					}
 
 				} )
-				.then( result => {
+				.then( buffer => {
 
 					// if it has been unloaded then the tile has been disposed
 					if ( tile.__loadIndex !== loadIndex ) {
@@ -576,8 +567,6 @@ export class TilesRendererBase {
 						return;
 
 					}
-
-					const { buffer, mediaType = '' } = result;
 
 					stats.downloading --;
 					stats.parsing ++;
@@ -593,7 +582,7 @@ export class TilesRendererBase {
 
 						}
 
-						const contentType = getTileContentType( buffer, mediaType );
+						const contentType = getTileContentType( buffer );
 
 						return this.parseTile( buffer, parseTile, contentType );
 
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -3,6 +3,7 @@ import { LRUCache } from '../utilities/LRUCache.js';
 import { PriorityQueue } from '../utilities/PriorityQueue.js';
 import { determineFrustumSet, toggleTiles, skipTraversal, markUsedSetLeaves, traverseSet } from './traverseFunctions.js';
 import { UNLOADED, LOADING, PARSING, LOADED, FAILED } from './constants.js';
+import { getTileContentType } from "../utilities/tileContentType";
 
 /**
  * Function for provided to sort all tiles for prioritizing loading/unloading.
@@ -166,7 +167,7 @@ export class TilesRendererBase {
 	}
 
 	// Overrideable
-	parseTile( buffer, tile, extension ) {
+	parseTile( buffer, tile, tileContentType ) {
 
 		return null;
 
@@ -549,7 +550,12 @@ export class TilesRendererBase {
 
 					if ( res.ok ) {
 
-						return res.arrayBuffer();
+						return {
+
+							mediaType: res.headers.get( 'Content-Type' ),
+							buffer: res.arrayBuffer()
+
+						};
 
 					} else {
 
@@ -558,7 +564,7 @@ export class TilesRendererBase {
 					}
 
 				} )
-				.then( buffer => {
+				.then( result => {
 
 					// if it has been unloaded then the tile has been disposed
 					if ( tile.__loadIndex !== loadIndex ) {
@@ -566,6 +572,8 @@ export class TilesRendererBase {
 						return;
 
 					}
+
+					const { buffer, mediaType = '' } = result;
 
 					stats.downloading --;
 					stats.parsing ++;
@@ -581,10 +589,9 @@ export class TilesRendererBase {
 
 						}
 
-						const uri = parseTile.content.uri;
-						const extension = getUrlExtension( uri );
+						const contentType = getTileContentType( buffer, mediaType );
 
-						return this.parseTile( buffer, parseTile, extension );
+						return this.parseTile( buffer, parseTile, contentType );
 
 					} );
 
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -167,7 +167,7 @@ export class TilesRendererBase {
 	}
 
 	// Overrideable
-	parseTile( buffer, tile, tileContentType ) {
+	parseTile( buffer, tile, extension, tileContentType ) {
 
 		return null;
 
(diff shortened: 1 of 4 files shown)


usefulthink:
I converted this to a draft for now, as I'm still testing it out.

One thing to note: I had a hard time wrapping my head around all these promise-chains so I converted part of the code to using async/await. Would you be open to that in general? Otherwise I can convert it back when it actually works.

gkjohnson:
> If we don't rely on the file-extension to determine the content-type, we will also not know in advance if the referenced content is an external tileset or actual tile-contents. I have to look at this a bit more to see how that could be done and what implications that might have.

To start I think it's okay if we just rely on the file extension to check for the "json" extension and assume it's an external tileset. Then in a future PR we can change the way this is handled if needed. Do we know how Cesium handles this? Does it load the buffer and check if the first token is `{` in order to determine whether it's an external tileset json?

> I had a hard time wrapping my head around all these promise-chains so I converted part of the code to using async/await. Would you be open to that in general? Otherwise I can convert it back when it actually works.

I'm not necessarily against it but I think I'd prefer doing that in a separate PR to avoid this one becoming too large. I don't think there should be any issue in using asnyc / await but I will want to take a close look during the conversion to make sure no behavior has meaningfully changed.

--

In terms of the PR I'm not sure if we need both support for detecting "mediaType" from the request header _and_ support for detecting from magic bytes. The glTF file format includes a magic bytes token in the header, as well, so that should be enough on it's own to be reliable and should generally be simpler. 

I also think the `parseTile` function can be adjusted to take the "magicBytes" as an ASCII string in addition to the "extension": `parseTile( buffer, tile, extension, mediaType )`. That function can be overridden to support user-specified formats so I think it's useful to include the extension, too.

usefulthink:
> To start I think it's okay if we just rely on the file extension to check for the "json" extension and assume it's an external tileset. Then in a future PR we can change the way this is handled if needed.

Alright, sounds good. I will revert this PR to where it was before starting to mess with loading external tilesets and start a seperate PR for other things.

> Do we know how Cesium handles this? Does it load the buffer and check if the first token is `{` in order to determine whether it's an external tileset json?

Yeah, so my solution is heavily inspired by how cesium handles this – all contents are downloaded as ArrayBuffer and converted to json if there's no matching magic-bytes (doesn't even check if it actually looks like json): https://github.com/CesiumGS/cesium/blob/c4946f24a7419a3c6e28f297533bb573096c352f/Source/Scene/preprocess3DTileContent.js

> I'm not necessarily against it but I think I'd prefer doing that in a separate PR to avoid this one becoming too large. I don't think there should be any issue in using asnyc / await but I will want to take a close look during the conversion to make sure no behavior has meaningfully changed.

That makes perfect sense, let me see if I find the time to tackle that as well. But I guess I will first go on with my experiments to see if we can get usable results with this library and our global tileset.

> In terms of the PR I'm not sure if we need both support for detecting "mediaType" from the request header and support for detecting from magic bytes.

My thoughts as well – I think I will remove the mediaType, it doesn't really add much value..

> I also think the parseTile function can be adjusted to take the "magicBytes" as an ASCII string in addition to the "extension": `parseTile( buffer, tile, extension, mediaType )`. That function can be overridden to support user-specified formats so I think it's useful to include the extension, too.

With that in mind, shouldn't we pass in the full URL instead of just the extension? It's less work for the library would allow overriding classes to also make decisions based on other parts of the path.

gkjohnson:
> `parseTile( buffer, tile, extension, mediaType )`

Oops - I meant for this to be `parseTile( buffer, tile, extension, magicBytes )` instead.

> With that in mind, shouldn't we pass in the full URL instead of just the extension.

Continuing to pass the extension in here is backwards compatible so unless there's a good use case that comes up that requires the full url I think I'd prefer to just stick with the extension for now.

src/base/TilesRendererBase.js
```diff
@@ -557,10 +559,10 @@ export class TilesRendererBase {
 
 						}
 
-						const uri = parseTile.content.uri;
+						const contentType = getTileContentType( buffer );
 						const extension = getUrlExtension( uri );
```
usefulthink, line 563: this is actually a change in behaviour: this now uses the result of preprocessURL to get the extension instead of the original url. Which one is more correct?

usefulthink:
I didn't quite get the last point with passing the magic bytes to the parseTile function. 

 - JSON-based formats will not have those, should we use a value like "JSON" for this?
 - should that be the Uint32-value or the string-representation? And should we also return unknown values?

src/utilities/tileContentType.js
```diff
@@ -0,0 +1,46 @@
...
+};
+
+// first four bytes ('magic bytes') in big-endian uint32-format
+const MAGIC_BYTES = {
+	0x6233646d: TileContentType.B3DM, // 'b3dm'
+	0x6933646d: TileContentType.I3DM, // 'i3dm'
+	0x706e7473: TileContentType.PNTS, // 'pnts'
+	0x636d7074: TileContentType.CMPT, // 'cmpt'
+	0x676c5446: TileContentType.GLB // 'glTF'
+};
```
gkjohnson, line 20: I think this is a lot of indirection and in the long run won't let users support their own formats with magic bytes like PLY, etc if they choose to use them. I think we can just convert the magic bytes to a string and pass that into the function.

Something like so (untested):

```js
function readMagicBytes( buffer ) {

  const view = new DataView( buffer );
  if ( String.fromCharCode( view.getUint8( 0 ) ) === '{' ) {

    return null;

  }

  let magicBytes = '';
  for ( let i = 0; i < 4; i ++ ) {

    res += String.fromCharCode( view.getUint8( i ) );

  }

  return magicBytes;

}
```

If the file seems to have no magic bytes (JSON) then maybe passing `null` in is reasonable? Then the user can figure out what to do when overriding the function. By default we can assume glTF or fall back to the file extension.

src/utilities/tileContentType.js
```diff
@@ -0,0 +1,46 @@
...
+
+	const view = new DataView( buffer );
+
+	// the glTF json format is the only json-based format supported, so
+	// we'll just assume something that looks like json has to be glTF.
+	if ( String.fromCharCode( view.getUint8( 0 ) ) === '{' ) {
+
+		return TileContentType.GLTF;
+
+	}
```
gkjohnson, line 40: In your case is your server providing glTF files with no file extension _and_ no magic bytes in the file?

src/base/TilesRendererBase.js
```diff
@@ -3,6 +3,7 @@ import { LRUCache } from '../utilities/LRUCache.js';
 import { PriorityQueue } from '../utilities/PriorityQueue.js';
 import { determineFrustumSet, toggleTiles, skipTraversal, markUsedSetLeaves, traverseSet } from './traverseFunctions.js';
 import { UNLOADED, LOADING, PARSING, LOADED, FAILED } from './constants.js';
+import { getTileContentType } from "../utilities/tileContentType";
```
gkjohnson, line 6: nit: single quotes (just updated the lint rules to catch this)

gkjohnson (commented):
Thanks! This is looking good. I think the big remaining disconnect is that I think we can just pass the magic bytes ascii directly into the `parseTile` function - would that work for your use case?

src/base/TilesRendererBase.js
gkjohnson: I think using the result of `preprocessURL` is good - it's really just there to let people add url parameters, etc to the result.

src/utilities/tileContentType.js
gkjohnson: Alternatively we could just read the magic bytes in the `parseTile` function itself since the buffer is passed in which would be even simpler. That might have my vote actually.

src/base/TilesRendererBase.js
usefulthink: Ah yes, thanks! Since prettier was released I started not paying attention to these sort of things and now I need a linter to tell me when I'm doing it wrong :)

src/utilities/tileContentType.js
usefulthink: this is just for the json-format, so no magic-bytes. And yes, no file-extensions (for all I know, the urls are something akin to uuids and I believe are dynamically generated). But it's only using tileset-json and .glb tiles with embedded textures.

src/utilities/tileContentType.js
usefulthink: Yeah, I can totally see your point and it wouldn't make much of a difference for what I'm doing. I'll give the last option a go then.

src/base/TilesRendererBase.js
gkjohnson: Yeah I didn't realize the linter wasn't catching this - I updated it a few days ago to catch quotes.

PR #258, opened 2022-06-01
Add method for retrying all failed tiles

### LINKED ISSUE #253: Add API for TilesRenderer for retrying failed tiles
```js
if ( tiles.stats.failed > 0 ) {

  // set all FAILED tile states to UNLOADED
  // reset failed stats
  tiles.resetFailedTiles();

}
```

jeffpamer:
Resolves #253
Adds a public API to the base class to allow the consuming client to retry any failed tiles as needed.


gkjohnson:
Nooice. Can we add the function to the docs in the README, as well?

PR #259, opened 2022-06-22
fix:Failed to load b3dm model with error code 404

familyaa:
Provided ionAccessToken and ionAssetId as tests.


familyaa:
before:
image
now:
image

gkjohnson:
Hello @familyaa! Thanks for the suggested fix. Did the Cesium Ion API change? Do you have any referenced to documentation on how to consume Cesium Ion urls? This example used to work with a the originally provided asset id and public cesium ion access token.

familyaa:
hello @gkjohnson 
I didn't find any description about the api change in the cesium ion api documentation, but found through the cesium sandcastle example that assetId is required when accessing the tiled data. 
You can see
https://cesium.com/learn/ion/rest-api/#operation/getAssetEndpoint
and
https://sandcastle.cesium.com/?src=3D%20Tiles%20Photogrammetry.html

gkjohnson:
It looks like this has been caused by #256. I've proposed a fix in #262

gkjohnson:
@familyaa with #262 merged can you give this another try

familyaa:
@gkjohnson 
Great, the demo is working fine.

PR #262, opened 2022-06-24
Readd trailing slash

gkjohnson:
cc @usefulthink 

It looks like #256 actually broke all the demos. Is there a reason the trailing slash is being removed entirely from the tileset basepath string? Readding it and ensuring it's there fixes the issue. Here's the difference in behavior with `URL`:

```js
new URL( 'page.html', 'http://www.test.com/end/path' ).toString()
// http://www.test.com/end/page.html

new URL( 'page.html', 'http://www.test.com/end/path/' ).toString()
// http://www.test.com/end/path/page.html
```

### DIFF
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -191,7 +191,7 @@ export class TilesRendererBase {
 			if ( tile.content.uri ) {
 
 				// tile content uri has to be interpreted relative to the tileset.json
-				tile.content.uri = new URL( tile.content.uri, tileSetDir ).toString();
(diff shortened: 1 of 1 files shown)


usefulthink:
Oh my, sorry :/

This is a completely reasonable fix. The alternative would be to make sure the trailing slash is always present by adding it here: https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/34df85eb0cb8e766a09b3c12b5550348c62e7c9a/src/base/TilesRendererBase.js#L341-L342

gkjohnson:
It happens! I'll stick with this change for now

PR #264, opened 2022-06-27
support boxes as part of frustum check

Gmadges:


gkjohnson:
Thanks! If we're going to add this I think we can remove [these conditions](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/19896fcd48ec6dfe3e329ce279b6e320a118955b/src/three/TilesRenderer.js#L528-L536) in `TilesRenderer` that initialize a sphere from a box if it doesn't exist for the sake of frustum culling. It does look like it's required for the debug renderer [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/19896fcd48ec6dfe3e329ce279b6e320a118955b/src/three/DebugTilesRenderer.js#L193-L202), though, so why don't we just use `maxDistance = this.maxDebugDistance` there.

Gmadges:
Awesome thanks, I'll try and get an updated PR to you soon

gkjohnson:
Great, thank you!

gkjohnson:
@Gmadges Actually this change isn't correct -- the boxes provided in the tileset hierarchy are actually OBBs rather than AABBs so the `tile.cached.boxTransform` needs to be included in that frustum check. A custom function would need to be written to support OBB frustum intersections, I think.

I'll revert this for now but feel free to add in OBB support and we can get this merged again!

PR #267, opened 2022-06-28
Fix mising typescript types

robertlong:
This PR fixes a couple missing types that were preventing our project from compiling.

Also, I have some more I'd like to do in the future if you're up for it @gkjohnson. We'll probably be using the base classes for most of the work we're doing on https://github.com/matrix-org/thirdroom. We're using Three.js but the actually loading of glTF files happens on another thread in our engine so for my next step in integrating 3D tiles, I need to rewrite the `/three` folder to work in our engine. That means I really need all the other files to have accurate typescript definitions. I'm up for submitting PRs for those typescript updates if you are! That way we don't have to maintain our fork of this library.

### DIFF
--- a/src/base/LoaderBase.d.ts
+++ b/src/base/LoaderBase.d.ts
@@ -1,0 +1,10 @@
+export class LoaderBase {
+
+	fetchOptions: any;
+	workingPath: string;
+	load( url: string ): Promise< any >;
+	resolveExternalURL( url: string ): string;
+	workingPathForURL( url: string ): string
+	parse( buffer: ArrayBuffer ): Promise< any >;
+
+}
--- a/src/utilities/LRUCache.d.ts
+++ b/src/utilities/LRUCache.d.ts
@@ -13,6 +13,6 @@ export class LRUCache {
 	markAllUnused() : void;
 
 	unloadUnusedContent() : void;
-	scheduleUnload( markAllUnused : Boolean );
+	scheduleUnload( markAllUnused? : Boolean ): void;
 
 }


gkjohnson:
Hey @robertlong! Thanks for the adjustments.

> We're using Three.js but the actually loading of glTF files happens on another thread in our engine so for my next step in integrating 3D tiles, I need to rewrite the /three folder to work in our engine.

Can you explain what is limiting about the current three.js wrapper implementation? Is the loading or the parsing happening in a worker? If possible it would be nice to include some adjustments to the three.js TilesRenderer class to support your use case since it sounds like it might benefit others.

> I'm up for submitting PRs for those typescript updates if you are! That way we don't have to maintain our fork of this library.

I'd like to avoid forks as much as possible! Definitely open to PRs for the classes or types files!

gkjohnson:
> The loading and parsing are happening in a worker and then the contents are pushed over to another render worker using OffscreenCanvas or to the main thread where the renderer is using a regular canvas.

Sounds good - just to be clear the actual tile downloading happens in the base TilesRenderer [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/base/TilesRendererBase.js#L394) so there may need to be some changes that occur in the base classes, as well, to fully support your use case. I'm happy to discuss those changes and what's needed to get them working, though. Maybe it's worth discussing in another issue?

> Also, I just noticed some things seem to break between what's on main and v0.3.9. I'm installing this branch directly from github which is based off main. It seems to be less reliable about loading the higher LOD tiles and the skybox in the Mars example seems to have some frustum culling issues.

Yes this is related to #264 that I just merged. I forgot that the bounding box is an OBB rather than axis aligned and more needs to happen to support frustum culling by box bounds. I'll revert that now.

PR #271, opened 2022-07-02
Use magic bytes to determine file type when parsing

### LINKED ISSUE #254: content loading relies on filename-extensions

gkjohnson:
Fix #254 

cc @usefulthink this is what I was thinking in place of #257. Let me know what you think! Maybe you can give it a try to make sure it works for your usecase.

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -15,6 +15,7 @@ import {
 	LoadingManager
 } from 'three';
 import { raycastTraverse, raycastTraverseFirstHit } from './raycastTraverse.js';
+import { readMagicBytes } from '../utilities/readMagicBytes.js';
 
 const INITIAL_FRUSTUM_CULLED = Symbol( 'INITIAL_FRUSTUM_CULLED' );
 const tempMat = new Matrix4();
@@ -600,7 +601,8 @@ export class TilesRenderer extends TilesRendererBase {
 
 		}
 
-		switch ( extension ) {
+		const fileType = readMagicBytes( buffer ) || extension;
+		switch ( fileType ) {
 
 			case 'b3dm': {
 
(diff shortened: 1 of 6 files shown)


src/three/TilesRenderer.js
```diff
@@ -600,7 +601,8 @@ export class TilesRenderer extends TilesRendererBase {
 
 		}
 
-		switch ( extension ) {
+		const fileType = readMagicBytes( buffer ) || extension;
```
gkjohnson, line 604: In the case of your API what is passed in as the "extension" here? If you're loading glTF files then there won't be a magic bytes token and we'll fall back to `extension` which likely won't work for you. Should we set it to this?

```js
const fileType = readMagicBytes( buffer ) || extension || 'gltf';
```

usefulthink: hmm, yes. I see the problem here - assuming that everything that doesn't have magic bytes is glTF-json doesn't feel right (and in fact it could actually be a external-tileset json as well).

Maybe we can add an instance-method (something like `TilesRenderer.getContentType(extension: string, buffer: ArrayBuffer): string`),  that can be overridden for a scenario where neither the content-type nor the extension gets a good result.

gkjohnson: I'm okay assuming that a model is a fallback to gltf for now - I just want to make sure it works for your use case. Is `extension` null or undefined when your extension-less URLs are processed?

usefulthink: It's not too relevant for my use case. As far as I know, all of the meshes will be in glTF-binary with embedded textures, so that should be fine either way.

PR #273, opened 2022-07-06
Fix RTC_CENTER being overwritten by the adjustmentTransform

nickguletskii:
The changes introduced in  d2de4c5bc3ccd3273a3c9f1b53bdf3f8b27a7e93 overwrite the offset introduced by `RTC_CENTER`. This PR fixes this by updating the scene's model matrix before applying the adjustment transform.


gkjohnson:
Great! Thanks for the find and the fix.

PR #282, opened 2022-08-25
add gltf extension cesium_rtc

peiyu7921:


src/three/GLTFCesiumRTCExtension.d.ts
```diff
@@ -0,0 +1,8 @@
+import {GLTF} from "three/examples/jsm/loaders/GLTFLoader";
```
gkjohnson, line 1: nit: Can we add spaces inside the braces and use single quotes

gkjohnson (commented):
This looks great! one comment on code style and we can merge this.

gkjohnson:
Great, thanks! I assume you've tested this with your use case. I'll publish a new version soon!

PR #286, opened 2022-09-27
Add support for tilesets with region bounds

### LINKED ISSUE #66: Add support for Region Bounding Volume
Regions should be specified in the local frame of the group treating the group as the [WGS84](https://en.wikipedia.org/wiki/World_Geodetic_System#/media/File:WGS_84_reference_frame_(vector_graphic).svg) coordinate frame.

Example data [here](https://github.com/CesiumGS/3d-tiles-samples/blob/main/1.0/TilesetWithRequestVolume/tileset.json).

Spec [here](https://github.com/CesiumGS/3d-tiles/tree/master/specification#region).

More information in https://github.com/CesiumGS/3d-tiles/issues/445

### LINKED ISSUE #283: ThreeTilesRenderer : Region bounds not supported.
**Why does loading the simplest model cause such errors? How can I solve them?**
[1.zip](https://github.com/NASA-AMMOS/3DTilesRendererJS/files/9580007/1.zip)
------------------------------------------------------------------------------------------

gkjohnson:
Fix #66 
Fix #283

image


gkjohnson:
cc @kingwangwang @waterseeding this PR should add support for the region bounding volume types if you'd like to test it.

PR #289, opened 2022-09-28
add dispose for TileRenderer

yangyanggu:
add dispose function for TileRenderer, clear all cache data.


gkjohnson:
Can you explain the value of this disposal method? Are you noticing that objects are not being disposed of?

The purpose of "dispose" functions in three.js is to release WebGL and GPU memory that otherwise isn't just naturally released with garbage collection. There is already a dispose function that releases all tiles from the cache which implicitly disposes of the tile geometry and materials so on the surface this change seems unnecessary.

yangyanggu:
> Can you explain the value of this disposal method? Are you noticing that objects are not being disposed of?
> 
> The purpose of "dispose" functions in three.js is to release WebGL and GPU memory that otherwise isn't just naturally released with garbage collection. There is already a dispose function that releases all tiles from the cache which implicitly disposes of the tile geometry and materials so on the surface this change seems unnecessary.

I has test the "dispose" funtion in TilesRendererBase.js, but the GPU memory can not release, I find that when i execute the dispose function, the Gltf cache is still in lruCache and TilesRenderer, it cause GPU memory can not release. After I add this dispose in TilesRenderer, the memory is released.

gkjohnson:
From my testing in the "mars.html" demo and looking at `renderer.info.memory` the geometry and texture memory is correctly released. If you're seeing different please provide some repro code and a jsfiddle if possible.

yangyanggu:
> From my testing in the "mars.html" demo and looking at `renderer.info.memory` the geometry and texture memory is correctly released. If you're seeing different please provide some repro code and a jsfiddle if possible.

https://github.com/yangyanggu/3d-renderer-gpu-release-demo   this is my demo. it has a 460M 3dtiles files

yangyanggu:
> Thanks - it looks like this is related to the number of external tilesets used in the model and the way they are disposed. Specifically the external tilesets are disposed of first which removes the children and before the rest of the tileset is traversed to dispose of each child. See [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/8a8d246d26e4089c79942ceec4c519c4d981068b/src/base/TilesRendererBase.js#L626-L631) and [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/8a8d246d26e4089c79942ceec4c519c4d981068b/src/base/TilesRendererBase.js#L418-L422).
> 
> The better fix would be to address that issue. Changing these lines:
> 
> https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/8a8d246d26e4089c79942ceec4c519c4d981068b/src/base/TilesRendererBase.js#L626-L631
> 
> to something like this:
> 
> ```js
> const toRemove = [];
> this.traverse( t => toRemove.push( t ) );
> for ( let i = 0, l = toRemove.length; i < l; i ++ ) {
> 
>   lruCache.remove( toRemove[ i ] );
> 
> }
> ```
> 
> should address the issue.

thanks, This method can fix this issue, I find the GPU memory can not be released is caused by the browser's devtools, when i open the edge'devtools, the memory can not be released. if I close it ,the memory is released immediately.
Finally, please modify the code in the next version， Thanks

gkjohnson:
> Finally, please modify the code in the next version， Thanks

Unfortunately I do not have the bandwidth to work on and fully test the issue. If you would like this addressed soon I would appreciate it if you updated the dispose function in this PR with the above code and tested it more thoroughly with your use case to make sure it works. Then we can get this merged.

PR #290, opened 2022-09-28
Add support for position quantized [PNTS]

### LINKED ISSUE #280: PNTSLoader Compatibility issues
I have a set of PNTS format data， When I loaded it, I reported an error，cesium load it is normal，the tile metadata has no POINTS_LENGTH attribute is under the header，This set of data is converted through cesium lab
image
image

pizza3:
Fix #280


gkjohnson:
Thanks! Have you been able to test this in a dataset? Are you able to provide a PNTS model to test this functionality with?

src/three/PNTSLoader.js
```diff
@@ -43,7 +43,26 @@ export class PNTSLoader extends PNTSLoaderBase {
 				} );
 
 				const geometry = new BufferGeometry();
-				geometry.setAttribute( 'position', new BufferAttribute( POSITION, 3, false ) );
+				if ( POSITION ) {
+
+  				geometry.setAttribute( 'position', new BufferAttribute( POSITION, 3, false ) );
```
gkjohnson, line 48: nit: this indentation looks incorrect

src/three/PNTSLoader.js
```diff
@@ -43,7 +43,26 @@ export class PNTSLoader extends PNTSLoaderBase {
...
+
+				} else if ( POSITION_QUANTIZED ) {
+
+					const decodedPositions = new Float32Array( POSITION_QUANTIZED.length );
+					for ( let i = 0; i < POSITION_QUANTIZED.length / 3; i ++ ) {
```
gkjohnson, line 53: For the sake of performance can we premptively declare the length to iterate over:

```js
for ( let i = 0, l = POSITION_QUANTIZED.length / 3; i < l; i ++ ) ...
```

pizza3:
Hey @gkjohnson, so I tested it using [this](https://github.com/CesiumGS/cesium/blob/main/Specs/Data/Cesium3DTiles/PointCloud/PointCloudQuantized/pointCloudQuantized.pnts) pnts file first on the cesium ion just to get the general idea of the result, then used it for the testing over here with the PNTSLoader, I was able to get the features following the spec document mentioned on the issue and produced the result.

src/three/PNTSLoader.js
```diff
@@ -43,7 +43,26 @@ export class PNTSLoader extends PNTSLoaderBase {
...
 				const geometry = new BufferGeometry();
-				geometry.setAttribute( 'position', new BufferAttribute( POSITION, 3, false ) );
+				if ( POSITION ) {
+
+					geometry.setAttribute( 'position', new BufferAttribute( POSITION, 3, false ) );
+
+				} else if ( POSITION_QUANTIZED ) {
+
+					const decodedPositions = new Float32Array( POINTS_LENGTH );
+					for ( let i = 0; i < POINTS_LENGTH / 3; i ++ ) {
```
gkjohnson, line 53: Is this correct? The `POSITION_QUANTIZED` field should be 3x the length of `POINTS_LENGTH` since it holds vector3s. It seems when using `POINTS_LENGTH` we shouldn't be dividing by 3. I'm looking at the spec here:

https://github.com/CesiumGS/3d-tiles/tree/main/specification/TileFormats/PointCloud#global-semantics

gkjohnson:
> Hey @gkjohnson, so I tested it using [this](https://github.com/CesiumGS/cesium/blob/main/Specs/Data/Cesium3DTiles/PointCloud/PointCloudQuantized/pointCloudQuantized.pnts) pnts file first on the cesium ion just to get the general idea of the result, then used it for the testing over here with the PNTSLoader, I was able to get the features following the spec document mentioned on the issue and produced the result.

Thanks! I'll give that a test when I get a chance.

src/three/PNTSLoader.js
pizza3: Oh my god, so sorry for this, I should've double checked it before making the commit, I've fixed it now

gkjohnson: Heh - that's what code reviews are for 😁

gkjohnson:
Just gave it a test - the test model you provided above as some very large offset values (`[ 1215007.8828876738, - 4736318.051199594, 4081600.22126042 ]`) which is causing the points to jitter when the camera is looking at them. This is [handled in the I3DMLoader](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/6664b51c4027c621372f1dc99487cc421a785583/src/three/I3DMLoader.js#L134-L141) by applying the offset as a transform rather than to the float 32 buffer values themselves. So I think we should instead apply this scale and position offset to the mesh transform rather than the points themselves since they'll be multiplied with the camera transform on the CPU which will alleviate floating point issues on the gpu.

src/three/PNTSLoader.js
```diff
@@ -43,7 +43,31 @@ export class PNTSLoader extends PNTSLoaderBase {
...
+
+							const index = 3 * i + j;
+							decodedPositions[ index ] = ( POSITION_QUANTIZED[ index ] / 65535.0 ) * QUANTIZED_VOLUME_SCALE[ j ];
+
+						}
+
+					}
+					averageVector.x += QUANTIZED_VOLUME_OFFSET[ 0 ];
+					averageVector.y += QUANTIZED_VOLUME_OFFSET[ 1 ];
+					averageVector.z += QUANTIZED_VOLUME_OFFSET[ 2 ];
```
gkjohnson, line 66: nit: not a huge deal but since these are just set once we can just assign the values to the x, y, z fields.

gkjohnson (commented):
@pizza3 Thanks for the update! Looks like there are some merge conflicts now since some work has happened since this was first opened. Can you fix those? And then I think this is ready for merging once I've tested it.

PR #299, opened 2022-10-31
Update type definitions to correctly type properties and expose available APIs

jesse-small:
TilesRendererBase.d.ts
- Added the _RequestInit_ type instead of _Object_ so that the expected type is used and typescript will not complain about accessing non-existing properties

TilesRenderer.d.ts
- Added getOrientedBounds
- Added getBoundingSphere
- Added raycast

### DIFF
--- a/src/three/TilesRenderer.d.ts
+++ b/src/three/TilesRenderer.d.ts
@@ -13,9 +13,11 @@ export class TilesRenderer extends TilesRendererBase {
 
 	group : TilesGroup;
 
-	getBoundsTransform( target: Matrix4 ) : Boolean;
-
 	getBounds( box : Box3 ) : Boolean;
+	getOrientedBounds( box, matrix ) : Boolean;
+	getBoundingSphere( sphere ) : Boolean;
+
+	raycast( raycaster, intersects ) : Boolean;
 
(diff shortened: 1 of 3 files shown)


gkjohnson:
Thanks!

src/three/TilesRenderer.d.ts
```diff
@@ -13,9 +13,11 @@ export class TilesRenderer extends TilesRendererBase {
 
 	group : TilesGroup;
 
-	getBoundsTransform( target: Matrix4 ) : Boolean;
-
 	getBounds( box : Box3 ) : Boolean;
+	getOrientedBounds( box, matrix ) : Boolean;
+	getBoundingSphere( sphere ) : Boolean;
+
+	raycast( raycaster, intersects ) : Boolean;
```
gkjohnson, line 20: This function is not documented and I didn't intend of it to be public. You can use the raycaster on the group if you need a raycasting API - so why don't we remove this function from the definition.

I also notice that `getBoundingSphere` isn't documented but it's probably worth adding it to the README.

jesse-small:
@gkjohnson I have pushed up the requested changes, let me know if there are any issues.
Thanks.

gkjohnson:
Great, thank you!

PR #301, opened 2022-11-02
Exporting GLTFCesiumRTCExtension

jesse-small:
Exporting GLTFCesiumRTCExtension so it can be used and registered with the GLTFLoader to support use with B3DMLoaders. 
Also added a check for the extensions property before running the extension to ensure no errors are thrown when non-CesiumRTC tiles are run against it.

### DIFF
--- a/src/three/GLTFCesiumRTCExtension.js
+++ b/src/three/GLTFCesiumRTCExtension.js
@@ -1,5 +1,4 @@
-
-export default class GLTFCesiumRTCExtension {
+export class GLTFCesiumRTCExtension {
 
 	constructor() {
 
@@ -9,10 +8,14 @@ export default class GLTFCesiumRTCExtension {
 
 	afterRoot( res ) {
 
-		const { center } = res.parser.json.extensions.CESIUM_RTC;
-		res.scene.position.x += center[ 0 ];
-		res.scene.position.y += center[ 1 ];
-		res.scene.position.z += center[ 2 ];
+		if ( res.parser.json.extensions ) {
+
+			const { center } = res.parser.json.extensions.CESIUM_RTC;
+			res.scene.position.x += center[ 0 ];
+			res.scene.position.y += center[ 1 ];
+			res.scene.position.z += center[ 2 ];
+
+		}
 
 	}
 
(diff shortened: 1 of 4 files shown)


src/three/GLTFCesiumRTCExtension.js
```diff
@@ -9,10 +8,14 @@ export default class GLTFCesiumRTCExtension {
...
-		res.scene.position.y += center[ 1 ];
-		res.scene.position.z += center[ 2 ];
+		if ( res.parser.json.extensions ) {
+
+			const { center } = res.parser.json.extensions.CESIUM_RTC;
+			res.scene.position.x += center[ 0 ];
+			res.scene.position.y += center[ 1 ];
+			res.scene.position.z += center[ 2 ];
+
+		}
```
gkjohnson, line 18: Can you explain the need for wrapping this in a condition? I'm less familiar with building glTF extensions so I'm just curious.

gkjohnson:
This solution looks good - thanks! Just one question above.

src/three/GLTFCesiumRTCExtension.js
jesse-small: Sure, when we register the CESIUM_RTC extension, GLTFLoader will attempt to run it against all GLTF even if they do not support CESIUM_RTC. 
So the above check ensures the properties required to run the extension are available, otherwise, we simply return as nothing needs to be done. Upon a second look, I should probably add a null check for the CESIUM_RTC property as well.

gkjohnson: Thanks for the explanation!

> I should probably add a null check for the CESIUM_RTC property as well.

This sounds good!

jesse-small: @gkjohnson I have pushed the extra changes just to ensure the center property is non-null before accessing it this should leave us in a pretty safe position moving forward.

jesse-small: Also, as I would love to pull and use the changes I've made, any ideas on when the next release will be? (0.3.15) Thanks!

gkjohnson: Thanks for the changes! I will merge and publish over the next couple days. I am in the middle of an international move so I'm moving a little slow online at the moment. Thanks for the patience!

gkjohnson:
Thanks! Just published v0.3.15:

https://github.com/NASA-AMMOS/3DTilesRendererJS/releases/tag/v0.3.15

PR #304, opened 2022-11-09
update GLTFExtensionLoader to import cesium extension correctly

jesse-small:
Just noticed that one of the imports needed to be updated. I have included this change, sorry for missing this. A new version should be released to fix this as it can cause issues otherwise.

### DIFF
--- a/src/three/GLTFExtensionLoader.js
+++ b/src/three/GLTFExtensionLoader.js
@@ -1,7 +1,7 @@
 import { DefaultLoadingManager } from 'three';
 import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader.js';
 import { LoaderBase } from '../base/LoaderBase.js';
-import GLTFCesiumRTCExtension from './GLTFCesiumRTCExtension.js';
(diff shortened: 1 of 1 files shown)


jesse-small:
@gkjohnson I found an import that I missed updating in my last PR, this resolves the issue. The update should be pushed in and released as soon as possible as it can cause issues otherwise when using GLTFLoader.

gkjohnson:
Great, thanks! I'll merge and publish later today.

I wonder why this wasn't caught with CI / linter?

jesse-small:
> Great, thanks! I'll merge and publish later today.
> 
> I wonder why this wasn't caught with CI / linter?

Sounds good. Yeah, it definitely seems like something that should've been caught by linter. I believe we use eslint-plugin-import and that checks that the imports are valid and not just formatted correctly.

gkjohnson:
> I believe we use eslint-plugin-import and that checks that the imports are valid and not just formatted correctly.

Oh I didn't know about that plugin - are you interested in making a PR to add that to the linting process? I'll get this published.

Issue #323, opened 2023-04-13
Is there a plan to support 3DTiles 1.1


ligaofeng0901:
3DTiles 1.1 includes several new features and improvements
https://github.com/CesiumGS/3d-tiles/tree/draft-1.1-release/specification
Can we support the new version?

dbuck:
I went looking for a summary of changes from 1.0 -> 1.1:

from 1.1 draft: [README.md#version-history](https://github.com/CesiumGS/3d-tiles/blob/draft-1.1-release/README.md#version-history)

- [3D Tiles 1.1](https://github.com/CesiumGS/3d-tiles)
  - Additions:
    - Support for structured metadata that can be associated with tilesets, tiles, tile content, and tile content groups
    - Directly support glTF assets as tile contents
    - Support for multiple tile contents
    - Support for implicit tiling schemes
  - Deprecations:
    - The original tile formats (b3dm, i3dm, pnts, and cmpt) are deprecated in favor of glTF content
    - The `tileset.properties` are deprecated, in favor of the more versatile metadata support

gltf extension works now, I'm using that in matterport tilesets
metadata would need additional work, there was some disconnect between getting the top level tileset blobs at tile parse iirc when I looked at whether that was one that could be added via 1.0 extensions, we opted to include our data in extras at the time.

implicit tiling and multiple tile contents would both need some substantial effort to implement I believe?

gkjohnson:
> implicit tiling and multiple tile contents would both need some substantial effort to implement I believe?

I'd have to do a deeper dive but for multiple tile contents it may be the case that the bulk of the changes should be able to be consolidated in the [requestTileContents](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/base/TilesRendererBase.js#L395) function. As far as I can tell it looks like it just means that multiple fetch and parse promises have to be managed per tile and evicted from the caches together.

For implicit tiling the easy thing to do would probably be to preprocess any loaded tilesets (including externally referenced tilesets) and assign the implicit bounding boxes to the children ahead of time - though I haven't looked deeply at how complex the format is here.

gkjohnson:
Closing in favor of #235

PR #326, opened 2023-04-27
Feature: PNTs Draco support

Gmadges:
This PR provides support for PNTs files that are using the `3DTILES_draco_point_compression` extension.

It only supports Positions & Colors currently.
This seemed fine because 3DTilesRendererJS only supports those attributes currently. 
It also allowed me to use the three.js DRACOLoader class which I don't think supports arbitrary Draco attributes either (I could be wrong there).

I'm not 100% happy with the way in which I inject the Draco loader (via the the .drc handler) but it seemed like the lightest touch method that vaguely makes sense.

### DIFF
--- a/src/three/PNTSLoader.js
+++ b/src/three/PNTSLoader.js
@@ -1,122 +1,134 @@
-import { PNTSLoaderBase } from "../base/PNTSLoaderBase.js";
+import { PNTSLoaderBase } from '../base/PNTSLoaderBase.js';
 import {
 	Points,
 	PointsMaterial,
 	BufferGeometry,
 	BufferAttribute,
 	DefaultLoadingManager,
-} from "three";
+} from 'three';
 
 const DRACO_ATTRIBUTE_MAP = {
-	RGB: "color",
-	POSITION: "position",
+	RGB: 'color',
+	POSITION: 'position',
 };
 export class PNTSLoader extends PNTSLoaderBase {
-	constructor(manager = DefaultLoadingManager) {
+
+	constructor( manager = DefaultLoadingManager ) {
+
 		super();
 		this.manager = manager;
 
 		// very hacky way of getting the draco loader
-		this.dracoLoader = this.manager.getHandler("test.drc");
+		this.dracoLoader = this.manager.getHandler( 'test.drc' );
+
 	}
 
-	parse(buffer) {
-		return super.parse(buffer).then(async (result) => {
+	parse( buffer ) {
+
+		return super.parse( buffer ).then( async ( result ) => {
+
 			const { featureTable } = result;
 
 			const geometry = new BufferGeometry();
 			const material = new PointsMaterial();
 
-			if (featureTable.isDraco()) {
+			if ( featureTable.isDraco() ) {
+
 				const dracoIDs = featureTable.getDracoProperties();
 				const attributeIDs = {};
 
-				for (let [key, value] of Object.entries(dracoIDs)) {
-					attributeIDs[DRACO_ATTRIBUTE_MAP[key]] = value;
+				for ( const [ key, value ] of Object.entries( dracoIDs ) ) {
+
+					attributeIDs[ DRACO_ATTRIBUTE_MAP[ key ] ] = value;
+
 				}
 
 				const taskConfig = {
 					attributeIDs,
 					attributeTypes: {
-						position: "Float32Array",
-						color: "Uint8Array",
+						position: 'Float32Array',
+						color: 'Uint8Array',
 					},
 					useUniqueIDs: true,
 				};
 
 				const buffer = featureTable.getDracoBuffer();
 
-				if (this.dracoLoader == null) {
-					throw new Error("PNTSLoader: dracoLoader not available.");
+				if ( this.dracoLoader == null ) {
+
+					throw new Error( 'PNTSLoader: dracoLoader not available.' );
+
 				}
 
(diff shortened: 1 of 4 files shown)


src/three/PNTSLoader.js
```diff
@@ -1,53 +1,84 @@
...
+				const attributeIDs = {};
 
-						console.warn( `PNTSLoader: Unsupported FeatureTable feature "${ feature }" detected.` );
+				for ( const [ key, value ] of Object.entries( dracoIDs ) ) {
 
-					}
+					attributeIDs[ DRACO_ATTRIBUTE_MAP[ key ] ] = value;
 
-				} );
+				}
```
gkjohnson, line 44: This will cause an undefined key if a DRACO property exists that is not "RGB" or "POSITION", right?

src/three/PNTSLoader.js
```diff
@@ -1,53 +1,84 @@
...
-				material.size = 2;
-				material.sizeAttenuation = false;
+					throw new Error( 'PNTSLoader: dracoLoader not available.' );
+
+				}
+
+				const dracoGeometry = await this.dracoLoader
+					.decodeGeometry( buffer, taskConfig );
+
+				geometry.copy( dracoGeometry );
```
gkjohnson, line 66: Why copy instead of just using the DRACO geometry directly?

src/three/PNTSLoader.js
```diff
@@ -1,53 +1,84 @@
...
+
+				if ( geometry.attributes.color ) {
+
+					geometry.attributes.color.normalized = true;
```
gkjohnson, line 70: Does the DRACO loader not set the color field to normalized correctly automatically? Is this something that should be fixed in three.js?

src/three/PNTSLoader.js
```diff
@@ -1,53 +1,84 @@
...
+};
 export class PNTSLoader extends PNTSLoaderBase {
 
 	constructor( manager = DefaultLoadingManager ) {
 
 		super();
 		this.manager = manager;
 
+		// hacky way of getting the draco loader from the manager
+		this.dracoLoader = this.manager.getHandler( 'draco.drc' );
```
gkjohnson, line 22: I think this should be done in the "parse" function rather than in the constructor.

src/three/PNTSLoader.js
```diff
@@ -1,53 +1,84 @@
...
+					useUniqueIDs: true,
+				};
+
+				const buffer = featureTable.getDracoBuffer();
+
+				if ( this.dracoLoader == null ) {
 
-				const material = new PointsMaterial();
-				material.size = 2;
-				material.sizeAttenuation = false;
```
gkjohnson, line 50: It looks like we lost these to initialized values for the points material.

src/utilities/FeatureTable.js
```diff
@@ -141,6 +141,41 @@ export class FeatureTable {
...
+		if ( ! this.isDraco() ) {
+
+			throw new Error( 'FeatureTable: Feature data is not draco encoded' );
+
+		}
+
+		return this.header.extensions[ '3DTILES_draco_point_compression' ].properties;
+
+	}
+
```
gkjohnson, line 178: I think ideally the `FeatureTable` class wouldn't have to change since this is all really just relevant to the points properties, right? Otherwise if there are other relevant compressed buffers in the FeatureTable then it might make most sense to make a derivative class that abstracts away the draco decompression and automatically decompresses the data when a user calls "getData", for example.

gkjohnson (commented):
Thanks! This looks like a good thing to add.

> It also allowed me to use the three.js DRACOLoader class which I don't think supports arbitrary Draco attributes either (I could be wrong there).

This is something we can clarify if we need to broaden support to other fields like NORMAL or BATCH_ID.

> It only supports Positions & Colors currently.
> This seemed fine because 3DTilesRendererJS only supports those attributes currently.

For now, yes - it looks like any non-glTF file formats are deprecated in the latest 1.1 spec (see https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/323#issuecomment-1529822491) so I'm no less worried about supporting these formats in a completely robust way at the moment.

> I'm not 100% happy with the way in which I inject the Draco loader (via the the .drc handler) but it seemed like the lightest touch method that vaguely makes sense.

Yeah but I don't know if I like the alternatives, either, especially given the above point:

- Specifically we could have a `setDRACOLoader` function on the tiles class.
- Ask the user to provide a custom handler for ".pnts" files that already has the draco loader set the same way we do for GLTF.

If anything maybe the second option makes the most sense? Though I'm not picky here.

src/three/PNTSLoader.js
Gmadges: Good idea, I'll have a look

src/utilities/FeatureTable.js
Gmadges: I've now moved more of the logic to a `PNTSFeatureTable` to keep the draco related logic away from other data types.

I wasn't able to abstract it fully so that you could just use `getData()` as normal. This was due to the `dracoLoader` returning geometry not just a typed array.

src/three/PNTSLoader.js
Gmadges: made an issue and PR [here](https://github.com/mrdoob/three.js/pull/26036)
unsure if this desired behaviour for everyone, but feels like it

gkjohnson:
Just published! Thanks again

PR #329, opened 2023-05-11
Feature: Draco PNTS proposal alternative

gkjohnson:
Related to #326 

@Gmadges What do you think of this? I've based this off your previous commit and changed it a bit to consolidate all the decoding logic in the PNTS class.

I haven't tested it, though, so I'll need your help with that.

### DIFF
--- a/src/three/PNTSLoader.js
+++ b/src/three/PNTSLoader.js
@@ -11,15 +11,13 @@ const DRACO_ATTRIBUTE_MAP = {
 	RGB: 'color',
 	POSITION: 'position',
 };
+
 export class PNTSLoader extends PNTSLoaderBase {
 
 	constructor( manager = DefaultLoadingManager ) {
 
 		super();
 		this.manager = manager;
-
-		// hacky way of getting the draco loader from the manager
-		this.dracoLoader = this.manager.getHandler( 'draco.drc' );
 
 	}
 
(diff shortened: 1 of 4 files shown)


Gmadges:
@gkjohnson Looks good to me! A bit cleaner than my PNTSFeatureTable refactor.

I'll give it a test, and let you know how it goes

Gmadges:
Tested, Requires a few tweaks and then should work 👍

src/three/PNTSLoader.js
```diff
@@ -12,68 +23,115 @@ export class PNTSLoader extends PNTSLoaderBase {
...
-					'NORMAL',
-					'NORMAL_OCT16P',
-				].forEach( feature => {
+				// map PNTS keys to draco types
+				const attributeIDs = {};
+				for ( const key in properties ) {
 
-					if ( feature in featureTable.header ) {
+					const mappedKey = DRACO_ATTRIBUTE_MAP[ key ];
+					if ( mappedKey in attributeIds ) {
```
Gmadges, line 50: I think this should be 
```
if ( ! ( mappedKey in attributeIDs ) )
```

src/three/PNTSLoader.js
```diff
@@ -12,68 +23,115 @@ export class PNTSLoader extends PNTSLoaderBase {
...
+					attributeTypes: {
+						position: 'Float32Array',
+						color: 'Uint8Array',
+					},
+					useUniqueIDs: true,
+				};
+
+				const buffer = featureTable.getDracoBuffer( byteOffset, byteLength );
+				const geometry = await dracoLoader.decodeGeometry( buffer, taskConfig );
+				geometry.copy( dracoGeometry );
```
Gmadges, line 70:
```
const buffer = featureTable.getBuffer( byteOffset, byteLength );
geometry = await dracoLoader.decodeGeometry( buffer, taskConfig );
```

Gmadges (changes requested):
Tested, Requires a few tweaks and then should work 👍

src/three/PNTSLoader.js
gkjohnson: Thanks!

src/three/PNTSLoader.js
gkjohnson: Oh oops - you're right this isn't correct. But I think it should be the following to make sure both fields are defined:

```js
if ( key in DRACO_ATTRIBUTE_MAP && key in properties ) {

  // ...

} 
```

gkjohnson:
> Tested, Requires a few tweaks and then should work 👍

Thanks! Just made the changes. Can you give it another test to make sure it looks right and then we can get this merged!

src/three/PNTSLoader.js
```diff
@@ -12,68 +23,115 @@ export class PNTSLoader extends PNTSLoaderBase {
...
+				const taskConfig = {
+					attributeIDs,
+					attributeTypes: {
+						position: 'Float32Array',
+						color: 'Uint8Array',
+					},
+					useUniqueIDs: true,
+				};
+
+				const buffer = featureTable.getDracoBuffer( byteOffset, byteLength );
```
Gmadges, line 68: `getDracoBuffer` -> `getBuffer`

Gmadges (changes requested):
Sorry, some minor items, and it works 👍

gkjohnson:
> Sorry, some minor items, and it works 👍

Just made another update. If there are more issues do you mind making a PR into this "draco-pnts-proposal" branch?

Gmadges (approved):
Tested, and all works 👍

PR #330, opened 2023-05-17
Add initial google maps 3D tiles example implementation

aiden-jeffrey:
Example added (googleMapsExample.html) rendering the new [google maps tileset](https://cloud.google.com/blog/products/maps-platform/create-immersive-3d-map-experiences-photorealistic-3d-tiles).

To use this, you need to get a google api key and add it to the slot in the UI panel (I worked out how to make a free key from the links in the tiles landing page [here](https://cloud.google.com/blog/products/maps-platform/create-immersive-3d-map-experiences-photorealistic-3d-tiles)).

~This appears to successfully download a limited (which is expected) number of gltf files from the [google maps tile api](https://developers.google.com/maps/documentation/tile/create-renderer), however none of them render and there are no error messages for me to know where to start looking. I added a test sphere mesh to the scene just to make sure the camera etc were working correctly.~

### DIFF
--- a/example/googleMapsExample.html
+++ b/example/googleMapsExample.html
@@ -1,0 +1,53 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
+		<meta charset="utf-8"/>
+
+		<title>3D Tiles Renderer Options Example</title>
+
+		<style>
+			* {
+				margin: 0;
+				padding: 0;
+			}
+
+			html {
+				overflow: hidden;
+				font-family: Arial, Helvetica, sans-serif;
+				user-select: none;
+			}
+
+			canvas {
+				image-rendering: pixelated;
+				outline: none;
+			}
+
+			#info {
+				position: absolute;
+				top: 0;
+				left: 0;
+				color: white;
+				width: 100%;
+				text-align: center;
+				padding: 5px;
+				pointer-events: none;
+				line-height: 1.5em;
+			}
+
+			#info a {
+				color: white;
+				pointer-events: all;
+			}
+        </style>
+    </head>
+    <body>
+		<div id="info">
+			<div>
+				Google Maps 3D tiles example.
+				<br/>
+			</div>
+		</div>
+        <script src="./googleMapsExample.js"></script>
+    </body>
+</html>
--- a/example/googleMapsExample.html
+++ b/example/googleMapsExample.html
@@ -4,7 +4,7 @@
         <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
 		<meta charset="utf-8"/>
 
-		<title>3D Tiles Renderer Options Example</title>
+		<title>Google 3D Tiles Example</title>
 
 		<style>
 			* {
@@ -44,7 +44,7 @@
     <body>
 		<div id="info">
 			<div>
-				Google Maps 3D tiles example.
+				Google Maps 3D tiles example. Double click to refocus the camera.
 				<br/>
 			</div>
 		</div>
--- a/example/googleMapsExample.js
+++ b/example/googleMapsExample.js
@@ -1,14 +1,4 @@
-import {
-	DebugTilesRenderer as TilesRenderer,
-	NONE,
-	SCREEN_ERROR,
-	GEOMETRIC_ERROR,
-	DISTANCE,
-	DEPTH,
-	RELATIVE_DEPTH,
-	IS_LEAF,
-	RANDOM_COLOR,
-} from '../src/index.js';
+import { DebugTilesRenderer as TilesRenderer } from '../src/index.js';
 import {
 	Scene,
 	DirectionalLight,
@@ -16,20 +6,13 @@ import {
 	WebGLRenderer,
 	PerspectiveCamera,
 	CameraHelper,
-	Raycaster,
-	Vector2,
 	Vector3,
 	Quaternion,
-	Mesh,
-	CylinderBufferGeometry,
-	MeshBasicMaterial,
 	Group,
-	TorusBufferGeometry,
 	sRGBEncoding,
 	Matrix4,
 	Box3,
 	Sphere,
-	SphereGeometry,
 } from 'three';
 import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls.js';
 import { BufferGeometryUtils } from 'three/examples/jsm/utils/BufferGeometryUtils.js';
@@ -38,36 +21,27 @@ import { DRACOLoader } from 'three/examples/jsm/loaders/DRACOLoader.js';
 import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
 import Stats from 'three/examples/jsm/libs/stats.module.js';
 
-const ALL_HITS = 1;
-const FIRST_HIT_ONLY = 2;
-
 const apiOrigin = 'https://tile.googleapis.com';
 
 const hashUrl = window.location.hash.replace( /^#/, '' );
 let camera, controls, scene, renderer, tiles, cameraHelper;
-let raycaster, mouse, rayIntersect, lastHoveredElement;
 let offsetParent;
 let statsContainer, stats;
 
 const params = {
 
 	'enableUpdate': true,
-	'raycast': NONE,
 	'enableCacheDisplay': false,
 	'enableRendererStats': false,
 
 	'apiKey': 'put-your-google-api-key-here',
 	'errorTarget': 6,
-	'errorThreshold': 60,
 	'maxDepth': 15,
 	'loadSiblings': true,
 	'stopAtEmptyTiles': true,
-	'displayActiveTiles': false,
 	'resolutionScale': 1.0,
 
-	'up': '+Y',
 	'displayBoxBounds': false,
-	'colorMode': 0,
 	'reload': reinstantiateTiles,
 
 };
@@ -230,12 +204,12 @@ function reinstantiateTiles() {
 
 			setupTiles();
 
-			} )
-			.catch( err => {
+		} )
+		.catch( err => {
 
-				console.error( 'Unable to get gmaps tileset:', err );
+			console.error( 'Unable to get gmaps tileset:', err );
 
-			} );
+		} );
 
 }
 
@@ -250,18 +224,11 @@ function init() {
 	renderer.outputEncoding = sRGBEncoding;
 
 	document.body.appendChild( renderer.domElement );
-	renderer.domElement.tabIndex = 1;
 
 	camera = new PerspectiveCamera( 60, window.innerWidth / window.innerHeight, 1, 1600000 );
 	camera.position.set( 7326000, 10279000, -823000 );
 	cameraHelper = new CameraHelper( camera );
 	scene.add( cameraHelper );
-
-	const geom = new SphereGeometry(1, 32, 8);
-	const mat = new MeshBasicMaterial( { color: 0xff0000 } );
-	const mesh = new Mesh(geom, mat);
-	mesh.position.x = 4;
-	scene.add(mesh);
 
 	// controls
 	controls = new OrbitControls( camera, renderer.domElement );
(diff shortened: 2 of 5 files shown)


example/googleMapsExample.js
```diff
@@ -0,0 +1,604 @@
...
+	renderer.setPixelRatio( window.devicePixelRatio );
+	renderer.setSize( window.innerWidth, window.innerHeight );
+	renderer.setClearColor( 0x151c1f );
+	renderer.outputEncoding = sRGBEncoding;
+
+	document.body.appendChild( renderer.domElement );
+	renderer.domElement.tabIndex = 1;
+
+	camera = new PerspectiveCamera( 60, window.innerWidth / window.innerHeight, 1, 4000 );
+	camera.position.set( 10, 10, 10 );
```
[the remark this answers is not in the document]
aiden-jeffrey: oh of course! thanks.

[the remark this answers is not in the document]
aiden-jeffrey: I've added some temporary positioning of the camera and it works!
image
It doesn't progressively load the more detailed tiles as I zoom in. Do you have any ideas why this is, or some general pointers on how to achieve the full DIY Google Earth experience?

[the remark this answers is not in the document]
gkjohnson: Awesome!

> It doesn't progressively load the more detailed tiles as I zoom in.

First thing that comes to mind is that the LRUCache is full. I'd increase the maxSize and maybe the minSize to a significantly larger value to ensure non of the tiles are prevented from loading. The goal of the LRUCache is to prevent the page from storing too much memory without ever releasing tiles but in some cases it needs to be increased.

Something like the following:

```js
tiles.lruCache.minSize = 3000;
tiles.lruCache.maxSize = 5000;
```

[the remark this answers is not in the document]
aiden-jeffrey: I've pushed up your suggestion - it doesn't resolve the issue. I'll dig into source code tomorrow, but maybe I'm not getting inside the bounding boxes dues to the scale of the model, and that's why the more detailed tiles aren't loading? Is that roughly how the renderer works (works out which nodes in the tileset are within some threshold of the camera position and render them)?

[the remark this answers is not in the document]
gkjohnson: I'm seeing that it is loading more tiles when zooming - but it it's trying to load tons of them because the camera frustum is encompassing the full radius of the globe. I'll look into making some adjustments that simplify and provide more of a maps like experience over the next couple days.

example/googleMapsExample.js
```diff
@@ -0,0 +1,605 @@
...
+				const rotationToNorthPole = rotationBetweenDirections( surfaceDirection, up );
+
+				tiles.group.quaternion.x = rotationToNorthPole.x;
+				tiles.group.quaternion.y = rotationToNorthPole.y;
+				tiles.group.quaternion.z = rotationToNorthPole.z;
+				tiles.group.quaternion.w = rotationToNorthPole.w;
+
+				tiles.group.position.y = - distanceToEllipsoidCenter;
+
+			};
```
gkjohnson, line 226: What is this callback implementation for?

aiden-jeffrey: I'm adapting your cesium sample while I get my head around the API / the 3Dtile format. I might not have pulled out everything non-essential.

example/googleMapsExample.js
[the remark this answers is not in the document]
aiden-jeffrey: I was going to make the frustum far plane progressively contract as you zoom in - do you think that'll be enough to stop the spurious loads?

gkjohnson:
I'm also noticing some odd rotations of the bounding boxes relative to the tiles, here:

image

It looks like the bounding boxes that converge on the left side should align with the poles 🤔

example/googleMapsExample.js
gkjohnson: That's in part what I was thinking, as well. I think that should help.

hybridherbst:
Just wanted to raise a bug that fileType can be "glTF" which is not handled right now as I was looking into 3D Tiles. Great that there's already progress here!

gkjohnson:
@aiden-jeffrey I've just pushed some changes that removes some of the unused fields from the example page and fixes the rotation bug mentioned above.

Here's what I would try to round the demo out and get a first pass at a more google maps-like experience:
- Use three.js' [MapControls](https://threejs.org/docs/?q=MapCon#examples/en/controls/MapControls) which pans along the camera's "up" direction rather than in screen space.
- Use the "Ellipsoid" class to derive latitude, longitude, and surface normal from the camera position position so we can adjust the plane the map controls pans on based on the position.
- Fix the controls target to be the WGS84 ellipsoid surface so we scroll on a reasonable plane.

Longer term:
- Use raycasting to adjust the camera / controls target height based on loaded geometry / mountains.
- Adjust MapControls to zoom in to the cursor position.
- When zooming out transition to an orbital camera


I'm not sure if three.js' MapControls will do everything we want it to here but it's worth a try. @aiden-jeffrey are you interested in giving some of this a shot?

aiden-jeffrey:
> I'm not sure if three.js' MapControls will do everything we want it to here but it's worth a try. @aiden-jeffrey are you interested in giving some of this a shot?

Yes I will work on this today - I got sidetracked this morning onto something else.

aiden-jeffrey:
@gkjohnson - this is getting there - need to tidy it up considerably and rethink the frustum plane contraction, but it sort of works.

snowzurfer:
> @gkjohnson - this is getting there - need to tidy it up considerably and rethink the frustum plane contraction, but it sort of works.

@aiden-jeffrey id love to help, I've take a look at the latest commit and it looks like it would be useful to be able to zoom into the mouse position and detect the terrain.

What things are already on your radar vs ones I could help with?

aiden-jeffrey:
> What things are already on your radar vs ones I could help with?

@snowzurfer - sure, feel free. This is really messy currently, I will spend some time over the weekend getting this into a review-able state. I've added some placeholder functionality similar to what you suggest - if you double click, it will move the orbit camera's target to the position on the globe, under the mouse.

You could take that and have it trigger just on mousemove. I think we probably want to move the camera position by the same amount we move the target though, as the camera behaviour might feel inconsistent otherwise.

The other thing that probably needs rethinking is the frustum correction as we zoom in. When you launch, the frustum is at 0,0,0 (the center of the planet), and as you scroll in, I do a raycast check at the extents of the screen; once I see 4 successful raycast intersections, I take the largest distance (I probably should project the intersections onto the camera direction vector) and use that as the `camera.far` setting. This works ok when the camera is square on with the surface, but it breaks down once you rotate, and so you end up downloading a load of unneeded  tiles.

andreasrosdal:
Can I use this to display historic famous buildings from Google Earth in the game Fciv.net? https://github.com/fciv-net/fciv-net

snowzurfer:
@gkjohnson I'm trying to also understand why it looks like when you're at a given altitude above the earth model, it looks like unless you set `loadSiblings` to `false`, the renderer keeps on loading finer and finer tiles, even though it looks like it "shouldn't" as the ones it has already displayed would be enough at that altitude.

I wonder if this is a consequence of this specific 3D tileset being a sphere, rather than just a 3D cartesian infinite plane? I.e. was the AMMOS renderer originally developed to handle tilesets that are just on an infinite 3D cartesian plane VS on a globe, maybe?

Just a guess; in any case, would you know why that's happening? Setting `loadSiblings` to false seems to stop this "issue" and make the renderer stop loading finer and finer tiles when you're not moving the camera and you're above the earth.

gkjohnson:
> I'm trying to also understand why it looks like when you're at a given altitude above the earth model, it looks like unless you set loadSiblings to false, the renderer keeps on loading finer and finer tiles, even though it looks like it "shouldn't" as the ones it has already displayed would be enough at that altitude.

You'll have to provide numbers and camera positions otherwise I can't see what you're referring to.

> I.e. was the AMMOS renderer originally developed to handle tilesets that are just on an infinite 3D cartesian plane VS on a globe, maybe?

No it just treats the tiles as a series of bounding volumes and computes distances and  geometric error to determine whether to load data. The one thing relevant to a globe that's not necessarily fully supported with closest distance computation and frustum culling - instead [spheres and box bounds are generated](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/ff7d9be3fc07e9ed89f432432460f0bcd3689d33/src/three/TilesRenderer.js#L542-L572) to use, which are less precise. If anyone is interested in taking on improving that it would be great!

example/googleMapsExample.js
```diff
@@ -0,0 +1,465 @@
...
+			} else {
+				return undefined;
+			}
+		}
+
+		const y = findSurface( top );
+		const xz = findSurface( side );
+
+		if ( y !== undefined && xz !== undefined ) {
+			ellipsoid.radius.set( xz, y, xz );
```
dbuck, line 334: you could use the WGS constants to start from?
```
import { WGS84_RADIUS, WGS84_HEIGHT } from '../src/constants.js';
let ellipsoid = new Ellipsoid(WGS84_RADIUS, WGS84_RADIUS, WGS84_HEIGHT);
```

It's not perfect depending on the elevation of the area you're looking at, but should be good enough to initialize the camera controls if that was the main thing you're trying to do?

gkjohnson:
How do we feel about merging this so others can start making PRs into the example to get it polished? We can make an issue to track some of the action items if people want to take them on.

example/googleMapsExample.js
```diff
@@ -0,0 +1,465 @@
...
+	PlaneGeometry,
+	Plane,
+	PlaneHelper,
+	DoubleSide,
+	SphereGeometry,
+	MeshBasicMaterial,
+	Mesh,
+} from 'three';
+import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls.js';
+import { BufferGeometryUtils } from 'three/examples/jsm/utils/BufferGeometryUtils.js';
```
mrdoob, line 25: These two imports are not unused.

example/googleMapsExample.js
```diff
@@ -0,0 +1,465 @@
...
+animate();
+
+function setupTiles() {
+
+	tiles.fetchOptions.mode = 'cors';
+
+	// Note the DRACO compression files need to be supplied via an explicit source.
+	// We use unpkg here but in practice should be provided by the application.
+	const dracoLoader = new DRACOLoader();
+	dracoLoader.setDecoderPath( 'https://unpkg.com/three@0.123.0/examples/js/libs/draco/gltf/' );
```
mrdoob, line 80: For anyone trying to use this with the latest three.js this should be:

```js
dracoLoader.setDecoderPath( 'https://unpkg.com/three@0.152.0/examples/jsm/libs/draco/' );
```

example/googleMapsExample.html
```diff
@@ -0,0 +1,53 @@
...
+        </style>
+    </head>
+    <body>
+		<div id="info">
+			<div>
+				Google Maps 3D tiles example.
+				<br/>
+			</div>
+		</div>
+        <script src="./googleMapsExample.js"></script>
```
mrdoob, line 51: For anyone trying to run this without having to deal with bundlers... 😇

```html
<script type="importmap">
	{
	  "imports": {
		"three": "https://unpkg.com/three@0.152.0/build/three.module.js",
		"three/": "https://unpkg.com/three@0.152.0/"
	  }
	}
</script>   
<script type="module" src="./googleMapsExample.js"></script>
```

example/googleMapsExample.js
gkjohnson: Yeah we should update the three.js version used for demos in the project.

aiden-jeffrey:
I ended up forking the three.js OrbitControls; the way it's written, it's hard to override / extend some of the internal functions (like the panUp|Left or rotate functions).

The big TODOs I see are:

1. Add in the attributions detailed here: https://developers.google.com/maps/documentation/tile/create-renderer (I couldn't work out which callback to use to access the raw tile response)
2. Adapt the GlobeOrbitCamera to use ellipsoid surface normal as the rotation axis (again, I tried and failed to do this - some TODOs / commented out code left in)
3. The rest of the longer term goals laid out by @gkjohnson above
4. Find a better strategy for updating the frustum plane

aiden-jeffrey:
@gkjohnson - I'm going to try and use this for my own project now. I only want to display tiles within some radius of a given lat/long, and to a given detail depth (i.e. independent of camera position). Is this use case already supported in your API?

aiden-jeffrey:
> @gkjohnson - I'm going to try and use this for my own project now. I only want to display tiles within some radius of a given lat/long, and to a given detail depth (i.e. independent of camera position). Is this use case already supported in your API?

Do I have to setCamera with a dummy fixed camera?

gkjohnson:
Thanks @aiden-jeffrey! The camera still needs a bit of work but it's good to have a starting point - hopefully we can remove the need to create a completely custom control scheme. Can you fix the lint issues and we can get this fixed?

> I only want to display tiles within some radius of a given lat/long, and to a given detail depth (i.e. independent of camera position)

You can't limit which tiles will be loaded to a specific region. But you can limit the where the camera looks and can move to.

gkjohnson:
I've made a new issue for those who want to help refine and polish the demo page further in #331. Feel free to chime in and claim a feature or subset of one it would be great!

Issue #331, opened 2023-05-23
Google Maps Tiles demo needed improvements


gkjohnson:
Here's a list of features needed to improve

**Improved Globe Orbit Controls**
- ~Use three.js' [MapControls](https://threejs.org/docs/?q=MapCon#examples/en/controls/MapControls) which pans along the camera's "up" direction rather than in screen space.~ #346
- ~Use the "Ellipsoid" class to derive latitude, longitude, and surface normal from the camera position position so we can adjust the plane the map controls pans on based on the position.~ #46
- ~Fix the controls target to be the WGS84 ellipsoid surface so we scroll on a reasonable plane.~ #346
- ~Use raycasting to adjust the camera / controls target height based on loaded geometry / mountains.~ #346
- When zooming out transition to an orbital camera

**Zoom Controls**
- Adjust MapControls to zoom in to the cursor position.

**Camera View Range**
- ~Scale the camera near and far so only necessary tiles are loaded - and tiles on the opposite side of the globe are not.~ #346
  - Perhaps keeping the camera far fixed at the perpendicular plane at the earth's center is good enough?

**Metadata**
- ~Display data copyrights based on visible tiles~ #337
  - See tile glb metadata
  - https://developers.google.com/maps/documentation/tile/create-renderer
- ~Display current latitude and longitude~ #338

**Jump to location**
- Use window hash to specify which latitude and longitude to jump to

gkjohnson:
> Note that MapControls projects the camera up onto the y=0 plane and pans along that vector.

Did you try adjusting camera.up? The docs say that MapControls pan orthogonal to the cameras up direction:

https://threejs.org/docs/?q=map#examples/en/controls/MapControls.screenSpacePanning

aiden-jeffrey:
> Did you try adjusting camera.up? The docs say that MapControls pan orthogonal to the cameras up direction:

Yes, but the [`panLeft`](https://github.com/mrdoob/three.js/blob/master/examples/jsm/controls/OrbitControls.js#L396) function needs patching to support the same functionality, so I don't think there's any way around forking. See my changes here: https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/example/GlobeOrbitControls.js#L412

gkjohnson:
> Also, I don't think camera.up ever gets updated (it's always normalised y vector).

[...]

PR #337, opened 2023-05-31
Google Earth Demo: Add support for retrieving credits

gkjohnson:
Related to #331 

image

- Add credits support
- Fix retrieval of session id

cc @snowzurfer I'm thinking we may want to make a "TilesGlobe" class or something that makes adding and interacting with the globe easier while still retaining flexibility of an individual object to drop in and a separate of camera controls. What do you think?

### DIFF
--- a/example/googleMapsExample.html
+++ b/example/googleMapsExample.html
@@ -39,6 +39,16 @@
 				color: white;
 				pointer-events: all;
 			}
+
+			#credits {
+				position: absolute;
+				bottom: 0;
+				left: 0;
+				color: white;
+				font-size: 10px;
+				opacity: 0.5;
+				padding: 5px;
+			}
         </style>
     </head>
     <body>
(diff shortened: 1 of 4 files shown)


snowzurfer:
@gkjohnson yeah I think that a good direction could be:

* Create controls that allow interacting with a globe rather than a flat map; i.e. kind of like `OrbitControls` but with the addition that it takes into account the terrain when zooming and rotating around a point

* Have functions to convert between geocoordinates and cartesian easily

Once those two are in place, then we can do a lot of cool stuff!

gkjohnson:
@snowzurfer Great thanks for taking a look at the controls. I'm happy to host them in this project so we can have the core TilesRenderer stuff and then API-specific implementations like those needed for `CesiumIonRenderer` or `GoogleTilesRenderer` along with `GeoUtils`,`GlobeControls`, and whatever ever might be needed. I think with this we can strike a good balance between convenience and control.

My next plan is to update the camera to be a bit more useable in the demo so you can get down closer to the surface until your controls are ready. And I'll be looking at moving some of the convenience geo util classes I've written for the demo to the core project.

Only other thing that might be nice would be a good atmosphere shader and maybe lighting? Not sure if there's already one out there 🤔 It's possible a basic fresnel shader material sphere is good enough.

PR #340, opened 2023-06-01
fix: use bounding sphere when bounding box not available

jesse-small:
- getBounds now returns a bounding box from the bounding sphere when a cached bounding box is not available
- calculateError now correctly applies the inverse transform matrix to the bounding sphere before calculating distance

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -23,7 +23,6 @@ const INITIAL_FRUSTUM_CULLED = Symbol( 'INITIAL_FRUSTUM_CULLED' );
 const tempMat = new Matrix4();
 const tempMat2 = new Matrix4();
 const tempVector = new Vector3();
-const tempSphere = new Sphere();
 const vecX = new Vector3();
 const vecY = new Vector3();
 const vecZ = new Vector3();
@@ -913,7 +912,6 @@ export class TilesRenderer extends TilesRendererBase {
 		const boundingSphere = cached.sphere;
 		const boundingBox = cached.box;
 		const boxTransformInverse = cached.boxTransformInverse;
-		const transformInverse = cached.transformInverse;
 		const useBox = boundingBox && boxTransformInverse;
 
 		let maxError = - Infinity;
@@ -949,13 +947,10 @@ export class TilesRenderer extends TilesRendererBase {
 
 				} else {
 
-					tempVector.applyMatrix4( transformInverse );
-					tempSphere.copy( boundingSphere );
-					tempSphere.applyMatrix4( transformInverse );
 					// Sphere#distanceToPoint is negative inside the sphere, whereas Box3#distanceToPoint is
 					// zero inside the box. Clipping the distance to a minimum of zero ensures that both
 					// types of bounding volume behave the same way.
-					distance = Math.max( tempSphere.distanceToPoint( tempVector ), 0 );
+					distance = Math.max( boundingSphere.distanceToPoint( tempVector ), 0 );
 
 				}
 
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -170,6 +170,17 @@ export class TilesRenderer extends TilesRendererBase {
 
 		} else {
 
+			const boundingSphere = cached.sphere;
+
+			if ( boundingSphere ) {
+
+				boundingSphere.getBoundingBox( box );
+				matrix.identity();
+
+				return true;
+
+			}
+
 			return false;
 
 		}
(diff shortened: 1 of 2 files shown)


src/three/TilesRenderer.js
```diff
@@ -133,6 +134,16 @@ export class TilesRenderer extends TilesRendererBase {
 
 		} else {
 
+			const boundingSphere = cached.sphere;
+
+			if ( boundingSphere ) {
```
jesse-small, line 139: If there is no cached bounding box, return a bounding box based on the cached bounding sphere

src/three/TilesRenderer.js
```diff
@@ -939,10 +950,12 @@ export class TilesRenderer extends TilesRendererBase {
 				} else {
 
 					tempVector.applyMatrix4( transformInverse );
+					tempSphere.copy( boundingSphere );
+					tempSphere.applyMatrix4( transformInverse );
 					// Sphere#distanceToPoint is negative inside the sphere, whereas Box3#distanceToPoint is
 					// zero inside the box. Clipping the distance to a minimum of zero ensures that both
 					// types of bounding volume behave the same way.
-					distance = Math.max( boundingSphere.distanceToPoint( tempVector ), 0 );
+					distance = Math.max( tempSphere.distanceToPoint( tempVector ), 0 );
```
jesse-small, line 958: Since we save a transformed bounding sphere this was causing the calculate error to always be wrong for bounding spheres when a transform matrix was present. This applies the inverseTransform to the bounding sphere before calculating the distance.

src/three/TilesRenderer.js
gkjohnson: Great thanks! We should do the same for the `getOrientedBounds` function.

src/three/raycastTraverse.js
```diff
@@ -87,45 +87,47 @@ export function raycastTraverseFirstHit( root, group, activeTiles, raycaster ) {
...
+				continue;
 
-				if ( Math.abs( Math.max( _vec2.x - _vec2.y, _vec2.x - _vec2.z ) ) > 1e-6 ) {
+			}
 
-					console.warn( 'ThreeTilesRenderer : Non uniform scale used for tile which may cause issues when raycasting.' );
+		} else {
 
-				}
+			_mat.invert();
```
gkjohnson, line 98: The logic changes in this raycast first hit function don't look right to me. If we want to accommodate bounding spheres we need to try to intersect both the sphere and the obb. If either ray misses we should skip that child and not add it to the array to check later. Otherwise we need to take the shortest distance between the two hits and save it in the array so we can use it for sorting and check the nearest bound first.

It might be best to make these raycast changes in a separate PR so they're easier to follow.

gkjohnson (commented):
Awesome thank you! I've left a few comments.

src/three/TilesRenderer.js
jesse-small: I was wondering if you could help me with the orientedBounds logic.
I'm not sure how to calculate the obb matrix based on the boundingSphere. Any ideas or recommendations?

src/three/raycastTraverse.js
```diff
@@ -70,7 +70,7 @@ export function raycastTraverseFirstHit( root, group, activeTiles, raycaster ) {
 
 			_sphere.copy( sphere );
 			_sphere.applyMatrix4( _mat );
-			if ( ! raycaster.ray.intersectsSphere( _sphere ) ) {
+			if ( ! raycaster.ray.intersectsSphere( _sphere, _vec ) ) {
```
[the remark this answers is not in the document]
jesse-small: Good catch, I was changing my library locally for testing and this should actually be changed to [intersectSphere](https://threejs.org/docs/index.html#api/en/math/Ray.intersectSphere).

src/three/raycastTraverse.js
jesse-small: I will remove the raycast changes from this PR to keep is simple and submit the raycast changes as a new PR.

src/three/TilesRenderer.js
gkjohnson: Yup! Computing an oriented bounding box I think is more work than is needed. I'm thinking in this case it's enough to just produce the same AABB and identity matrix. At least then the use won't get an empty result when they need a valid bounds to work with. In the future I think these functions need to be reconsidered so users can make sure they're using the natively defined bounds if necessary, I think.

jesse-small:
I added the raycasting logic along with an explanation to this PR https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/347

gkjohnson:
Great thank you!

PR #346, opened 2023-06-05
Demo update

gkjohnson:
Related to #331 

- Remove globe controls
- Add slightly modified OrbitControls that supports a more controllable pan plane
- Adjust view frustum based on camera position to remove artifacts
- Adjust the camera target based on raycast position
- Bump the queue sizes up to improve download speeds

**TODO**
- Update to latest three.js
- Add orbit
- Fix lat lon -> vector and back functionality
- Add lat lon hash to jump to a position
- Add tilt to the camera

image


gkjohnson:
cc @snowzurfer - looks like the orbit controls plane panning is more limited than I thought but I've got something going that at least allows you to zoom in closer to the surface with the current controls. As you've suggested - a more customized version seems like it's going to be needed.

PR #347, opened 2023-06-05
fix: if bounding box not available use bounding sphere for raycast hit

jesse-small:
The following code runs as follows:

- Check bounding sphere for raycast hit
  - If raycast miss, continue to next tile
- Check if bounding box exists
  - If bounding box exists, check for raycast hit
    - If raycast miss, continue to next tile
- If we haven't continue to next tile that means either 1) bounding sphere had a raycast hit and bounding box does not exist or 2) bounding box exists and had a hit

Now we can run the normal logic of saving a raycast hit only using the bounding sphere raycast hit if the bounding box does not exist (if bounding box exists we will still use that as primary hit detection).

Previously if a bounding box did not exist we would assume the raycast missed. The logic now states that if a bounding box **does not** exist, the bounding sphere is assumed to be the tightest bounding volume and a valid intersect results in a hit. If the bounding box **does** exist that will still require an intersect in the bounding box for a valid hit, even if the bounding sphere has an intersect.

### DIFF
--- a/src/three/raycastTraverse.js
+++ b/src/three/raycastTraverse.js
@@ -70,7 +70,7 @@ export function raycastTraverseFirstHit( root, group, activeTiles, raycaster ) {
 
 			_sphere.copy( sphere );
 			_sphere.applyMatrix4( _mat );
-			if ( ! raycaster.ray.intersectsSphere( _sphere ) ) {
+			if ( ! raycaster.ray.intersectSphere( _sphere, _vec ) ) {
 
 				continue;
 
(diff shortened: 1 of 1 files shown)


jesse-small:
@gkjohnson that makes sense to me I am more than happy to add those changes. I will push up the new logic sometime today! Always appreciate the help you provide!

jesse-small:
@gkjohnson I have added a new utility function to calculate the distance just to simplify the code.
Now if the bounding sphere raycast hits, it will calculate the distance and add it to the array, if the bounding box exists and raycast hits it will also add it to the sorted array.

I kept the logic that if the bounding sphere exists and raycast misses, we quit early and continue to the next tile.

jesse-small:
@gkjohnson was able to test it this morning using my sphere volume set and it worked as expected. I was able to test on both sphere volume and box volume sets successfully.

gkjohnson:
Thanks! Just noticed it should have been using `max` instead of `min` so I made that change. Thanks again for your help!

PR #350, opened 2023-06-06
Set script type to module in googleMapsExample.html

jo-chemla:
Set script type to module in googleMapsExample.html page
This should fix the `Cannot use import statement outside a module (at googleMapsExample.js:1:1)` at [this url](https://nasa-ammos.github.io/3DTilesRendererJS/example/googleMapsExample.html)


gkjohnson:
Great, thank you!

PR #352, opened 2023-06-06
Fix lat lon transformation

gkjohnson:
Lat lon in the hash now works - loading a page that jumps to mt fuji:

image

### DIFF
--- a/example/googleMapsExample.js
+++ b/example/googleMapsExample.js
@@ -302,10 +302,9 @@ function updateControls() {
 function updateHash() {
 
 	const res = {};
-	const pos = controls.target.clone();
-	GeoUtils.swapToGeoFrame( pos );
-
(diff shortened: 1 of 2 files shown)


aiden-jeffrey:
@gkjohnson - do you have any tips on doing the reverse of this (i.e. transforming the globe so that the surface at a given lat/long is at the origin, with the surface coplanar with the y=0 plane)? I'm trying to use lookAt, something like:
```
	WGS84_ELLIPSOID.getCartographicToPosition( lat * MathUtils.DEG2RAD, lon * MathUtils.DEG2RAD, 0, position );

	tiles.group.up.set( 0, 0, 1 ); // this is almost certainly wrong - just to show that I've tried different "up" values
	tiles.group.lookAt( temp.negate() );
```
I'm not sure if I should be using some variant of the fixed `getPositionToSurfacePoint` rather than `getCartographicToPosition`.

PR #357, opened 2023-06-12
Progressively preprocess the child nodes

### LINKED ISSUE #355: Improve "internal tileset" preprocess timing
Time it to see if it's a problem. If it's slow then instead of doing it all at once only process the tiles on first visit when needed.

gkjohnson:
Fix #355

**TODO**
- Test / validate all examples
- ~Try it with empty nodes / skipping empty nodes~
- ~Consider other "traverse" cases though this is slow~
  - Perhaps DebugTilesRenderer can only trigger the preprocessing when it's needed?
- ~Ensure we're preprocessing in all the right spots and before everything is needed (possibly in "traverse"?)~
- ~Raycasting seems to traverse everything if update has not been called, yet~

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,7 +539,7 @@ Disposes of all the tiles in the renderer. Calls dispose on all materials, textu
 
 _extends [TilesRenderer](#TilesRenderer)_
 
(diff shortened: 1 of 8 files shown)


gkjohnson:
**Before**
- Up to and over 100ms taken on a single frame to pre process the loaded nested tilesets
- Up to 5 tilesets loaded and process on a single frame (possibly more in some cases)

PR #362, opened 2023-06-15
Update googleMapsExample.js with default apiKey

jo-chemla:
Add default `apiKey` into params with value `put-your-api-key-here`. Otherwise gui.add wont add that parameter to the controls and it cannot be edited.


gkjohnson:
Oh thanks! Looks like I accidentally removed that.

PR #374, opened 2023-06-19
Fix points check in obb frustum function

### LINKED ISSUE #369: Missing tile in Google Tiles example
Hard to see but in the bottom right one building piece will consistently pop in and out even after refreshing. Full screen the browser page, rotate the camera to the lowest angle, and rotate until one building pops out and then back in. Zoom out shows the building correctly.

Possibly related to #155

Is it a frustum culling issue? Error calculation?

image

image

gkjohnson:
Fix #369

### DIFF
--- a/src/three/math/OBB.js
+++ b/src/three/math/OBB.js
@@ -48,7 +48,7 @@ export class OBB {
 
 			const plane = planes[ i ];
 			let maxDistance = - Infinity;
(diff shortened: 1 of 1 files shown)


gkjohnson:
**Tiles Loaded after fix**

| | Before | After | % |
|---|---|---|---|
| cached tiles | 1408 | 900 | ~36% |
| visible tiles | 787 | 414 | ~47% |

PR #381, opened 2023-06-20
fix: Invalid import of constants

aiden-jeffrey:
I'm not sure if these type files are auto generated - I couldn't find a package.json script for it.

### DIFF
--- a/src/three/math/Ellipsoid.d.ts
+++ b/src/three/math/Ellipsoid.d.ts
@@ -1,9 +1,5 @@
 import { Vector3, Matrix4 } from 'three';
 
-export const WGS84_RADIUS;
-export const WGS84_FLATTENING;
-export const WGS84_HEIGHT;
-
 export class Ellipsoid {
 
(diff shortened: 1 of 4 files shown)


src/three/math/Ellipsoid.d.ts
```diff
@@ -1,9 +1,5 @@
 import { Vector3, Matrix4 } from 'three';
 
-export const WGS84_RADIUS;
-export const WGS84_FLATTENING;
-export const WGS84_HEIGHT;
-
```
gkjohnson, line 6: Thanks for the catch - can we define a "constants.d.ts" file that exports these and add it to the index.d.ts file, as well?

gkjohnson, line 6: Thanks for the catch! Can we add these to a constants.d.ts file and export that from index.d.ts, as well, then?

gkjohnson: Thanks!

PR #385, opened 2023-06-23
Store google-api-key also for googleMapsAerial.js via localStorage

jo-chemla:


gkjohnson:
Great idea - thank you!

Issue #426, opened 2023-12-21
the model disappears when the rotated map is greater than 180 degrees


xiaxiangfeng:
**Describe the bug**

After adding 3Dtiles to mapboxgl using 3DFilesRenderJS, the model disappears when the rotated map is greater than 180 degrees

**To Reproduce**

1. Using mapboxgl to customize layers and add 3D tiles
2. Use the right mouse button to rotate the map
3. When the rotation is greater than 180 degrees, the model disappears

***Code***
The address of the code on GitHub

https://github.com/xiaxiangfeng/3DTilesRendererJS.Test/tree/main/docs

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>Display a map on a webpage</title>
    <meta
      name="viewport"
      content="initial-scale=1,maximum-scale=1,user-scalable=no"
    />
    <link
      href="https://api.mapbox.com/mapbox-gl-js/v2.13.0/mapbox-gl.css"
      rel="stylesheet"
    />
    <script src="https://api.mapbox.com/mapbox-gl-js/v2.13.0/mapbox-gl.js"></script>
    <script
      async
      src="https://unpkg.com/es-module-shims@1.3.6/dist/es-module-shims.js"
    ></script>
    <style>
      body {
        margin: 0;
        padding: 0;
      }

      #map {
        position: absolute;
        top: 0;
        bottom: 0;
        width: 100%;
      }
    </style>
  </head>

  <body>
    <div id="map"></div>
    <script type="importmap">
      {
        "imports": {
          "three": "https://unpkg.com/three@0.135.0/build/three.module.js",
          "three/examples/jsm/": "https://unpkg.com/three@0.140.0/examples/jsm/"
        }
      }
    </script>

    <script type="module">
      import * as THREE from "three";
      import { DebugTilesRenderer as TilesRenderer } from "./3DTilesRendererJS/index.js";

      window.THREE = THREE;

      var origin = [113.37796671195645, 22.994043025862794];

      mapboxgl.accessToken =
        "pk.eyJ1IjoieGlheGlhbmdmbmVnIiwiYSI6ImNscWYyNjU3azByd3gya3JxOTVrc2NkY3UifQ.362MspMnDi9ZGH-D6P1CtQ";
      const map = new mapboxgl.Map({
        container: "map",
        style: "mapbox://styles/mapbox/dark-v9",
        center: origin,
        zoom: 17.86614600777933,
[...]

PR #435, opened 2023-12-29
Example for using with Mapbox GL JS

xiaxiangfeng:
add a Example for using with Mapbox GL JS #426


gkjohnson (commented):
Thank you! I've added a few notes - and do you mind adding comments to the code? Especially the "mapboxExampleCamera". It's not easily clear what it's for or what it does from looking at the code. Is it the case that it needs to be a class like this?

example/mapboxExampleCamera.js
```diff
@@ -0,0 +1,287 @@
...
+
+		const topHalfSurfaceDistance =
+      ( Math.sin( fovAboveCenter ) * cameraToSeaLevelDistance ) /
+      Math.sin(
+      	utils.clamp(
+      		Math.PI - groundAngle - fovAboveCenter,
+      		0.01,
+      		Math.PI - 0.01
+      	)
+      );
```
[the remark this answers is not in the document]
xiaxiangfeng: ok

example/mapboxExample.js
```diff
@@ -0,0 +1,145 @@
...
+
+const origin = [ 116, 35 ];
+
+// TO MAKE THE MAP APPEAR YOU MUST
+// ADD YOUR ACCESS TOKEN FROM
+// https://account.mapbox.com
+// This is my private key, it is prohibited to use in the project
+window.mapboxgl.accessToken =
+        'pk.eyJ1IjoieGlheGlhbmdmbmVnIiwiYSI6ImNscTRzc245NTA5c3cya3BhdXA4MDY0NWQifQ._9T_J9z7Hvcd00nQL3UuSw';
+const map = new window.mapboxgl.Map( {
```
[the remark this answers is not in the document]
xiaxiangfeng: This is a key created by my own account, which can be used normally. I just don't want anyone to use my key in their project. If someone uses it, they will have to bear the risk of failure themselves

example/mapboxExample.js
```diff
@@ -0,0 +1,145 @@
...
+
+			tilesGroup = new THREE.Group();
+
+			world.add( tilesGroup );
+
+			new CameraSync( map, camera, world );
+
+			const url = 'https://xiaxiangfeng.github.io/3DTilesRendererJS.Test/baowei21/tileset.json';
+
+			tiles = new TilesRenderer( url );
```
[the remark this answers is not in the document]
xiaxiangfeng: This is the tileset I generated using the tool, and all the b3DM files are placed in it https://xiaxiangfeng.github.io/3DTilesRendererJS.Test/ Under the directory, it can be downloaded

xiaxiangfeng:
> Thank you! I've added a few notes - and do you mind adding comments to the code? Especially the "mapboxExampleCamera". It's not easily clear what it's for or what it does from looking at the code. Is it the case that it needs to be a class like this?

You can add comments or modify some code

example/mapboxExample.js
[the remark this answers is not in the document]
xiaxiangfeng: This comment can be deleted

// This is my private key, it is promoted to use in the project

example/mapboxExample.js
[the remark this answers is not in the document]
xiaxiangfeng: Alternatively, you can modify it to another tileset

xiaxiangfeng:
> > Thank you! I've added a few notes - and do you mind adding comments to the code? Especially the "mapboxExampleCamera". It's not easily clear what it's for or what it does from looking at the code. Is it the case that it needs to be a class like this?
> 
> You can add comments or modify some code

The absence of comments in mapboxExampleCamera does make it difficult for people to understand. I can add some comments to make it easier to understand

example/mapboxExample.js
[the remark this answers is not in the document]
gkjohnson: I think it's best to use the same model as the other examples for the moment and require that users provide their own key in the GUI to use the demo. Otherwise it's difficult to track who is responsible for which key and where they've come from.

example/mapboxExample.js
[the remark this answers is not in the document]
gkjohnson: What is the original source of the model, though? I would like to confirm the license and data used for the model.

example/mapboxExample.js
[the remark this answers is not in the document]
xiaxiangfeng: ok

example/mapboxExample.js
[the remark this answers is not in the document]
xiaxiangfeng: https://raw.githubusercontent.com/NASA-AMMOS/3DTilesSampleData/master/msl-dingo-gap/0528_0260184_to_s64o256_colorize/0528_0260184_to_s64o256_colorize/0528_0260184_to_s64o256_colorize_tileset.json

xiaxiangfeng:
modify：1.users provide their own key in the GUI to use the demo

xiaxiangfeng:
modify：
1.users provide their own key in the GUI to use the demo
2.changed tileset url
3.Add comments to the core of the code in the mapboxExampleCamera file

example/mapboxExample.js
```diff
@@ -0,0 +1,176 @@
...
+import CameraSync, {
+	projectToWorld,
+	projectedUnitsPerMeter,
+} from './mapboxExampleCamera.js';
+import { DebugTilesRenderer as TilesRenderer } from '../src/index.js';
+import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
+
+const params = {
+
+	'accessToken': 'pk.eyJ1IjoieGlheGlhbmdmbmVnIiwiYSI6ImNscTRzc245NTA5c3cya3BhdXA4MDY0NWQifQ._9T_J9z7Hvcd00nQL3UuSw',
```
gkjohnson, line 11: It looks like we're still providing a token? I think we should default to a string called "put-your-api-key-here" as the Google Maps example does.

example/mapboxExample.js
```diff
@@ -0,0 +1,176 @@
...
+				world = new THREE.Group();
+				scene.add( world );
+
+				// lights
+				const dirLight = new THREE.DirectionalLight( 0xffffff, 0.8 );
+				dirLight.position.set( 1, 2, 3 );
+				world.add( dirLight );
+
+				const ambLight = new THREE.AmbientLight( 0xffffff, 1 );
+				world.add( ambLight );
```
gkjohnson, line 90: The tileset used is using basic materials which don't use lighting - I think we can remove these for now.

example/mapboxExampleCamera.js
```diff
@@ -0,0 +1,284 @@
...
+	DEG2RAD: Math.PI / 180,
+	RAD2DEG: 180 / Math.PI,
+	EARTH_RADIUS: EARTH_RADIUS,
+	EARTH_CIRCUMFERENCE: 2 * Math.PI * EARTH_RADIUS,
+	EARTH_CIRCUMFERENCE_EQUATOR: EARTH_CIRCUMFERENCE_EQUATOR,
+	FOV_ORTHO: FOV_ORTHO,
+	FOV: FOV,
+	FOV_DEGREES: ( FOV * 180 ) / Math.PI,
+	TILE_SIZE: 512,
+};
```
gkjohnson, line 59: It looks like a number of these constants aren't used in the demo - can we remove the unnecessary ones?

example/mapboxExample.js
```diff
@@ -0,0 +1,176 @@
...
+	rotation.z = a.z;
+	rotation.w = 1 + dir1.clone().dot( dir2 );
+	rotation.normalize();
+
+	return rotation;
+
+}
+
+let map;
+const origin = [ 116, 35 ];
```
gkjohnson, line 40: What are the units fore these values? Are these lat / lon values?

example/mapboxExampleCamera.js
```diff
@@ -0,0 +1,284 @@
+import * as THREE from 'three';
+
+const utils = {
+	clamp: function ( n, min, max ) {
+
+		return Math.min( max, Math.max( min, n ) );
+
+	},
```
gkjohnson, line 8: This function is redundant to three.js' `MathUtils.clamp` function

example/mapboxExampleCamera.js
```diff
@@ -0,0 +1,284 @@
...
+	EARTH_RADIUS: EARTH_RADIUS,
+	EARTH_CIRCUMFERENCE: 2 * Math.PI * EARTH_RADIUS,
+	EARTH_CIRCUMFERENCE_EQUATOR: EARTH_CIRCUMFERENCE_EQUATOR,
+	FOV_ORTHO: FOV_ORTHO,
+	FOV: FOV,
+	FOV_DEGREES: ( FOV * 180 ) / Math.PI,
+	TILE_SIZE: 512,
+};
+
+const Constants = MapConstants;
```
gkjohnson, line 61: This redefinition of the constants object seems unnecessary?

example/mapboxExampleCamera.js
```diff
@@ -0,0 +1,284 @@
...
+
+		this.map = map;
+		this.camera = camera;
+		this.active = true;
+		// We're in charge of the camera now!
+		this.camera.matrixAutoUpdate = false;
+		// Postion and configure the world group so we can scale it appropriately when the camera zooms
+		this.world = world || new THREE.Group();
+		this.world.position.x = this.world.position.y =
+    MapConstants.WORLD_SIZE / 2;
```
gkjohnson, line 105: The rationale for some of these changes should be documented, I think. What's the reasoning for setting the position here? We disable matrix auto update and set the matrix explicitly later so this will never really be used, right?

example/mapboxExampleCamera.js
```diff
@@ -0,0 +1,284 @@
...
+			.on( 'move', function () {
+
+				_this.updateCamera();
+
+			} )
+			.on( 'resize', function () {
+
+				_this.setupCamera();
+
+			} );
```
gkjohnson, line 129: If we use arrow functions there's no need to declare a local scope "_this". ie

```js
this.map
	.on( 'move', () => {

		this.updateCamera();

	} )
```

example/mapboxExampleCamera.js
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	circumferenceAtLatitude( latitude ) {
+
+		return (
+			MapConstants.EARTH_CIRCUMFERENCE *
+      Math.cos( ( latitude * Math.PI ) / 180 )
+		);
+
+	}
```
gkjohnson, line 264: This function looks like it's not used?

example/mapboxExampleCamera.js
```diff
@@ -0,0 +1,284 @@
...
+		const nearZ = Math.max( nz * pitchAngle, nz ); // on changes in the pitch nz could be too low
+
+		const h = t.height;
+		const w = t.width;
+		this.camera.projectionMatrix = utils.makePerspectiveMatrix(
+			t._fov,
+			w / h,
+			nearZ,
+			farZ
+		);
```
gkjohnson, line 213: There shouldn't be a need to manually construct a perspective matrix. These are all values that can be set on the camera itself and "updateProjectionMatrix" can be called to correctly update the projection matrix and ensure the inverse projection matrix are set

example/mapboxExampleCamera.js
```diff
@@ -0,0 +1,284 @@
...
+		this.cameraToCenterDistance = ( 0.5 / Math.tan( this.halfFov ) ) * t.height;
+
+		let pixelsPerMeter = 1;
+		const worldSize = this.worldSize();
+
+		pixelsPerMeter = this.mercatorZfromAltitude( 1, t.center.lat ) * worldSize;
+		const fovAboveCenter = t._fov * ( 0.5 + t.centerOffset.y / t.height );
+
+		// Adjust distance to MSL by the minimum possible elevation visible on screen,
+		// this way the far plane is pushed further in the case of negative elevation.
```
gkjohnson, line 175: I assume MSL means "mean sea level" here?

example/mapboxExampleCamera.js
```diff
@@ -0,0 +1,284 @@
...
+		const minElevationInPixels = t.elevation
+			? t.elevation.getMinElevationBelowMSL() * pixelsPerMeter
+			: 0;
```
gkjohnson, line 178: How can we compute the elevation in pixels? This is a non constant value since we're using a perspective camera. 1 meter far from the camera will be represented by fewer pixels than a meter closer to the camera.

example/mapboxExample.js
xiaxiangfeng: ok

example/mapboxExample.js
xiaxiangfeng: Yes, it's longitude and latitude

example/mapboxExample.js
xiaxiangfeng: ok

example/mapboxExampleCamera.js
xiaxiangfeng: ok

example/mapboxExampleCamera.js
xiaxiangfeng: yes

example/mapboxExampleCamera.js
xiaxiangfeng: Our purpose is to calculate farZ, which all happens before projection

example/mapboxExampleCamera.js
xiaxiangfeng: Unlike the Mapbox GL JS camera, separate camera translation and rotation out into its world matrix If this is applied directly to the projection matrix, it will work OK but break raycasting

example/mapboxExampleCamera.js
xiaxiangfeng: Used in the mergetorZfromAltitude function

example/mapboxExampleCamera.js
xiaxiangfeng: ok

example/mapboxExampleCamera.js
xiaxiangfeng: ok

xiaxiangfeng:
> Thanks for adding some comments! I've written a few more notes. I think the "updateCamera" portion of the code should be able to be simplified quite a bit. It would be best to use as many built in three.js functions to compute the position and camera matrices as possible so it's easier to follow. Accessing direct matrix elements and keeping copies of different matrices around makes the code more difficult to follow. As does keeping some functions around that are unused.
> 
> I'm happy to clean things up a bit afterward but I just want to make sure I understand what's happening since I'm new to MapBox and it's clear to others, as well.
> 
> Also a couple comments about the demo:
> 
> * It looks like the camera world matrix is being updated a frame late which is causing pieces of the tileset to load in delayed which causes a flash. We may need to call "camera.updateMatrixWorld" before triggering "tiles.update()" in the render function.
> * When zooming into the 3d tiles model from above parts of the model disappears - it seems like something isn't being updated correctly:
> 
> image

1. I added camera.updateMatrixWorld but nothing seems to change
2. The tileset used now has terrain, it needs to be placed in its correct position, and mapboxgl needs to be turned on for terrain. We can also set tilesGroup.position.z = 1; make the tileset higher than the mapboxgl ground

example/mapboxExampleCamera.js
gkjohnson: I'm not following. I don't think we should include position or rotation values in the projection matrix. With the values you have here you can set the three.js camera parameters like so:

```js
camera.fov = t._fov;
camera.aspect = w / h;
camera.near = nearZ;
camera.far = farZ;
camera.updateProjectionMatrix();
```

example/mapboxExampleCamera.js
gkjohnson: I don't understand. How can you calculate a distance in pixels if it's not relative to the screen?

gkjohnson:
> I added camera.updateMatrixWorld but nothing seems to change

Interesting - once the code is simplified more I can take a look.

> The tileset used now has terrain, it needs to be placed in its correct position, and mapboxgl needs to be turned on for terrain. We can also set tilesGroup.position.z = 1; make the tileset higher than the mapboxgl ground

Does this explain why the map plane seems to be moving, though? As you zoom in and out and rotate the camera the portion of the terrain being hidden seems to change.

example/mapboxExampleCamera.js
xiaxiangfeng: We need to use the same projectionMatrix as mapboxgl. I modified this place and used the method this.camera.projectionMatrix.elements = t._camera.getCameraToClipPerspective provided by mapboxgl. I found that the calculation methods of projectionMatrix of mapboxgl and threejs are somewhat different. Mapboxgl uses the gl-matrix library.

xiaxiangfeng:
> > I added camera.updateMatrixWorld but nothing seems to change
> 
> Interesting - once the code is simplified more I can take a look.
> 
> > The tileset used now has terrain, it needs to be placed in its correct position, and mapboxgl needs to be turned on for terrain. We can also set tilesGroup.position.z = 1; make the tileset higher than the mapboxgl ground
> 
> Does this explain why the map plane seems to be moving, though? As you zoom in and out and rotate the camera the portion of the terrain being hidden seems to change.

When z is increased, there is still a problem, it seems to be related to depth writing, I am not sure

example/mapboxExampleCamera.js
xiaxiangfeng: I can't seem to answer your question. These codes refer to the source code of mapboxgl. I have modified this place now. farZ directly uses the one provided by mapboxgl.

xiaxiangfeng:
modify：
1. Delete the default accessToken
2. Add camera.updateMatrixWorld
3. Simplify the code of mapboxExampleCamera
4.Delete lights
5.Set tilesGroup.position.z = 1

example/mapboxExampleCamera.js
xiaxiangfeng: > pixels

The pixels here are the sizes under the Mercator projection, which is just a possible value calculated.

example/mapboxExampleCamera.js
gkjohnson: I've taken a look at the gl-matrix code [here](https://glmatrix.net/docs/mat4.js.html#line1474) and the three.js projection matrix code [here]() and I don't see any difference in calculations beyond superficial organization. Can you explain what the differences you saw are?

gkjohnson:
> When z is increased, there is still a problem, it seems to be related to depth writing, I am not sure

I don't this can be merged until it works correctly and I don't have the bandwidth to debug these issues. I recommend asking at the three.js forum to find help with how to properly integrate three.js with mapbox and extract the camera transform and camera frustum separately. Alternatively we could update the tiles renderer project to be resilient to this kind of matrix but I'd prefer to avoid that and this doesn't fix the raycasting issue. Ideally mapbox would provide a separate view and projection matrix (there was a discussion [here](https://github.com/mapbox/mapbox-gl-js/issues/7395) but it didn't go anywhere).

For reference it looks like there's MapBoxGL documentation for three.js integration [here](https://docs.mapbox.com/mapbox-gl-js/example/add-3d-model/) and [threebox](https://github.com/jscastro76/threebox/tree/master) provides some integration utilities.

---

Relating to threebox - it looks like you have have copy and pasted quite a bit of code from the threebox project without attribution. At least from the CameraSync file and other constants. Is this the case? Its your responsibility to properly attribute code and notify projects of where the code has come from if it is not yours.

xiaxiangfeng:
> > When z is increased, there is still a problem, it seems to be related to depth writing, I am not sure
> 
> I don't this can be merged until it works correctly and I don't have the bandwidth to debug these issues. I recommend asking at the three.js forum to find help with how to properly integrate three.js with mapbox and extract the camera transform and camera frustum separately. Alternatively we could update the tiles renderer project to be resilient to this kind of matrix but I'd prefer to avoid that and this doesn't fix the raycasting issue. Ideally mapbox would provide a separate view and projection matrix (there was a discussion [here](https://github.com/mapbox/mapbox-gl-js/issues/7395) but it didn't go anywhere).
> 
> For reference it looks like there's MapBoxGL documentation for three.js integration [here](https://docs.mapbox.com/mapbox-gl-js/example/add-3d-model/) and [threebox](https://github.com/jscastro76/threebox/tree/master) provides some integration utilities.
> 
> Relating to threebox - it looks like you have have copy and pasted quite a bit of code from the threebox project without attribution. At least from the CameraSync file and other constants. Is this the case? Its your responsibility to properly attribute code and notify projects of where the code has come from if it is not yours.

The code in this file comes from https://github.com/jscastro76/threebox/tree/master I have simplified and modified some content

xiaxiangfeng:
First of all, I would like to thank gkjohnson for his support these days. I will close this PR for now. I feel that it cannot solve the existing problems at the moment. I also hope that someone who encounters a similar situation in the future can provide help or complete this PR.

Issue #389, opened 2023-07-04
Add callback for when all loading starts and ends


gkjohnson:
Related to #378 

Such as `tiles.onLoadStart` and `tiles.onLoadComplete`.

This can already be implicitly done by checking the download and parse queue on every frame or in the `onLoadModel` callback (only can detect on end).

arthurschiller:
Hi Garret, thanks for this great library! Could you give a brief example on how this would be accomplished via onLoadModel? How do I know the total amount of models that are going to be loaded so I can compare it with the current count? Sorry in case this is actually obvious. I just hit my quota rate limit on the google tiles API but wanted to give it a go tomorrow.

gkjohnson:
You can check the amount of models currently in the queue to be parsed or downloaded using `downloadQueue.items.size` and `parseQueue.items.size`. `onLoadModel` is called every time a model has been downloaded and parsed so when those fields are both set to 0 you know it's complete.

What's your use case exactly? If you have a compelling one and would like to open a PR to make this more convenient I can point you in the right direction.

arthurschiller:
Thank you, that works! :) 
My use case is similar to here: https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/378
I'm building a location based 3D editor and so I'm fetching tiles at a specified location via an Orthographic Camera.

I could certainly try, but I come from iOS development and am not too familiar with typescript.

gkjohnson:
> I could certainly try, but I come from iOS development and am not too familiar with typescript.

The project is in Javascript if that changes anything. Either way up to you - also nice to have help. Happy to give feedback if you'd like to try.

PR #440, opened 2024-01-06
Added onLoadStart & onLoadComplete callbacks

pizza3:
This pr adds the callback `onLoadStart` & `onLoadComplete` for all loading starts and ends.
#389 

How to test:
I used mars.js example for the test and assigned `onLoadStart` & `onLoadComplete` callbacks to the tiles.

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -68,6 +68,7 @@ export class TilesRenderer extends TilesRendererBase {
 		this.visibleTiles = new Set();
 		this._autoDisableRendererCulling = true;
 		this.optimizeRaycast = true;
+		this.loading = false;
 
 		this.onLoadTileSet = null;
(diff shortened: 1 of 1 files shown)


src/three/TilesRenderer.js
```diff
@@ -675,6 +676,29 @@ export class TilesRenderer extends TilesRendererBase {
...
+			if ( ! this.loading && this.parseQueue.items.length > 0 ) {
+
+				if ( this.onLoadStart ) {
+
+					this.onLoadStart( );
+
+				}
+				this.loading = true;
+
+			}
```
gkjohnson, line 688: It occurs to me that this will not trigger if we're in a case where only one tile has downloaded - because this condition will be checked only once a file has been fully parsed. Perhaps this "onStart" condition should go at the end of the update function, instead?

PR #460, opened 2024-01-18
call updateMatrixWorld after performing setLatLonToYUp

tobias74:
In case other components rely on the matrixWorld of the group they might only get the correct value after the next rendered frame. This might cause short "blips" with one frame being out of place, offset by the distance between the previous and current (lat,lon).


gkjohnson:
Great, thank you!

PR #474, opened 2024-02-08
Fix CesiumIon shadeless MeshStandardMaterial and readme typo preprocessURL 

jo-chemla:
- Fixes black tilesets loaded from Cesium ion as described in [this issue](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/475)
 - Updates README.md typo code blocks with preprocessURL instead of onPreprocessURL

### DIFF
--- a/example/ionExample.js
+++ b/example/ionExample.js
@@ -5,7 +5,6 @@ import {
 	PerspectiveCamera,
 	Vector3,
 	Quaternion,
-	Group,
 	Sphere,
 } from 'three';
 import { FlyOrbitControls } from './src/controls/FlyOrbitControls.js';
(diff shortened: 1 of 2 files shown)


example/ionExample.js
```diff
@@ -91,6 +91,29 @@ function reinstantiateTiles() {
...
+					// c.material = new THREE.MeshBasicMaterial({map: c?.material?.map});
+
+				}
+
+			}
+
+		} );
+
+	};
+
```
gkjohnson, line 116: Do we need to do this if we've already added an ambient light to the scene? I'd prefer to not modify the model materials

jo-chemla: The MeshStandardMaterial of that new-york tileset has a `metalness` value of 1, and therefore appears full black even with white ambient light with power 1. A third option would be to enforce its metalness to 0. This seems less intrusive than replacing the tiles material with a MeshBasicMaterial, or than defining an emissiveMap on top of the PBR MeshStandardMaterial. Pushing a commit with this option.

gkjohnson: People generally wind up copying code from examples without fully understanding the context, at least for the core portion of the code like loading the tileset. So I'd prefer to not build in code that modifies the material of the tileset otherwise it will likely wind up in others projects.

Constructing a 2x2 env map or something would probably be the simplest solution.

Also generally the goal of the ion example was to show how to load a tileset and others would make any other necessary changes to view or interact their tileset in their project. Are you using this to view / test tilesets from ion? It's a good improvement to have I'm just trying to understand the context for the example changes.

jo-chemla: Thanks for getting back. Indeed I was testing some Ion public tilesets with this repo example page, to try to understand why that one was black, before understanding it was a material problem in this tileset's case. And you're right it would be best for Cesium to simply have tilesets with correct materials in the first place. 

Just tried adding a white 2x2 envMap, the background appears correctly white but the tileset material remains black. 
```js
texture = new TextureLoader().load( 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAYAAABytg0kAAAAAXNSR0IArs4c6QAAABdJREFUGFdj/Pfv/38Ghv8MjP//g1kMAGafCfYnZokCAAAAAElFTkSuQmCC' );
texture.mapping = EquirectangularReflectionMapping;
scene.background = texture;
scene.environment = texture;
```

gkjohnson: It looks like three.js environment textures must be 64x64 at a minimum? I'm not sure why but it may be related to the PMREM generation step. I might be worth making an issue report if you'd like. But it should be easy enough to make a large enough texture on the fly:

```js
const env = new DataTexture( new Uint8Array( 64 * 64 * 4 ).fill( 255 ), 64, 64 );
env.mapping = EquirectangularReflectionMapping;
env.needsUpdate = true;
```

jo-chemla: Thanks for reporting this, adding a 64x64 DataTexture env map works indeed - pushing this if you want to merge!

example/ionExample.js
```diff
@@ -99,14 +102,28 @@ function init() {
 
 	scene = new Scene();
 
+	// Add scene light for MeshStandardMaterial to react properly if its metalness is set to 0
+	const light = new AmbientLight( 0xffffff, 1 );
+	scene.add( light );
+
```
gkjohnson, line 108: I'll remove it - but now that we have an env map this ambient like will no longer be needed

gkjohnson:
Thanks!

PR #483, opened 2024-02-16
GlobeControls limit Frustum with horizon for performance

AlaricBaraou:
Currently for simplicity GlobeControls camera far is far enough to include the entire earth resulting in unnecessary tiles being loaded.

With these changes we limit the frustum to the distance of the horizon based on our current latitude and elevation.

This seems to work well but some tiles are being loaded while they don't intersect the frustum.

I edited the googleMapsExample.js to visualize it.

I'm not sure where is the issue but I'm pretty sure it's either the frustum or the OOB.
This particular example might help identifying the issue, when the camera is zoomed in on the north pole we get false positive on the intersectsFrustum check. 
If we look at it from the side, it's considered like an intersection as it seems like the frustum far limit does intersect with the obbs.
wrongFrustumCheck
But if we look at it from above, we can see that the frustum doesn't.
wrongFrustumCheckFromAbove

Those false positive unload as soon as on the side view, the frustum move outside of those obb.

### DIFF
--- a/example/googleMapsExample.js
+++ b/example/googleMapsExample.js
@@ -9,14 +9,17 @@ import {
 	WebGLRenderer,
 	PerspectiveCamera,
 	MathUtils,
+	CameraHelper,
 } from 'three';
 import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader.js';
 import { DRACOLoader } from 'three/examples/jsm/loaders/DRACOLoader.js';
 import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
 import { estimateBytesUsed } from 'three/examples/jsm/utils/BufferGeometryUtils.js';
 import Stats from 'three/examples/jsm/libs/stats.module.js';
+import { FlyOrbitControls } from './src/controls/FlyOrbitControls.js';
 
-let camera, controls, scene, renderer, tiles;
+let camera, controls, scene, renderer, tiles, cameraHelper;
+let thirdPersonCamera, thirdPersonRenderer, thirdPersonControls;
 let statsContainer, stats;
 
 const apiKey = localStorage.getItem( 'googleApiKey' ) ?? 'put-your-api-key-here';
@@ -30,6 +33,8 @@ const params = {
 	'reload': reinstantiateTiles,
 	displayBoxBounds: false,
 	displayRegionBounds: false,
+
+	showThirdPerson: false,
 
 };
 
@@ -83,6 +88,31 @@ function init() {
 	camera.position.set( 4800000, 2570000, 14720000 );
 	camera.lookAt( 0, 0, 0 );
 
+	// Third person camera view
+	thirdPersonCamera = new PerspectiveCamera( 60, window.innerWidth / window.innerHeight, 1, 160000000 );
+	thirdPersonCamera.position.set( 4800000, 2570000, 14720000 );
+	thirdPersonCamera.lookAt( 0, 0, 0 );
+
+	thirdPersonRenderer = new WebGLRenderer( { antialias: true } );
+	thirdPersonRenderer.setPixelRatio( window.devicePixelRatio );
+	thirdPersonRenderer.setSize( window.innerWidth, window.innerHeight );
+	thirdPersonRenderer.setClearColor( 0x0f1416 );
+
+	document.body.appendChild( thirdPersonRenderer.domElement );
+	thirdPersonRenderer.domElement.style.position = 'fixed';
+	thirdPersonRenderer.domElement.style.left = '5px';
+	thirdPersonRenderer.domElement.style.bottom = '5px';
+	thirdPersonRenderer.domElement.tabIndex = 1;
+
+	thirdPersonControls = new FlyOrbitControls( thirdPersonCamera, thirdPersonRenderer.domElement );
+	thirdPersonControls.screenSpacePanning = false;
+	thirdPersonControls.minDistance = 1;
+	thirdPersonControls.maxDistance = 14720000000;
+
+	cameraHelper = new CameraHelper( camera );
+	cameraHelper.raycast = function () {};
+	scene.add( cameraHelper );
+
 	// controls
 	controls = new GlobeControls( scene, camera, renderer.domElement, null );
 
@@ -106,6 +136,7 @@ function init() {
 	const exampleOptions = gui.addFolder( 'Example Options' );
 	exampleOptions.add( params, 'enableCacheDisplay' );
 	exampleOptions.add( params, 'enableRendererStats' );
+	exampleOptions.add( params, 'showThirdPerson' );
 
 	statsContainer = document.createElement( 'div' );
 	document.getElementById( 'info' ).appendChild( statsContainer );
@@ -122,6 +153,10 @@ function init() {
 }
 
 function onWindowResize() {
+
+	thirdPersonCamera.aspect = window.innerWidth / window.innerHeight;
+	thirdPersonCamera.updateProjectionMatrix();
+	thirdPersonRenderer.setSize( Math.floor( window.innerWidth / 3 ), Math.floor( window.innerHeight / 3 ) );
 
 	camera.aspect = window.innerWidth / window.innerHeight;
 	renderer.setSize( window.innerWidth, window.innerHeight );
@@ -175,6 +210,8 @@ function animate() {
 
 	if ( ! tiles ) return;
 
+	cameraHelper.visible = false;
+
 	controls.update();
 
 	// update options
@@ -189,6 +226,17 @@ function animate() {
 
 	renderer.render( scene, camera );
 	stats.update();
+
+	// render third person view
+	thirdPersonRenderer.domElement.style.visibility = params.showThirdPerson ? 'visible' : 'hidden';
+	if ( params.showThirdPerson ) {
+
+		cameraHelper.update();
+		cameraHelper.visible = true;
+
+		thirdPersonRenderer.render( scene, thirdPersonCamera );
+
+	}
 
 	updateHtml();
 
--- a/example/googleMapsExample.js
+++ b/example/googleMapsExample.js
@@ -72,6 +72,13 @@ function reinstantiateTiles() {
 
 	controls.setTilesRenderer( tiles );
 
+	tiles.loadedModels = 0;
+	tiles.onLoadModel = () => {
+
+		tiles.loadedModels ++;
+
+	};
+
 }
 
 function init() {
@@ -181,7 +188,11 @@ function updateHash() {
 
 	res.lat *= MathUtils.RAD2DEG;
 	res.lon *= MathUtils.RAD2DEG;
-	window.history.replaceState( undefined, undefined, `#${ res.lat.toFixed( 4 ) },${ res.lon.toFixed( 4 ) }` );
+
+	//todo above camer position use a matrix to get the position, should getPositionElevation be adjusted to work the same ?
+	const elevation = WGS84_ELLIPSOID.getPositionElevation( camera.position );
+
+	window.history.replaceState( undefined, undefined, `#${ res.lat.toFixed( 4 ) },${ res.lon.toFixed( 4 ) },${ elevation.toFixed( 4 ) }` );
 
 }
 
@@ -189,17 +200,17 @@ function initFromHash() {
 
 	const hash = window.location.hash.replace( /^#/, '' );
 	const tokens = hash.split( /,/g ).map( t => parseFloat( t ) );
-	if ( tokens.length !== 2 || tokens.findIndex( t => Number.isNaN( t ) ) !== - 1 ) {
+	if ( tokens.length < 3 || tokens.findIndex( t => Number.isNaN( t ) ) !== - 1 ) {
 
 		return;
 
 	}
 
-	const [ lat, lon ] = tokens;
-	WGS84_ELLIPSOID.getCartographicToPosition( lat * MathUtils.DEG2RAD, lon * MathUtils.DEG2RAD, 0, camera.position );
+	const [ lat, lon, height ] = tokens;
+	WGS84_ELLIPSOID.getCartographicToPosition( lat * MathUtils.DEG2RAD, lon * MathUtils.DEG2RAD, height, camera.position );
 
 	tiles.group.updateMatrixWorld();
-	camera.position.applyMatrix4( tiles.group.matrixWorld ).multiplyScalar( 2 );
+	camera.position.applyMatrix4( tiles.group.matrixWorld );
 	camera.lookAt( 0, 0, 0 );
 
 }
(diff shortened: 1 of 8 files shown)


gkjohnson:
Awesome! I'll take a deeper look at this next week. It would be nice to have some before / after stats on the number of tiles loaded and number of tiles rendered after the changes you've made so far.

> I'm not sure where is the issue but I'm pretty sure it's either the frustum or the OOB.
> This particular example might help identifying the issue, when the camera is zoomed in on the north pole we get false positive on the intersectsFrustum check.

I took a look and the issue is in the OBB frustum intersect code. In fact three.js' AABB frustum intersect code that the OBB logic is based on isn't correct, either. It will mark bounds as in the frustum even when they're clearly outside of it since it's only checking if the box is completely on one side of any plane which misses the common case of a box that intersects multiple planes if it's large. Here's a demo:

image

https://jsfiddle.net/89rtuaqp/2/

[This post](https://iquilezles.org/articles/frustumcorrect/) discusses the problem and a correct application of separating axis theorem. I'll make a three.js issue. We can see what they say - we may have to make our own extended frustum class that stores enough data for performing a more correct OBB intersection test.

gkjohnson:
https://github.com/mrdoob/three.js/issues/27756 may take a bit to resolve so I think it's best for us to modify our current implementation to be more robust, but I think that can happen in another PR. This can be done by storing the frustum corners as points on the Frustum so they can be tested against the OBB planes as referenced in the inigo post.

AlaricBaraou:
Good timing, I was working on a custom Frustum but it's hard to test it with the tiles logic, I might try in a simpler environment.

I added a class AdvFrustum that calculate the points of the Frustum to use them for different operations as suggested in the article you linked. But I think there are still case that are not handled or it might be the logic of the nested tiles etc that I don't really grasp at the moment.

Also the latest commit take care of "Google Maps Example: Add ability to specify height in hash" #482

AlaricBaraou:
a simpler example to play with it 

https://jsfiddle.net/0sb3gden/30/

The extended Frustum works well with axis aligned box but not with OBB at the moment.

src/three/math/OBB.js
```diff
@@ -13,9 +14,10 @@ export class OBB {
 
 	update() {
 
-		const { points, inverseTransform, transform, box } = this;
+		const { points, inverseTransform, transform, box, orientedBox } = this;
 		inverseTransform.copy( transform ).invert();
 
+		orientedBox.copy( box ).applyMatrix4( transform );
```
gkjohnson, line 20: If you transform the box by the transform you just get an enlarged AABB. We'll want to store the planes of the OBB (or generate them on the fly) so we can use those to test the frustum points.

src/three/math/OBB.js
```diff
@@ -64,6 +66,13 @@ export class OBB {
 
 		}
 
+		const boxInFrustum = frustum.boxInFrustum( this.orientedBox );
+		if ( ! boxInFrustum ) {
+
+			return false;
+
+		}
+
```
gkjohnson, line 75: Just like we had to make a custom implementation for testing all the OBB points against the frustum planes above -we'll need to do the same thing for the frustum points against the OBB planes here. The "boxInFrustum" function shouldn't be needed in this case.

AlaricBaraou:
I was about to write that it's good now but apparently now we have some false negative.

Looking good here
frustumOk

But here it discard some tiles that are in obvious intersections. 
frustumNOk

I`ll try to reproduce that one in jsfiddle

src/three/math/Ellipsoid.js
```diff
@@ -200,4 +203,43 @@ export class Ellipsoid {
...
+
+	getPositionElevation( pos ) {
+
+		_pos.copy( pos );
+		this.worldToLocal( _pos );
+
+		this.getPositionToSurfacePoint( _pos, _vec3 );
+		const elevation = _vec3.distanceTo( _pos );
+
+		return elevation;
```
gkjohnson, line 241: There shouldn't be any notion of a transform here - the 3d point should be passed in the same frame that the ellipsoid is stored in. If we assume a 90 degree rotation here then if the rotation of the globe object is set to anything different then this will break.

The 90 globe rotation is set on the tile root group [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/5d156142485764e1c1d64d3d024f8e7a9efb6edb/example/googleMapsExample.js#L52). So calling this function should look like this:

```js
_invMatrix.copy( tilesRenderer.group.matrixWorld ).invert();
_tempVec.copy( camera.position ).applyMatrix4( _invMatrix );
const elevation = ellipsoid.getPositionElevation();
```

The same pattern is used in GlobeControls [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/5d156142485764e1c1d64d3d024f8e7a9efb6edb/src/three/controls/GlobeControls.js#L101-L103).

src/three/math/Ellipsoid.js
```diff
@@ -200,4 +203,43 @@ export class Ellipsoid {
...
+		const semiMajorAxis = this.radius.x;
+		const semiMinorAxis = this.radius.z;
+		const eSquared = 1 - ( semiMinorAxis ** 2 / semiMajorAxis ** 2 );
+		const phi = latitude * MathUtils.DEG2RAD;
+
+		const sinPhiSquared = Math.sin( phi ) ** 2;
+		const N = semiMajorAxis / Math.sqrt( 1 - eSquared * sinPhiSquared );
+		return N;
+
+	}
```
gkjohnson, line 225: Are these formulas from somewhere? Can we add more comments?

src/three/math/OBB.js
```diff
@@ -17,25 +24,42 @@ export class OBB {
...
+		// Left and right
+		this.planes[ 2 ].setFromNormalAndCoplanarPoint( normals[ 1 ], this.points[ 0 ] );
+		this.planes[ 3 ].setFromNormalAndCoplanarPoint( normals[ 1 ].negate(), this.points[ 2 ] );
+
+		// Top and bottom
+		this.planes[ 4 ].setFromNormalAndCoplanarPoint( normals[ 2 ], this.points[ 0 ] );
+		this.planes[ 5 ].setFromNormalAndCoplanarPoint( normals[ 2 ].negate(), this.points[ 1 ] );
+
+		this.planeNeedsUpdate = false;
```
gkjohnson, line 62: I think these functions have become a bit difficult to follow. Instead of pulling points out of the points array we can generate the planes in a more explicit method that's hopefully easier to validate:

```js
_worldMin.copy( this.box.min ).applyMatrix4( this.transform );
_worldMax.copy( this.box.max ).applyMatrix4( this.transform );

_norm.set( 0, 0, 1 ).transformDirection( this.transform );
plane[ 0 ]. setFromNormalAndCoplanarPoint( _norm, _worldMin );
plane[ 1 ]. setFromNormalAndCoplanarPoint( _norm, _worldMax ).negate();

// ...
```

Note that just like the frustum planes the "positive" side of the plane will need to face "inward".

src/three/math/OBB.js
```diff
@@ -17,25 +24,42 @@ export class OBB {
...
+		points[ 1 ].set( max.x, min.y, min.z ).applyMatrix4( transform ); // Front-bottom-right
+		points[ 2 ].set( max.x, max.y, min.z ).applyMatrix4( transform ); // Front-top-right
+		points[ 3 ].set( min.x, max.y, min.z ).applyMatrix4( transform ); // Front-top-left
 
-			for ( let y = - 1; y <= 1; y += 2 ) {
+		points[ 4 ].set( min.x, min.y, max.z ).applyMatrix4( transform ); // Back-bottom-left
+		points[ 5 ].set( max.x, min.y, max.z ).applyMatrix4( transform ); // Back-bottom-right
+		points[ 6 ].set( max.x, max.y, max.z ).applyMatrix4( transform ); // Back-top-right
+		points[ 7 ].set( min.x, max.y, max.z ).applyMatrix4( transform ); // Back-top-left
+		this.planeNeedsUpdate = true;
```
gkjohnson, line 36: I don't think there's any reason to defer this plane generation, right? We can just update them here after the points in the "update" function.

gkjohnson (commented):
It looks like we're getting some depth fighting issues - but it looks like this was already happening 🤔 

image

We can look at this in another PR but we might want to find a smarter way to set the "near" value.

src/three/math/OBB.js
AlaricBaraou: The idea was often / sometimes, the planes are not needed as intersectsFrustum can return early before reaching the planes check.
I didn't test for any tradeoff or if neglectable. 
I'll go with the idea you prefer.

src/three/math/OBB.js
AlaricBaraou: oh I missed this one, I'll use setFromNormalAndCoplanarPoint instead

src/three/math/Ellipsoid.js
AlaricBaraou: I added source for the formulas

src/three/math/AdvFrustum.js
```diff
@@ -0,0 +1,64 @@
...
+			plane3.normal.x, plane3.normal.y, plane3.normal.z
+		);
+
+		// Create the vector B using the constants of the planes
+		const B = _constants.set( - plane1.constant, - plane2.constant, - plane3.constant );
+
+		// Solve for X by applying the inverse matrix to B
+		const X = B.applyMatrix3( A.invert() );
+
+		return [ X.x, X.y, X.z ];
```
gkjohnson, line 58: Let's use the common three.js pattern of passing a `target` vector variable into this function so we can avoid allocating a new array.

src/three/math/AdvFrustum.js
```diff
@@ -0,0 +1,64 @@
+import { Frustum, Matrix3, Vector3 } from 'three';
+
+const _mat3 = new Matrix3();
+const _constants = new Vector3();
+
+class AdvFrustum extends Frustum {
```
gkjohnson, line 6: _Tiny nit_ if you're in here modifying stuff - I'd call this "ExtendedFrustum".

src/three/controls/GlobeControls.js
```diff
@@ -152,7 +153,19 @@ export class GlobeControls extends EnvironmentControls {
 		// update the projection matrix
 		const largestDistance = Math.max( ...ellipsoid.radius );
 		camera.near = Math.max( 1, distanceToCenter - largestDistance * 1.25 );
-		camera.far = distanceToCenter + largestDistance + 0.1;
+
+		// update the far plane to the horizon distance
+		ellipsoid.getPositionToCartographic( camera.position, _vec );
```
gkjohnson, line 158: This function call doesn't look right? The camera position value needs to be transformed into the local "tilesGroup" space, first. This function also should take an empty object since it assigns values to "lat" and "lon". Right now this is modifying a "Vector3" object.

example/googleMapsExample.js
```diff
@@ -190,6 +238,17 @@ function animate() {
...
+	thirdPersonRenderer.domElement.style.visibility = params.showThirdPerson ? 'visible' : 'hidden';
+	if ( params.showThirdPerson ) {
+
+		cameraHelper.update();
+		cameraHelper.visible = true;
+
+		thirdPersonRenderer.render( scene, thirdPersonCamera );
+
+	}
+
```
gkjohnson, line 251: Do we need to keep all of these debug rendering additions? Ideally this example would just show a simple approach to adding a new google maps globe without too much extra.

gkjohnson (commented):
Added some comments for some small things to change - and a few questions on the example and GlobeControls additions.

test/Ellipsoid.test.js
```diff
@@ -101,7 +101,7 @@ describe( 'Ellipsoid', () => {
 		const LOCAL_EPSILON = 1e-8;
 		for ( let i = 0; i < 100; i ++ ) {
 
-			v.random().normalize().multiplyScalar( WGS84_RADIUS * 2 );
+			v.random().normalize().multiplyScalar( WGS84_HEIGHT * 2 );
```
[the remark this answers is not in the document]
AlaricBaraou: I was making tests and forgot to replace with the correct variable. fixing it

src/three/math/OBB.js
```diff
@@ -17,25 +23,39 @@ export class OBB {
...
+		points[ 0 ].set( min.x, min.y, min.z ).applyMatrix4( transform ); // Front-bottom-left
+		points[ 1 ].set( max.x, min.y, min.z ).applyMatrix4( transform ); // Front-bottom-right
+		points[ 2 ].set( max.x, max.y, min.z ).applyMatrix4( transform ); // Front-top-right
+		points[ 3 ].set( min.x, max.y, min.z ).applyMatrix4( transform ); // Front-top-left
 
-			for ( let y = - 1; y <= 1; y += 2 ) {
+		points[ 4 ].set( min.x, min.y, max.z ).applyMatrix4( transform ); // Back-bottom-left
+		points[ 5 ].set( max.x, min.y, max.z ).applyMatrix4( transform ); // Back-bottom-right
+		points[ 6 ].set( max.x, max.y, max.z ).applyMatrix4( transform ); // Back-top-right
+		points[ 7 ].set( min.x, max.y, max.z ).applyMatrix4( transform ); // Back-top-left
```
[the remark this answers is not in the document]
AlaricBaraou: Since the update to updatePlanes yes, doing it now

src/three/math/AdvFrustum.js
```diff
@@ -0,0 +1,64 @@
...
+
+		// Create the vector B using the constants of the planes
+		const B = _constants.set( - plane1.constant, - plane2.constant, - plane3.constant );
+
+		// Solve for X by applying the inverse matrix to B
+		const X = B.applyMatrix3( A.invert() );
+
+		return [ X.x, X.y, X.z ];
+
+	}
```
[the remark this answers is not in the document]
AlaricBaraou: Make sense, I'll try to follow that one from now on, updating it

example/googleMapsExample.js
AlaricBaraou: no need indeed, it was just for me

src/three/controls/GlobeControls.js
AlaricBaraou: yes nice catch, I guess the difference / influence of that variable was minimal I don't notice any change, but yes it's absolutely the right way to do it.

AlaricBaraou:
I addressed all points, just a note about this issue https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/482 
It can't be init under a certain height, at least in the area I was testing it.
I added a comment to check later

gkjohnson:
> It can't be init under a certain height, at least in the area I was testing it.

Likely what's happening is that the coarse levels of detail on the model load and have a much higher height than the lower LoDs which pushes the camera up due to the "GlobeControls" update function. I think it's okay to say that this fixes #482.

gkjohnson:
A few stats - some very nice improvements!

**Top Down View**

| | Before | After | % improvement |
|---|---|---|---|
| visible | 104 | 55 | ~47% |
| loaded | 307 | 189 | ~38% |

image

**Horizon View**

| | Before | After | % improvement |
|---|---|---|---|
| visible | 332 | 186 | ~43% |
| loaded | 574 | 376 | ~34% |

image

gkjohnson:
One thing I did notice is that the globe controls far plane _will_ clip the horizon a bit. It's subtle but you can see it change a little. Likely due to the earth model not exactly matching the ellipsoid. We don't have to fix it immediately but may something to keep in mind for the future there are ideas on how to address it.

**Before**

image

**After**
image

PR #499, opened 2024-03-08
Fix memory leak when loading texture with ImageBitmap

Junior2Ran:
During rendering large-scale 3DTiles with high resolution textures, we identified a memory leak problem. When continuously loading new tiles, memory usage would constantly increase without being released. And the browser will crash after console reported the error `Couldn't load texture blob: http://xxx`
image

I found that it cause by GLTFLoader loaded textures with ImageBitmap and did not close or dispose it when tiles be disposed. https://github.com/mrdoob/three.js/blob/master/examples/jsm/loaders/GLTFLoader.js#L2547

After this fix, I have tested that the memory won't increase endlessly. It will reach the top when geometry and textures in scene maintain in a reasonable value.

related issue: https://github.com/mrdoob/three.js/issues/23953

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -765,6 +765,13 @@ export class TilesRenderer extends TilesRendererBase {
 			for ( let i = 0, l = textures.length; i < l; i ++ ) {
 
 				const texture = textures[ i ];
+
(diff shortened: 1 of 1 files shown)


gkjohnson:
Great, thank you!

PR #500, opened 2024-03-08
WIP: Add support for custom fetch functions

ShinNoNoir:
This PR adds support for replacing the current `fetch()` calls with a custom fetch function. This support is realized via a new `fetcher` property in the `LoaderBase` and `TilesRendererBase` classes. The ability to use a custom fetch function is useful in projects that already use e.g. Angular's `HttpClient` with `HttpInterceptors` for fetching resources.

### DIFF
--- a/src/base/I3DMLoaderBase.js
+++ b/src/base/I3DMLoaderBase.js
@@ -84,7 +84,7 @@ export class I3DMLoaderBase extends LoaderBase {
 		} else {
 
 			const externalUri = this.resolveExternalURL( arrayToString( bodyBytes ) );
-			promise = fetch( externalUri, this.fetchOptions )
+			promise = this.fetcher( externalUri, this.fetchOptions )
 				.then( res => {
 
 					if ( ! res.ok ) {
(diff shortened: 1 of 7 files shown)


gkjohnson:
Hello! Thank you for the contribution.

Can you explain a bit more about what's not possible right now and what kind of use cases this enables? How are three.js' loaders dealt with since they use fetch, as well?

ShinNoNoir:
You raise a valid point regarding three.js' loaders. I'm not yet too familiar with 3d-tiles-renderer's source code and how exactly it interacts with other parts of three.js, so I'll have to take another look at it all.

Regarding use cases, one use case can be found in one of the projects I work on, in which all fetch-related logic is consolidated in one place. For example, the fetch logic may add authorization headers to requests only for certain domains and omit them for other requests. The ability to provide our own fetch function to the 3d-tiles-renderer library would mean all existing fetch logic could be easily reused.

In any case, I'll mark this PR as WIP for now until further investigation.

gkjohnson:
In this project and native three.js loaders there are potentially further fetches that occur once file load has begun for textures, other file dependencies, etc. If there are use cases that are strictly not possible we can discuss API changes but otherwise I think expecting every library to change to provide an optional callback for every possible call to fetch so it's more convenient for your application is a bit too much. Setting the headers and other fetch options is possible through the `TilesRenderer.fetchOptions` field.

PR #501, opened 2024-03-11
Make pointer tracker handle different canvas positions and sizes

AlaricBaraou:
In the case of a canvas non aligned on the top-left corner of the viewport the previous pointer tracker would end up with wrong positions.

This PR fix this.

There might be room for minor adjustments by trying to avoid calling getBoundingClientRect each time but optimising it by caching it might cause bugs in app with unusual canvas resize / reposition events.

### DIFF
--- a/src/three/controls/PointerTracker.js
+++ b/src/three/controls/PointerTracker.js
@@ -2,8 +2,9 @@ import { Vector2 } from 'three';
 
 export class PointerTracker {
 
-	constructor() {
+	constructor( domElement ) {
 
+		this.domElement = domElement;
 		this.buttons = 0;
 		this.pointerType = null;
 		this.pointerOrder = [];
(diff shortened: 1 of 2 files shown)


src/three/controls/PointerTracker.js
```diff
@@ -75,10 +75,22 @@ export class PointerTracker {
...
 
+	// get the pointer position in the coordinate system of the target element
+	getAdjustedPointer( e ) {
+
+		const rect = e.target.getBoundingClientRect();
+		const x = e.clientX - rect.left;
+		const y = e.clientY - rect.top;
+		return { x, y };
+
+	}
```
gkjohnson, line 86: I think we should provide a dom element to adjust the client x and y to be relative to as a member of the tracker. There's no real guarantee that the target is identical or the intended surface for each event.

We should also pass in a target object so we don't create a new object on each call.

Issue #508, opened 2024-03-19
Considering supporting legacy b3dm headers?


Insopitus:
**Describe the bug**

My project uses some legacy old b3dm format (still version 1 though) which 3DTilesRendererJS failed to parse. It throws: `Uncaught RangeError: Invalid typed array length: -1179912019`. I did some research and found out it is because of the legacy header the b3dm models use.
The cesium repository has some [code](https://github.com/CesiumGS/cesium/blob/59e1f74c85167b8468ac93c8d5cf3241f8f506e2/packages/engine/Source/Scene/B3dmParser.js#L64) to handle it, which I didn't find [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/1ce1070f38dfa3dc0e87f7d93343bf66924dbeab/src/base/B3DMLoaderBase.js#L45) in 3DTilesRendererJS.

**To Reproduce**

1. download and unzip the [file](https://github.com/NASA-AMMOS/3DTilesRendererJS/files/14648866/testb3dm.zip) attached
2. parse it with `B3DMLoaderBase`

gkjohnson:
As far as I know these legacy formats are from pre release (version 0) variants of the 3d tiles specification which are not supported. Even if the legacy header did parse the gltf contents contained within the file are also a legacy 1.0 version which the three.js library cannot parse, anyway:

```js
"asset":{
  "generator":"infoearth",
  "premultipliedAlpha":true,
  "profile":{"api":"webgl","version":"1.0.2"},
  "version":1
}
```

You'll have to upgrade your files if you'd like to use them with this project.

Insopitus:
Would there be options to use custom loaders for parsing and decoding rather than the default ones? I have written a gltf 1.0 parser and b3dm parser that supports legacy headers. For now, I want to migrate to 3DTilesRendererJS but still need to support the legacy model files. I could modify the source code of 3DTilesRendererJS to meet the requirements. It would be better to directly use the library so it's easier to keep up with future updates though.

gkjohnson:
I'll have to think about this - maybe supporting the ability to get a loader constructor from `LoadingManager.getHandler`? But generally I don't love the idea of adding a feature just to enable support long-deprecated, pre-release file formats. Even the newer B3DM version and the other original 3d tiles geometry formats (PNTS, I3DM, CMPT) are deprecated in favor of glTF. Is there no Cesium tool or something that will migrate a format to a newer version of the format?

[...]

PR #517, opened 2024-03-25
Added `loaderMap` property on `TilesRenderer` class to enable using custom loaders for tile formats

Insopitus:
#508 

I didn't go the `LoadingManager.getHandler` solution, since `B3DMLoader` and other tile format loaders aren't inherited from three.js' `Loader` class. It would at least cause ts errors for the library users.

Instead, a property called `loaderMap` is added to the `TilesRenderer` class to define which loader should be used for each format.

```js
const tilesRenderer = new TilesRenderer( './path/to/tileset.json' );
tilesRenderer.loaderMap.set('b3dm', new YourCustomB3DMLoader(tilesRenderer.manager));
tilesRenderer.setCamera( camera );
tilesRenderer.setResolutionFromRenderer( camera, renderer );
scene.add( tilesRenderer.group );
```
@gkjohnson how do you like it?

### DIFF
--- a/src/three/TilesRenderer.d.ts
+++ b/src/three/TilesRenderer.d.ts
@@ -3,6 +3,7 @@ import { Tile } from '../base/Tile';
 import { Tileset } from '../base/Tileset';
 import { TilesRendererBase } from '../base/TilesRendererBase';
 import { TilesGroup } from './TilesGroup';
+import { LoaderBase } from '../base/LoaderBase';
 
 export class TilesRenderer extends TilesRendererBase {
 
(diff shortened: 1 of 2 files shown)


Insopitus:
I think I forgot the loaders used in `CMPTLoader`. I'll update it tomorrow if think the `loaderMap` thing is acceptable.

gkjohnson:
> Instead, a property called loaderMap is added to the TilesRenderer class to define which loader should be used for each format.

B3DM, I3DM, PNTS, and CMPT are all deprecated now and glTF is the only officially supported 3d tiles geometry format - so ultimately this is an addition that's only for your usecase. I'd prefer not to modify the public facing API for this change.

Insopitus:
Okay, understood.

Issue #438, opened 2024-01-02
GlobeControls: Further Improvements


gkjohnson:
**High Priority**
- Retain a height when close to the surface? Ie snap back down when possible? The initial tiles are low res which pushes the tiles upward
- Normalize movement events using device pixel ratio and pixel distance rather than screen-space distance
- Enable drag with zoom when pinched
- Only perform zoom, rotation, pivot raycasts when the camera or tiles have changed

**Other** 
- Auto orient the camera to a reasonable position based on initial position
- Add events for knowing when control beings and ends
- Slow to a stop on zoom out to globe
- Trend camera towards rotation that guarantees camera visibility when updating the orientation
- There's an order of operations issue between tile update and controls update. You need to update the camera to update the tiles - but you also want to update the camera position based on the visible tiles.
- Fall back to intersection with WGS84 ellipse / plane for intersection
- Shift + scroll could adjust altitude
- Fade pivot icon in and out

PR #531, opened 2024-04-26
EnvironmentControls updates

### LINKED ISSUE #530: EnvironmentControls: Get last known / used distance or point
Can be useful for determining the focus point especially when converting between ortho and perspective cameras

gkjohnson:
Related to #438 
Fix #530 

- Fix zoomed out rotation action scale
- Fix GlobeControls: Zooming out past threshold increases zoom speed

**TODO**
- Separate PR for simplification & polish, first
  - Simplify logic
  - Simplify use of zoom point where possible
  - Ensure actions scale based on pixel density
- Add function for retrieving the last interacted point
- The logic of the controls relies on positioning - see if we can fix that? Perhaps set the camera distance at a certain position based on zoom?

### DIFF
--- a/src/three/controls/EnvironmentControls.js
+++ b/src/three/controls/EnvironmentControls.js
@@ -378,6 +378,14 @@ export class EnvironmentControls extends EventDispatcher {
 
 			e.preventDefault();
 
+			const { pointerTracker } = this;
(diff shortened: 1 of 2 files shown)


gkjohnson:
Closing in favor of #537

PR #555, opened 2024-05-11
Remove Matrix4.identity(), fix typo

Nmzik:
1. Remove the call that resets the matrix to the identity matrix `Matrix4.identity()`, as the constructor initializes it by default.
2. Fix typo: `material` -> `materials`

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -484,10 +484,6 @@ export class TilesRenderer extends TilesRendererBase {
 
 			}
 
-		} else {
-
-			transform.identity();
-
 		}
 
 		if ( parentTile ) {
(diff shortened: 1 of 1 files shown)


src/three/TilesRenderer.js
```diff
@@ -566,7 +563,6 @@ export class TilesRenderer extends TilesRendererBase {
 				break;
 
 			case 'z':
-				upAdjustment.identity();
```
gkjohnson, line 569: Thanks for the clean up! Why don't we remove the `case 'z':` block in this case. I think leaving an empty case will look like a mistake in the future.

Nmzik: @gkjohnson Fixed

gkjohnson:
Thank you!

PR #564, opened 2024-05-22
Fix #561

### LINKED ISSUE #561: Bug: DOMException: signal is aborted without reason
I've encountered a bug related to external tilesets. It appears that the tile somehow was aborted before. However, I have no idea how that's possible, considering an AbortController instance is created per each load. Therefore, at this point, I need to check additionally if the signal was aborted before (`signal.aborted`). It seems to happen randomly (but only if you have externals tilesets defined in the main Tileset.json).

Screenshot 2024-05-20 143649

Screenshot 2024-05-20 144303

```
{
    "stack": "Error: signal is aborted without reason\n    at http://localhost:3000/node_modules/3d-tiles-renderer/src/base/TilesRendererBase.js?v=02784ebd:439:19\n    at LRUCache.unloadUnusedContent (http://localhost:3000/node_modules/3d-tiles-renderer/src/utilities/LRUCache.js?v=02784ebd:162:26)\n    at http://localhost:3000/node_modules/3d-tiles-renderer/src/utilities/LRUCache.js?v=02784ebd:180:10"
}
```

**Update:** I'm unable to check the status of the variable `signal.aborted`
```
if (t.__loadAbort.signal.aborted === false)
   t.__loadAbort.abort(); //Bypasses the if statement and throws an exception. What's going on?
```

Nmzik:
Fixes #561 promise not getting caught. @gkjohnson Thank you!

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -360,6 +360,10 @@ export class TilesRenderer extends TilesRendererBase {
 			} );
 
 
+		} ).catch( () => {
+
+			// error is logged internally
+
 		} );
 		return pr;
 


gkjohnson:
Looks perfect, thank you!

PR #576, opened 2024-06-07
fix: const reassignment error in resolveDefault function

imoken777:
## Description

In the file [src/three/gltf/classes/PropertySetAccessor.js](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/453db64ee26da92481ea90373204273c915aed64/src/three/gltf/classes/PropertySetAccessor.js), at line 194, in the `resolveDefault` function, the `for` loop variable `i` is incorrectly declared as `const`.

As it is currently written, it is not possible to reassign the `const` variable `i`, resulting in an error. Since `i` is incremented within the loop, it should be declared as `let` instead of `const`.


gkjohnson:
Great, thank you!

XieYongHong:
How long will the new version be released

PR #590, opened 2024-06-21
LRUCache: don't mark unused items

Nmzik:
Don't mark unused items as "used" in LRUCache. Frees space for more tiles.


gkjohnson:
Thanks for the contribution but unfortunately I don't think we can make this change. The intermediate parent tiles are all loaded so they can be swapped to when the camera moves. If the parent tiles are not loaded (eg unloaded, etc) and the children are being displayed then the parent tile will be loaded again and only tiles up to the loaded parent set will be displayed until parent tiles down to the children are loaded.

PR #591, opened 2024-06-22
Fix illegal re-assign of `const`

Kronos3:
Building `0.3.34` with ESBuild triggers some build failures:

```
> node_modules/3d-tiles-renderer/src/three/loaders/gltf/metadata/utilities/ClassPropertyHelpers.js:207:50: error: Cannot assign to "i" because it is a constant
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,6 +17,7 @@
 		"no-throw-literal": [
 			"error"
 		],
+		"no-const-assign": "error",
 		"prefer-const": [
 			"error",
 			{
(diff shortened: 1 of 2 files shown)


gkjohnson:
Thanks! If you're familiar with eslint do you mind adding a rule to catch this kind of issue, as well? I'm surprised this wasn't caught.

Kronos3:
I added the const checking rule but it might be good to use the full set of eslint recommended rules and blacklist noisy rules off of that

gkjohnson:
> it might be good to use the full set of eslint recommended rules and blacklist noisy rules off of that

Thanks - yeah I thought this was implicitly used by some of the other configuration in place. I'll set this up and then publish. Thanks again!

gkjohnson:
Published in v0.3.35

PR #596, opened 2024-07-02
Load cesium ion tileset right when available

jailln:
Load cesium ion tilesets directly after getting the url from the api instead of waiting for the next animation frame.

This is particularly useful in cases where render is triggered asynchronously (and not at each frame) in which case we ask for `TilesRenderer` `update` at init and then when `load-tile-set` and `load-model` are triggered.

### DIFF
--- a/src/three/renderers/CesiumIonTilesRenderer.js
+++ b/src/three/renderers/CesiumIonTilesRenderer.js
@@ -60,6 +60,8 @@ const CesiumIonTilesRendererMixin = base => class extends base {
 				this.rootURL = url;
 				this.fetchOptions.headers = this.fetchOptions.headers || {};
 				this.fetchOptions.headers.Authorization = `Bearer ${ json.accessToken }`;
+				// Actually load the tileset now that we got its url from the cesium ion server
+				super.update();
 
 			} ).catch( () => {
 


gkjohnson:
Is there a reason you can't just call update immediately after instantiating the tiles renderer?

```js
tiles = new CesiumIonTilesRenderer( ... );
tiles.update();
```

jailln:
It's already what I'm doing but then I'm expecting for the tileset to load and wait for `load-tile-set` event to execute another `tiles.update()`. However `load-tile-set` is never raised since `tiles.update()` triggers the [first call to the cesium ion API](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/renderers/CesiumIonTilesRenderer.js#L41) which [gets resolved with the tileset url](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/renderers/CesiumIonTilesRenderer.js#L55) but doesn't load the tileset until there is a new `tiles.update()`.

I figured that would make sence that the first call to `tiles.update()` loads the tileset like for other `TilesRenderer`.

gkjohnson:
Oh I see, now - thank you. I think this a fine change to make. Do you mind adding a comment describing why we need to call "update" again here?

jailln:
Yes of course, it's done

gkjohnson:
Great, thank you!

PR #601, opened 2024-07-05
Lazy initialization of the upAdjustment matrix (Three.js)

Nmzik:
1. Instead of creating a new Matrix4 per tile, reuse it.
2. Pass the upAdjustment matrix to each loader constructor, keeping a reference to it without instantiating and copying the matrix itself.

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -580,9 +580,11 @@ export class TilesRenderer extends TilesRendererBase {
 
 			case 'b3dm': {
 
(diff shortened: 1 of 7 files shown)


gkjohnson (commented):
Thanks! I'd prefer not to change the Loader constructor signatures, though, and think we should keep the new Matrix4 on the class. There's already a ton of memory allocation that's going to necessarily happen during during loading and parsing of model geometry and one more Matrix4 is going to be really minor.

gkjohnson:
Closing in favor of #610

PR #605, opened 2024-07-06
Make use of cached._loadIndex

Nmzik:
Upon further investigation, I noticed that the `loadIndex` initialized in the `tile.cached` object to zero is not being used. This commit renames `loadIndex` to `_loadIndex` and utilizes it to prevent further tile loading if it has been unloaded.
**PS**: I believe there was some confusion/oversight between `tile.__loadIndex` and `tile.cached.loadIndex` before. #600


gkjohnson:
Thanks!

PR #623, opened 2024-07-16
fix: incorrect distanceToPoint for OBB

### LINKED ISSUE #622: Incorrect distance calculation for OBB
### Describe the bug

The `TileBoundingVolume` module assumes that the `transform` matrix does not contain a scale component. This is incorrect since it is calculated from `tile.transform` and breaks both distance and SSE calculations of tiles using non-identity scaling.

### Code

The following code is no longer valid with non-identity scaling:
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/c076d8b1de7cfb06881393c6c7c2a472abd10d9e/src/three/math/TileBoundingVolume.js#L118-L120

I'll push a PR to fix this issue, but is there a reason why you're not using `three.js` `OBB`?

Desplandis:
### List of changes
- Fixes #622.
- Add `clampToPoint` and `distanceToPoint` methods in `OBB` in a similar fashion as `THREE.Box3` and `THREE.OBB`.

### DIFF
--- a/src/three/math/OBB.js
+++ b/src/three/math/OBB.js
@@ -16,6 +16,33 @@ export class OBB {
 
 	}
 
+	/**
+	 * Clamps the given point within the bounds of this OBB
+	 * @param {Vector3} point
+	 * @param {Vector3} result
+	 * @returns {Vector3}
+	 */
+	clampPoint( point, result ) {
+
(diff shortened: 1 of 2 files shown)


gkjohnson:
Excellent - thank you!

PR #625, opened 2024-07-18
OBB: Encapsulate ray functions into OBB class

gkjohnson:
In the interest of encapsulating some of the bounding volume logic like #623 - I've moved the ray intersection functions over onto the OBB class.

cc @Desplandis

### DIFF
--- a/src/three/math/OBB.js
+++ b/src/three/math/OBB.js
@@ -1,8 +1,9 @@
-import { Matrix4, Box3, Vector3, Plane } from 'three';
+import { Matrix4, Box3, Vector3, Plane, Ray } from 'three';
 
 const _worldMin = new Vector3();
 const _worldMax = new Vector3();
 const _norm = new Vector3();
+const _ray = new Ray();
 
 export class OBB {
 
@@ -40,6 +41,23 @@ export class OBB {
 	distanceToPoint( point ) {
 
 		return this.clampPoint( point, _norm ).distanceTo( point );
+
+	}
+
+	// returns boolean indicating whether the ray intersected the box
+	intersectsRay( ray ) {
+
+		_ray.copy( ray ).applyMatrix4( this.inverseTransform );
+		return _ray.intersectsBox( this.box );
+
+	}
+
+	// Sets "target" equal to the intersection point.
+	// Returns "null" if no intersection found.
+	intersectRay( ray, target ) {
+
+		_ray.copy( ray ).applyMatrix4( this.inverseTransform );
+		return _ray.intersectBox( this.box, target );
 
 	}
 
--- a/src/three/math/OBB.js
+++ b/src/three/math/OBB.js
@@ -44,7 +44,7 @@ export class OBB {
 
 	}
 
-	// returns boolean indicating whether the ray intersected the box
+	// returns boolean indicating whether the ray has intersected the obb
 	intersectsRay( ray ) {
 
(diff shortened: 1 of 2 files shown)


src/three/math/TileBoundingVolume.js
```diff
@@ -73,10 +66,9 @@ export class TileBoundingVolume {
 		if ( obb ) {
 
 			// the obb transform contains no scale
-			_ray.copy( ray ).applyMatrix4( obb.inverseTransform );
-			if ( _ray.intersectBox( obb.box, _obbVec ) ) {
+			if ( obb.intersectsRay( ray, _obbVec ) ) {
 
-				obbDistSq = obb.box.containsPoint( _ray.origin ) ? 0 : _ray.origin.distanceToSquared( _obbVec );
+				obbDistSq = obb.box.containsPoint( ray.origin ) ? 0 : ray.origin.distanceToSquared( _obbVec );
```
gkjohnson, line 71: This this point had the same issue as the function pointed out in #623 - the transform scale wasn't accounted for.

src/three/math/OBB.js
```diff
@@ -43,6 +44,32 @@ export class OBB {
 
 	}
 
+	// returns boolean indicating whether the ray has intersected the obb
+	intersectsRay( ray ) {
+
+		_ray.copy( ray ).applyMatrix4( this.inverseTransform );
+		return _ray.intersectsBox( this.box );
```
Desplandis, line 51: If you wish to follow `three` interface for `OBB#intersectsRay`, it should be of type `Ray -> boolean`. You could also remove code duplication.

```suggestion
		return _ray.intersectRay( this.box, _norm ) !== null;
```

src/three/math/TileBoundingVolume.js
```diff
@@ -73,10 +66,9 @@ export class TileBoundingVolume {
 		if ( obb ) {
 
 			// the obb transform contains no scale
-			_ray.copy( ray ).applyMatrix4( obb.inverseTransform );
-			if ( _ray.intersectBox( obb.box, _obbVec ) ) {
+			if ( obb.intersectsRay( ray, _obbVec ) ) {
```
Desplandis, line 69: Shall be `intersectRay`, `intersectsRay` only have one parameter and  thus `_obbVec` components will never be assigned.

```suggestion
			if ( obb.intersectRay( ray, _obbVec ) ) {
```

src/three/math/TileBoundingVolume.js
Desplandis: Thanks for fixing this issue too. =)

Desplandis (commented):
Looks good except a single typo, thanks for fixing the issue with the projection. =)
I will try tomorrow with some of my "scaled" tilesets!

src/three/math/TileBoundingVolume.js
gkjohnson: Oops, thank you!

src/three/math/OBB.js
gkjohnson: > it should be of type Ray -> boolean

What do you mean by this? `Ray.intersectsBox` already [returns a boolean value](https://threejs.org/docs/?q=ray#api/en/math/Ray.intersectsBox)

Desplandis: My bad, I read `intersectBox`.

PR #627, opened 2024-07-19
[Batch table] `getDataForId` and `3DTILES_batch_table_hierarchy` support

### LINKED ISSUE #595: [Batch table] implement a method to get data for a batch `ID`

Hi, thanks for this awesome library ! I'm currently working on using it to render 3d tiles data in [iTowns](https://github.com/iTowns/itowns) which is a geographic data visualization framework. I am more specifically working on batch tables at the moment.

In our previous implementation, we had a method that read all data in a batch table for a given batch `ID`, whether those data are stored directly in the batch table or using [`3DTILES_batch_table_hierarchy`](https://github.com/CesiumGS/3d-tiles/blob/main/extensions/3DTILES_batch_table_hierarchy/README.md) extension.

If you would be willing, I wish to implement such a method in your library.

The [`BatchTable`](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/70ae645ebb1b76b6dd1153818af20f8199d7810f/src/utilities/FeatureTable.js#L153) you already implemented provides properties and methods which allow users to implement such a method on their side. However, I considered two elements on that idea. 

- The first is that such a method could perhaps be useful to other users than us, though I admit I didn't investigate on this matter.
- Moreover, reading all entries in a binary batch table for a given batch `ID` with your [`BatchTable.getData`](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/70ae645ebb1b76b6dd1153818af20f8199d7810f/src/utilities/FeatureTable.js#L162) method requires parsing the whole binary batch table when we could just parse the buffer part corresponding to the batch `ID`. I think this is an optimization issue, especially when dealing with a large number of batches.

I propose to implement a `getDataForBatchId` method in [`BatchTable`](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/70ae645ebb1b76b6dd1153818af20f8199d7810f/src/utilities/FeatureTable.js#L153) (the name is just a suggestion) which, for a given `batchID` parameter, would return an object as such : 

```js
{
    batchTableProperty1: value1,
    batchTableProperty2: value2,
    // ...All other properties in the batch table
    extensions: {
        3DTILES_batch_table_hierarchy: {
            className1: { ...valuesForClass1Properties },
            className2: { ...valuesForClass2Properties },
            // ...All other classes in the extension of the batch table
        },
    },
}
```

Would the idea of implementing such a method suit you ?

mgermerie:
This PR adds : 

- A `BatchTable.getDataForId` method which returns all data in a batch table for a given batch `ID`.
- The support for the [`3DTILES_batch_table_hierarchy`](https://github.com/CesiumGS/3d-tiles/blob/main/extensions/3DTILES_batch_table_hierarchy/README.md) extension to enrich `BatchTable.getDataForId` with data eventually stored using this extension.

This PR should close #595.

**Here is an explanation on a few things I did :**

I added a [`BatchTableHierarchyExtension`](https://github.com/mgermerie/3DTilesRendererJS/blob/5660f778cafe7cf2fb14373f0371698f2c0ed960/src/utilities/BatchTableHierarchyExtension.js) class that allows parsing batch table data formatted using [`3DTILES_batch_table_hierarchy`](https://github.com/CesiumGS/3d-tiles/blob/main/extensions/3DTILES_batch_table_hierarchy/README.md) extension.

I defined a `parseBinArray` function in [`FeatureTable.js`](https://github.com/mgermerie/3DTilesRendererJS/blob/5660f778cafe7cf2fb14373f0371698f2c0ed960/src/utilities/FeatureTable.js#L3) file. It is simply a factorization of already existing code that allows parsing data from a binary batch table in [`BatchTableHierarchyExtension`](https://github.com/mgermerie/3DTilesRendererJS/blob/5660f778cafe7cf2fb14373f0371698f2c0ed960/src/utilities/BatchTableHierarchyExtension.js) without needing to rewrite this function.

I defined `BatchTable` in its own [`BatchTable.js`](https://github.com/mgermerie/3DTilesRendererJS/blob/5660f778cafe7cf2fb14373f0371698f2c0ed960/src/utilities/BatchTable.js) file, outside of [`FeatureTable.js`](https://github.com/mgermerie/3DTilesRendererJS/blob/5660f778cafe7cf2fb14373f0371698f2c0ed960/src/utilities/FeatureTable.js) file. This is just to prevent circular dependency. 
Otherwise, we would have the following :
```js
// In BatchTableHierarchyExtension.js file :
import { parseBinArray } from './FeatureTable.js';
```
```js
// In FeatureTable.js file :
import { BatchTableHierarchyExtension } from './BatchTableHierarchyExtension.js';
```

**Notes and questions :**

- I never wrote type definition files before so don't hesitate to mention if I'm missing some points in those I edited.
- If I understood well, everything that is defined within a type definition file describes what is in the public API (plus the exports in `index.js`). Should I add a `BatchTableHierarchyExtension.d.ts` file or would you rather have it not exposed ?
- I chose to register the `BatchTableHierarchyExtension` within `BatchTable` constructor, but wonder of different ways to do it : would it be better to register it within [`B3DMLoaderBase`](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/ac83abe6f82974ae2f2d4d146a99f33b406bd414/src/base/loaders/B3DMLoaderBase.js#L70), [`I3DMLoaderBase`](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/ac83abe6f82974ae2f2d4d146a99f33b406bd414/src/base/loaders/I3DMLoaderBase.js#L73) and [`PNTSLoaderBase`](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/ac83abe6f82974ae2f2d4d146a99f33b406bd414/src/base/loaders/PNTSLoaderBase.js#L69) (a bit like `3DTILES_draco_point_compression` extension is registered in [`PNTSLoader`](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/ac83abe6f82974ae2f2d4d146a99f33b406bd414/src/three/loaders/PNTSLoader.js#L39)) ?

Don't hesitate if you need more info on the PR !

### DIFF
--- a/src/utilities/BatchTable.d.ts
+++ b/src/utilities/BatchTable.d.ts
@@ -10,10 +10,9 @@ export class BatchTable {
 
 	getKeys() : Array< String >;
 
-	getData(
-		key : String,
-		componentType : String | null,
-		type : String | null
-	) : Number | String | ArrayBufferView;
+	getDataFromId(
+		id: Number,
+		target?: Object
+	) : Object;
 
 }
--- a/src/utilities/BatchTable.d.ts
+++ b/src/utilities/BatchTable.d.ts
@@ -1,0 +1,19 @@
+export class BatchTable {
+
+	constructor(
+		buffer : ArrayBuffer,
+		batchSize : Number,
+		start : Number,
+		headerLength : Number,
+		binLength : Number
+	);
+
+	getKeys() : Array< String >;
+
+	getData(
+		key : String,
+		componentType : String | null,
+		type : String | null
+	) : Number | String | ArrayBufferView;
+
+}
--- a/src/utilities/BatchTable.js
+++ b/src/utilities/BatchTable.js
@@ -11,7 +11,30 @@ export class BatchTable extends FeatureTable {
 
 	getData( key, componentType = null, type = null ) {
 
+		console.warn( 'BatchTable: BatchTable.getData is deprecated. Use BatchTable.getDataFromId instead.' );
 		return super.getData( key, this.batchSize, componentType, type );
+
+	}
+
+	getDataFromId( id, target = {} ) {
+
+		if ( id < 0 || id >= this.batchSize ) {
+
+			throw new Error( `BatchTable: id value "${ id }" out of bounds for "${ this.batchSize }" features number.` );
+
+		}
+
+		for ( const key of this.getKeys() ) {
+
+			if ( key !== 'extensions' ) {
+
+				target[ key ] = super.getData( key, this.batchSize )[ id ];
+
+			}
+
+		}
+
+		return target;
 
 	}
 
--- a/src/utilities/BatchTable.js
+++ b/src/utilities/BatchTable.js
@@ -1,0 +1,18 @@
+import { FeatureTable } from './FeatureTable.js';
+
+export class BatchTable extends FeatureTable {
+
+	constructor( buffer, batchSize, start, headerLength, binLength ) {
+
+		super( buffer, start, headerLength, binLength );
+		this.batchSize = batchSize;
+
+	}
+
+	getData( key, componentType = null, type = null ) {
+
+		return super.getData( key, this.batchSize, componentType, type );
+
+	}
+
+}
--- a/src/utilities/BatchTable.js
+++ b/src/utilities/BatchTable.js
@@ -1,3 +1,4 @@
+import { BatchTableHierarchyExtension } from './BatchTableHierarchyExtension.js';
 import { FeatureTable } from './FeatureTable.js';
 
 export class BatchTable extends FeatureTable {
@@ -6,6 +7,18 @@ export class BatchTable extends FeatureTable {
 
 		super( buffer, start, headerLength, binLength );
 		this.batchSize = batchSize;
+
+		this.extensions = {};
+		const extensions = this.header.extensions;
+		if ( extensions ) {
+
+			if ( extensions[ '3DTILES_batch_table_hierarchy' ] ) {
+
+				this.extensions[ '3DTILES_batch_table_hierarchy' ] = new BatchTableHierarchyExtension( this );
+
+			}
+
+		}
 
 	}
 
@@ -29,6 +42,19 @@ export class BatchTable extends FeatureTable {
 			if ( key !== 'extensions' ) {
 
 				target[ key ] = super.getData( key, this.batchSize )[ id ];
+
+			}
+
+		}
+
+		for ( const extensionName in this.extensions ) {
+
+			const extension = this.extensions[ extensionName ];
+
+			if ( extension.getDataFromId instanceof Function ) {
+
+				target[ extensionName ] = target[ extensionName ] || {};
+				extension.getDataFromId( id, target[ extensionName ] );
 
 			}
 
--- a/src/utilities/BatchTableHierarchyExtension.js
+++ b/src/utilities/BatchTableHierarchyExtension.js
@@ -1,0 +1,127 @@
+import { parseBinArray } from './FeatureTable.js';
+
+export class BatchTableHierarchyExtension {
+
+	constructor( batchTable ) {
+
+		this.batchTable = batchTable;
+
+		const extensionHeader = batchTable.header.extensions[ '3DTILES_batch_table_hierarchy' ];
+
+		this.classes = extensionHeader.classes;
+		for ( const classDef of this.classes ) {
+
+			const instances = classDef.instances;
+			for ( const property in instances ) {
+
+				classDef.instances[ property ] = this._parseProperty( instances[ property ], classDef.length, property );
+
+			}
+
+		}
+
+		this.instancesLength = extensionHeader.instancesLength;
+
+		this.classIds = this._parseProperty( extensionHeader.classIds, this.instancesLength, 'classIds' );
+
+		if ( extensionHeader.parentCounts ) {
+
+			this.parentCounts = this._parseProperty( extensionHeader.parentCounts, this.instancesLength, 'parentCounts' );
+
+		} else {
+
+			this.parentCounts = new Array( this.instancesLength ).fill( 1 );
+
+		}
+
+		if ( extensionHeader.parentIds ) {
+
+			const parentIdsLength = this.parentCounts.reduce( ( a, b ) => a + b, 0 );
+			this.parentIds = this._parseProperty( extensionHeader.parentIds, parentIdsLength, 'parentIds' );
+
+		} else {
+
+			this.parentIds = null;
+
+		}
+
+		this.instancesIds = [];
+		const classCounter = {};
+		for ( const classId of this.classIds ) {
+
+			classCounter[ classId ] = classCounter[ classId ] ?? 0;
+			this.instancesIds.push( classCounter[ classId ] );
+			classCounter[ classId ] ++;
+
+		}
+
+	}
+
+	_parseProperty( property, propertyLength, propertyName ) {
+
+		if ( Array.isArray( property ) ) {
+
+			return property;
+
+		} else {
+
+			const { buffer, binOffset } = this.batchTable;
+
+			const byteOffset = property.byteOffset;
+			const componentType = property.componentType || 'UNSIGNED_SHORT';
+
+			const arrayStart = binOffset + byteOffset;
+
+			return parseBinArray( buffer, arrayStart, propertyLength, 'SCALAR', componentType, propertyName );
+
+		}
+
+	}
+
+	getDataFromId( id, target = {} ) {
+
+		// Get properties inherited from parents
+
+		const parentCount = this.parentCounts[ id ];
+
+		if ( this.parentIds && parentCount > 0 ) {
+
+			let parentIdsOffset = 0;
+			for ( let i = 0; i < id; i ++ ) {
+
+				parentIdsOffset += this.parentCounts[ i ];
+
+			}
+
+			for ( let i = 0; i < parentCount; i ++ ) {
+
+				const parentId = this.parentIds[ parentIdsOffset + i ];
+				if ( parentId !== id ) {
+
+					this.getDataFromId( parentId, target );
+
+				}
+
+			}
+
+		}
+
+		// Get properties proper to this instance
+
+		const classId = this.classIds[ id ];
+		const instances = this.classes[ classId ].instances;
+		const className = this.classes[ classId ].name;
+		const instanceId = this.instancesIds[ id ];
+
+		for ( const key in instances ) {
+
+			target[ className ] = target[ className ] || {};
+			target[ className ][ key ] = instances[ key ][ instanceId ];
+
+		}
+
+		return target;
+
+	}
+
+}
--- a/src/utilities/FeatureTable.js
+++ b/src/utilities/FeatureTable.js
@@ -149,20 +149,3 @@ export class FeatureTable {
 	}
 
 }
-
-export class BatchTable extends FeatureTable {
-
-	constructor( buffer, batchSize, start, headerLength, binLength ) {
-
-		super( buffer, start, headerLength, binLength );
-		this.batchSize = batchSize;
-
-	}
-
-	getData( key, componentType = null, type = null ) {
-
-		return super.getData( key, this.batchSize, componentType, type );
-
-	}
-
-}
--- a/src/utilities/FeatureTable.js
+++ b/src/utilities/FeatureTable.js
@@ -1,5 +1,77 @@
 import { arrayToString } from './arrayToString.js';
 
+export function parseBinArray( buffer, arrayStart, count, type, componentType, propertyName ) {
+
+	let stride;
+	switch ( type ) {
+
+		case 'SCALAR':
+			stride = 1;
+			break;
+
+		case 'VEC2':
+			stride = 2;
+			break;
+
+		case 'VEC3':
+			stride = 3;
+			break;
+
+		case 'VEC4':
+			stride = 4;
+			break;
+
+		default:
+			throw new Error( `FeatureTable : Feature type not provided for "${ propertyName }".` );
+
+	}
+
+	let data;
+	const arrayLength = count * stride;
+
+	switch ( componentType ) {
+
+		case 'BYTE':
+			data = new Int8Array( buffer, arrayStart, arrayLength );
+			break;
+
+		case 'UNSIGNED_BYTE':
+			data = new Uint8Array( buffer, arrayStart, arrayLength );
+			break;
+
+		case 'SHORT':
+			data = new Int16Array( buffer, arrayStart, arrayLength );
+			break;
+
+		case 'UNSIGNED_SHORT':
+			data = new Uint16Array( buffer, arrayStart, arrayLength );
+			break;
+
+		case 'INT':
+			data = new Int32Array( buffer, arrayStart, arrayLength );
+			break;
+
+		case 'UNSIGNED_INT':
+			data = new Uint32Array( buffer, arrayStart, arrayLength );
+			break;
+
+		case 'FLOAT':
+			data = new Float32Array( buffer, arrayStart, arrayLength );
+			break;
+
+		case 'DOUBLE':
+			data = new Float64Array( buffer, arrayStart, arrayLength );
+			break;
+
+		default:
+			throw new Error( `FeatureTable : Feature component type not provided for "${ propertyName }".` );
+
+	}
+
+	return data;
+
+}
+
 export class FeatureTable {
 
 	constructor( buffer, start, headerLength, binLength ) {
@@ -61,74 +133,10 @@ export class FeatureTable {
 
 			}
 
-			let stride;
-			switch ( featureType ) {
-
-				case 'SCALAR':
-					stride = 1;
-					break;
-
-				case 'VEC2':
-					stride = 2;
-					break;
-
-				case 'VEC3':
-					stride = 3;
-					break;
-
-				case 'VEC4':
-					stride = 4;
-					break;
-
-				default:
-					throw new Error( `FeatureTable : Feature type not provided for "${ key }".` );
-
-			}
-
-			let data;
 			const arrayStart = binOffset + byteOffset;
-			const arrayLength = count * stride;
+			const data = parseBinArray( buffer, arrayStart, count, featureType, featureComponentType, key );
 
-			switch ( featureComponentType ) {
-
-				case 'BYTE':
-					data = new Int8Array( buffer, arrayStart, arrayLength );
-					break;
-
-				case 'UNSIGNED_BYTE':
-					data = new Uint8Array( buffer, arrayStart, arrayLength );
-					break;
-
-				case 'SHORT':
-					data = new Int16Array( buffer, arrayStart, arrayLength );
-					break;
-
-				case 'UNSIGNED_SHORT':
-					data = new Uint16Array( buffer, arrayStart, arrayLength );
-					break;
-
-				case 'INT':
-					data = new Int32Array( buffer, arrayStart, arrayLength );
-					break;
-
-				case 'UNSIGNED_INT':
-					data = new Uint32Array( buffer, arrayStart, arrayLength );
-					break;
-
-				case 'FLOAT':
-					data = new Float32Array( buffer, arrayStart, arrayLength );
-					break;
-
-				case 'DOUBLE':
-					data = new Float64Array( buffer, arrayStart, arrayLength );
-					break;
-
-				default:
-					throw new Error( `FeatureTable : Feature component type not provided for "${ key }".` );
-
-			}
-
-			const dataEnd = arrayStart + arrayLength * data.BYTES_PER_ELEMENT;
+			const dataEnd = arrayStart + data.byteLength;
 			if ( dataEnd > binOffset + binLength ) {
 
 				throw new Error( 'FeatureTable: Feature data read outside binary body length.' );
--- a/README.md
+++ b/README.md
@@ -286,12 +286,8 @@ if ( intersects.length ) {
 		// Log the batch data
 		const batchTable = batchTableObject.batchTable;
 		const hoveredBatchid = batchidAttr.getX( face.a );
-		const batchData = batchTable.getData( 'BatchTableKey' );
-		if ( batchData ) {
-
-			console.log( batchData[ hoveredBatchid ] );
-
-		}
+		const batchData = batchTable.getDataFromId( hoveredBatchid );
+		console.log( batchData );
 
 	}
 
--- a/example/b3dmExample.js
+++ b/example/b3dmExample.js
@@ -186,12 +186,13 @@ function onMouseMove( e ) {
 			// Log the batch data
 			const batchTable = batchTableObject.batchTable;
 			hoveredBatchid = batchidAttr.getX( face.a );
+			const batchData = batchTable.getDataFromId( hoveredBatchid );
 
 			infoEl.innerText =
 				`_batchid   : ${ hoveredBatchid }\n` +
-				`Latitude   : ${ batchTable.getData( 'Latitude' )[ hoveredBatchid ].toFixed( 3 ) }\n` +
-				`Longitude  : ${ batchTable.getData( 'Longitude' )[ hoveredBatchid ].toFixed( 3 ) }\n` +
-				`Height     : ${ batchTable.getData( 'Height' )[ hoveredBatchid ].toFixed( 3 ) }\n`;
+				`Latitude   : ${ batchData[ 'Latitude' ].toFixed( 3 ) }\n` +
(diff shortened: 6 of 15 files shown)


src/utilities/BatchTable.js
```diff
@@ -0,0 +1,60 @@
...
+
+			}
+
+		}
+
+		for ( const extension in this.extensions ) {
+
+			target[ extension ] = this.extensions[ extension ].getDataForId( id );
+
+		}
```
gkjohnson, line 54: What's the rationale here? Is it the case that _every_ extension will have `getDataForId`?

src/utilities/BatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,199 @@
+import { parseBinArray } from './FeatureTable.js';
+
+function parseProperty( property, buffer, binOffset, length ) {
+
+	if ( ! ( property instanceof Object ) ) {
+
+		throw new Error( 'BatchTableHierarchyExtension: unexpected value found in extension properties.' );
```
gkjohnson, line 7: Lets remove this - I think we have to rely on the input data being well-formed otherwise we will have these kinds of checks everywhere.

gkjohnson:
> everything that is defined within a type definition file describes what is in the public API

Right - I've just been using the d.ts files to define the "public" API. So no need to add the hierarchical BatchedTable there.

> I chose to register the BatchTableHierarchyExtension within BatchTable constructor, but wonder of different ways to do it 

I guess the thing to consider here is whether these classes should _always_ be included in a build causing larger file sizes. Ideally this kind of thing would be a user-registerable plugin but I think this is okay for now. It might be worth just adding a "TODO" by the extension initialization noting that it's worth considering a plugin in the future.

But realistically the B3DM, I3DM, BatchTable, etc classes from 3D Tiles 1.0 are all deprecated so it may make sense to move all of that functionality to a 1.0-backwards compatible plugin later, anyway, to make the core module smaller.

src/utilities/BatchTable.d.ts
```diff
@@ -0,0 +1,24 @@
...
+	getData(
+		key : String,
+		componentType : String | null,
+		type : String | null
+	) : Number | String | ArrayBufferView;
+
+	getDataForId(
+		id: Number,
+		target?: Object,
+	) : Object;
```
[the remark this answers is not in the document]
mgermerie: No problem :) I'll rename it.

src/utilities/BatchTable.d.ts
```diff
@@ -0,0 +1,24 @@
...
+		binLength : Number
+	);
+
+	getKeys() : Array< String >;
+
+	getData(
+		key : String,
+		componentType : String | null,
+		type : String | null
+	) : Number | String | ArrayBufferView;
```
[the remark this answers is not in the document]
mgermerie: Agreed, batch tables work more on an ID basis (according to my understanding of specifications).

src/utilities/FeatureTable.js
```diff
@@ -1,5 +1,77 @@
...
+		case 'VEC3':
+			stride = 3;
+			break;
+
+		case 'VEC4':
+			stride = 4;
+			break;
+
+		default:
+			return 'typeError';
```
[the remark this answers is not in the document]
mgermerie: Agree with you, though we don't have access to the `key` the `parseBinArray` method was called to parse data from. This changes the message that was sent to users before this PR. Is it ok for you to throw an error with a less specific message (without specifying the `key`)?

src/utilities/BatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,199 @@
...
+			const classCounter = {};
+			for ( const classId of this.classIds ) {
+
+				classCounter[ classId ] = classCounter[ classId ] ?? 0;
+				this._cache.instancesIds.push( classCounter[ classId ] );
+				classCounter[ classId ] ++;
+
+			}
+
+		}
```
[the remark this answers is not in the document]
mgermerie: I opted for the lazy generation so that instancing a `BatchTableHierarchyExtension` does almost nothing but storing the raw data. It saves user who does not care for the batch table in his/her data the cost of parsing batch table hierarchy data, however small that cost. This is inspired by what's done in `FeatureTable` and `BatchTable`, where data is only parsed when `getData` method is called.

I admit it complicates `BatchTableHierarchyExtension` code, with the cache logic. If you would rather have data parsed in constructor, I can change it accordingly.

src/utilities/BatchTable.js
mgermerie: Whoops! Indeed adding a check for `getDataForId` existence is a bug prevention I completely missed. I'll add that.

I know of a research lab that uses custom made 3D tiles extensions, but I don't know if they use some on batch tables (perhaps @jailln knows more of it). Anyway, `BatchTable.extensions` would need to be public, with an extension registration tool for `BatchTable` to allow adding custom extensions. So I agree that we will probably only deal with batch table hierarchy extension.

src/utilities/BatchTableHierarchyExtension.js
mgermerie: Agreed, I'll remove that.

src/utilities/BatchTable.d.ts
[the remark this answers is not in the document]
gkjohnson: Great - do you mind adding a log warning the user that the function is deprecated if they use it, then? Like so:

```js
getData( ...args ) {

  console.warn( ... );
  return this._getData( ...args );

}
```

src/utilities/BatchTableHierarchyExtension.js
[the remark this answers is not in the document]
gkjohnson: I think we should opt for code simplicity to start and optimize later when needed. The code in `parseBinArray` is very lightweight and I don't think it should cause any issues.

src/utilities/FeatureTable.js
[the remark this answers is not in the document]
gkjohnson: I'm inclined to say we should just add a parameter to `parseBinArray` that takes the name of property so we can throw a nice human-understandable error. Anything else that comes to mind feels unnecessarily messy and redundant (validating types beforehand, catching and rethrowing an error).

src/utilities/BatchTable.d.ts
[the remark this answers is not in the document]
mgermerie: Yep I'll add that.

src/utilities/FeatureTable.js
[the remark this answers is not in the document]
mgermerie: Okay, I'll add the parameter.

src/utilities/BatchTableHierarchyExtension.js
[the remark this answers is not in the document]
mgermerie: Understood, I'll parse data in constructor.

src/utilities/BatchTable.d.ts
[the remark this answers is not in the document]
mgermerie: `BatchTable.getData` method being used only once and its code being a one liner, I did not define a `BatchTable._getData` method. Let me know if that suits you.

mgermerie:
I pushed specific commits for each modification you suggested to ease the review. I'll squash them before merging.

src/utilities/BatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,123 @@
...
+		// Get properties proper to this instance
+
+		const classId = this.classIds[ id ];
+		const instances = this.classes[ classId ].instances;
+		const className = this.classes[ classId ].name;
+		const instanceId = this.instancesIds[ id ];
+
+		for ( const key in instances ) {
+
+			if ( Object.prototype.hasOwnProperty.call( instances, key ) ) {
```
gkjohnson, line 110: From the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) - we should be using `Object.hasOwn`, instead.

src/utilities/BatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,123 @@
...
+				parentIdsOffset += this.parentCounts[ i ];
+
+			}
+
+			const parentIds = this.parentIds.slice(
+				parentIdsOffset,
+				parentIdsOffset + parentCount,
+			);
+
+			for ( const parentId of parentIds ) {
```
gkjohnson, line 89: Can we just use a normal for loop here instead of making a new array?

```js
for ( let i = 0; i < parentCount; i ++ ) {

  const parentId = parentIds[ parentIdsOffset + i ];
```

src/utilities/BatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,123 @@
...
+
+			const arrayStart = binOffset + byteOffset;
+
+			return parseBinArray( buffer, arrayStart, propertyLength, 'SCALAR', componentType );
+
+		}
+
+	}
+
+	getDataFromId( id ) {
```
gkjohnson, line 67: For consistency this function should take an optional `target = {}`, as well.

src/utilities/BatchTable.js
```diff
@@ -0,0 +1,67 @@
...
+
+		for ( const extensionName in this.extensions ) {
+
+			const extension = this.extensions[ extensionName ];
+
+			if ( extension.getDataFromId instanceof Function ) {
+
+				target[ extensionName ] = extension.getDataFromId( id );
```
gkjohnson, line 57: Is there a reason we can't just set the hierarch fields directly onto the target object? Is there risk of name overlap between the extension and the original BatchTable values?

src/utilities/BatchTableHierarchyExtension.js
gkjohnson: Also is there a reason `instances` _wouldn't_ have "key" at this point? Is this check needed?

src/utilities/BatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,123 @@
...
+		}
+
+		// Get properties proper to this instance
+
+		const classId = this.classIds[ id ];
+		const instances = this.classes[ classId ].instances;
+		const className = this.classes[ classId ].name;
+		const instanceId = this.instancesIds[ id ];
+
+		for ( const key in instances ) {
```
gkjohnson, line 108: From the `_parseProperty` function above - it looks like `instances` can be a value like a string or number (see the `! ( property instanceof Object )`) and therefore couldn't be iterated over - what happens here in that case?

src/utilities/BatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,199 @@
...
+
+		const arrayStart = binOffset + byteOffset;
+
+		return parseBinArray( buffer, arrayStart, length, 'SCALAR', componentType );
+
+	}
+
+}
+
+export class BatchTableHierarchyExtension {
```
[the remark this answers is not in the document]
gkjohnson: Thanks for the detailed explanation! I've left a few more comments now that I understand a bit more.

src/utilities/FeatureTable.js
```diff
@@ -61,74 +133,10 @@ export class FeatureTable {
...
-					break;
-
-				case 'FLOAT':
-					data = new Float32Array( buffer, arrayStart, arrayLength );
-					break;
-
-				case 'DOUBLE':
-					data = new Float64Array( buffer, arrayStart, arrayLength );
-					break;
+			const data = parseBinArray( buffer, arrayStart, count, featureType, featureComponentType );
```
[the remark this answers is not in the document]
mgermerie: Whoops, added it.

src/utilities/BatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,123 @@
...
+		} else {
+
+			const { buffer, binOffset } = this.batchTable;
+
+			const byteOffset = property.byteOffset;
+			const componentType = property.componentType || 'UNSIGNED_SHORT';
+
+			const arrayStart = binOffset + byteOffset;
+
+			return parseBinArray( buffer, arrayStart, propertyLength, 'SCALAR', componentType );
```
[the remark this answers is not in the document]
mgermerie: The only way an error that uses the property name can be thrown when `parseBinArray` is called from `BatchTableHierarchyExtension` is having a `componentType` in data that is neither `undefined` nor one of the types described [here](https://github.com/CesiumGS/3d-tiles/blob/87218a7c2030890b5200cf16cb466dbbd1480a20/extensions/3DTILES_batch_table_hierarchy/schema/3DTILES_batch_table_hierarchy.schema.json#L56-L86).

Currently in `BatchTableHierarchyExtension`, an `undefined` `componentType` defaults to `'UNSIGNED_SHORT'` and `type` is hard coded as `'SCALAR'` (following the specification). Therefore, I think it is quite unlikely that the error messages thrown in `parseBinArray` is ever thrown, except for some weirdly ill-formatted data.

Moreover `BatchTableHierarchyExtension._parseProperty` is parametrized with property values instead of name (which is not the case in `FeatureTable.getData`, the other place `parseBinArray` is called from). I made it that way so that any property can be parsed using `_parseProperty`, and not only properties that share the same parent object. We can thus parse `extensionHeader.classIds` with the same method we use to parse `extensionHeader.classes[ x ].instances[ someProperty ]`). We would therefore need to add a parameter to `BatchTableHierarchyExtension._parseProperty` just to pass the property name (which would be hard coded each time).

Therefore, I would suggest not to add the property name in this situation, though I'll add it if you still want it added :smile:

src/utilities/BatchTableHierarchyExtension.js
mgermerie: Indeed, it's something I added without reflection in the `for ... in` loop to prevent iterating through prototype and inherited properties but that is not needed here since the object is just a parsed JSON object. I'll remove it.

src/utilities/BatchTable.js
mgermerie: You named the reason : nothing in the extension specifications prevents having classes that share the same name with any property outside the extension data.

src/utilities/BatchTableHierarchyExtension.js
mgermerie: Indeed, it is simpler. I changed it.

src/utilities/BatchTableHierarchyExtension.js
mgermerie: It's an inconsistency in `_parseProperty` function. Actually, according to the spec, properties parsed using `_parseProperty` are `Array`, `Object` defining a location in binary part, or `undefined` but only for `parentCounts` and `parentIds` properties.
I changed `_parseProperty` and its usage a bit [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/627/commits/e49fc92aa5b28451e4bd22b7ece458755ce02c1a) to make it more relevant to the spec.

src/utilities/BatchTable.js
```diff
@@ -0,0 +1,68 @@
...
+
+	constructor( buffer, batchSize, start, headerLength, binLength ) {
+
+		super( buffer, start, headerLength, binLength );
+		this.batchSize = batchSize;
+
+		const extensions = this.header.extensions;
+		if ( extensions ) {
+
+			this.extensions = {};
```
gkjohnson, line 14: We should always set `this.extensions = {}` even if there are no extensions. The class surface should be consistent, I think.

src/utilities/BatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,122 @@
...
+
+		} else {
+
+			this.parentCounts = new Array( this.instancesLength ).fill( 1 );
+
+		}
+
+		if ( extensionHeader.parentIds ) {
+
+			this.parentIds = this._parseProperty( extensionHeader.parentIds );
```
gkjohnson, line 39: The `this._parseProperty` function takes a `propertyLength` but it's not passed in here. And if `parentIds` is always an array then there's no reason to pass it in to the function since the parseProperty doesn't do anything in that case.

It looks like [this isn't always an array, though](https://github.com/CesiumGS/3d-tiles/blob/87218a7c2030890b5200cf16cb466dbbd1480a20/extensions/3DTILES_batch_table_hierarchy/schema/3DTILES_batch_table_hierarchy.schema.json#L44-L46).

gkjohnson, line 39: Lets also make sure `parentIds` is set to a value regardless of whether it exists in the extension - even if that's an empty array or "null".

src/utilities/BatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,122 @@
...
+
+		this.classIds = this._parseProperty( extensionHeader.classIds, this.instancesLength );
+
+		if ( extensionHeader.parentCounts ) {
+
+			this.parentCounts = this._parseProperty( extensionHeader.parentCounts, this.instancesLength );
+
+		} else {
+
+			this.parentCounts = new Array( this.instancesLength ).fill( 1 );
```
gkjohnson, line 33: Why default this to 1? Shouldn't it be defaulted to 0 if there are no parents specified?

src/utilities/BatchTableHierarchyExtension.js
```diff
@@ -0,0 +1,122 @@
...
+			}
+
+			for ( let i = 0; i < parentCount; i ++ ) {
+
+				const parentId = this.parentIds[ parentIdsOffset + i ];
+				if ( parentId !== id ) {
+
+					Object.assign( target, this.getDataFromId( parentId ) );
+
+				}
```
gkjohnson, line 98: Is there a reason to no just pass target into `this.getDataFromId` here, as well? We're assigning all the same values onto the object.

src/utilities/BatchTableHierarchyExtension.js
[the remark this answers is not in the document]
gkjohnson: Got it - we should add something for an argument here or set the `propertyName` field to use some kind of reasonable default as a name in the function parameters, then. At the moment it looks like a bug, I feel.

gkjohnson (commented):
Just added a few more comments - thanks for bearing with me. I just want to make sure I understand the class and align the functions with the philosophies used in the rest of the project. I think it's just about there!

One last thing to do is update [this](https://github.com/NASA-AMMOS/3DTilesRendererJS?tab=readme-ov-file#read-batch-id-and-batch-table-data) section of the README to use `getDataFromId` instead of `getData`

src/utilities/BatchTable.js
mgermerie: You're right, I added it.

src/utilities/BatchTableHierarchyExtension.js
mgermerie: Yes indeed, I added the length of `parentIds` to `_parseProperty` parameters.

mgermerie: I copied `classIds` as a default values for `parentIds`, since an instance whose parent is itself is considered as without parent (as mentioned [here](https://github.com/CesiumGS/3d-tiles/blob/87218a7c2030890b5200cf16cb466dbbd1480a20/extensions/3DTILES_batch_table_hierarchy/README.md?plain=1#L194)).

I get that we want it done to keep consistency in class properties, but be aware that this could cause issue in a specific case. Indeed, if user has data with a `parentCounts` different from an array of `1` and has no defined `parentIds`, this will create unwanted inheritance between instances that should not inherit any property. I get that this is a weird situation (having a `parentCounts` without `parentIds`) but that's not something prohibited by the specification.

src/utilities/BatchTableHierarchyExtension.js
mgermerie: I agree it feels more logical but the specification defaults `parentCounts` to an array of ones ([here](https://github.com/CesiumGS/3d-tiles/blob/87218a7c2030890b5200cf16cb466dbbd1480a20/extensions/3DTILES_batch_table_hierarchy/README.md?plain=1#L190))

src/utilities/BatchTableHierarchyExtension.js
mgermerie: That's an oversight from me. I added it.

src/utilities/BatchTableHierarchyExtension.js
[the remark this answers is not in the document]
mgermerie: Okay, I'll add the property name as an argument then.

mgermerie:
> Just added a few more comments - thanks for bearing with me. I just want to make sure I understand the class and align the functions with the philosophies used in the rest of the project. I think it's just about there!
> 
> One last thing to do is update [this](https://github.com/NASA-AMMOS/3DTilesRendererJS?tab=readme-ov-file#read-batch-id-and-batch-table-data) section of the README to use `getDataFromId` instead of `getData`

No problem, thanks for the reviews ! :wink: 
I added the modifications you suggested and updated the README.

src/utilities/BatchTableHierarchyExtension.js
gkjohnson: I think it's okay to set `parentIds` to `null`, and that sounds like the better option given your explanation - I just wanted it to be set to _something_ in the general case so it's more clear what's going on when debugging.

src/utilities/BatchTable.js
```diff
@@ -0,0 +1,67 @@
...
+
+			}
+
+		}
+
+	}
+
+	getData( key, componentType = null, type = null ) {
+
+		console.warn( 'BatchTable: BatchTable.getData is deprecated. Use BatchTable.getDataFromId instead.' );
```
gkjohnson, line 27: Forgot about the b3dmExample that tests batch tables - [these lines](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/a232cb76591f88241abc5e67ba4b91a95bc7d549/example/b3dmExample.js#L191-L194) should be changed now that we've changed the interface.

gkjohnson (commented):
A couple final things but overall this looks good. Last thing - what's you're process for testing the changes so we can reference it in the future if changes need to be made?

src/utilities/BatchTableHierarchyExtension.js
mgermerie: Good idea setting `parentIds` to `null`. I changed it.

src/utilities/BatchTable.js
mgermerie: Well spotted ! I updated the example.

mgermerie:
> A couple final things but overall this looks good. Last thing - what's you're process for testing the changes so we can reference it in the future if changes need to be made?

The testing process is not ideal : I tested the changes I made by plugging my local 3DTilesRendererJS branch to a local branch of [iTowns](https://github.com/iTowns/itowns) which I'll push soon (can't access my computer at the moment 😄). The local iTowns branch is basically [this one](https://github.com/iTowns/itowns/tree/3d-tiles-migration) where I added a call to `BatchTable.getDataFromId` in [this example](https://github.com/iTowns/itowns/blob/3d-tiles-migration/examples/3dtiles_batch_table.html) and logged its output.

I think I could as well add an example inspired by your [batch table example](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/a232cb76591f88241abc5e67ba4b91a95bc7d549/example/b3dmExample.js), replacing the dataset by the one used in iTowns example. It would be simpler and would have the same value in terms of testing quality.

That's only a functional test that is not testing all functionalities (as I mentioned I did not manage to find a dataset with binary batch table hierarchy extension), and I did not write proper unit tests.

gkjohnson:
> I think I could as well add an example inspired by your [batch table example](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/a232cb76591f88241abc5e67ba4b91a95bc7d549/example/b3dmExample.js), replacing the dataset by the one used in iTowns example. It would be simpler and would have the same value in terms of testing quality.

I don't think we need to add a new test page - just modifying that example to use a sample file with a batched hierarchy example is probably good enough as long as the basic "BatchedMesh" and highlighting functionality still works.

> That's only a functional test that is not testing all functionalities (as I mentioned I did not manage to find a dataset with binary batch table hierarchy extension), and I did not write proper unit tests.

That's okay. Writing unit tests for these kinds of things is difficult, I think. And it's even more difficult to test in general when we don't have any example data. It's okay to wait until we have user complaints and examples to address any issues.

mgermerie:
I updated the example to use a dataset using both batch table using hierarchy extension and a "classic" batch table. Would you rather have me squash all commits or keep the tree as it is now ?

gkjohnson:
Excellent! Thanks for all your work on this. We can squash commits when merging the PR.

mgermerie:
Thanks for the reviews :smiley:

Issue #632, opened 2024-07-22
Ideation for removing stalls, improving time to tiles load


gkjohnson:
Related to #604 

Aimed primarily at improving the time-to-completion and removing load stalls for Googles Photorealistic tiles.

**Increase Download Queue Job Count**

- TODO

**Locally Cache Geometry Data** 

- TODO

**Improve "Skip" Traversal Options**

- TODO

**Data Preupload**

- TODO

PR #633, opened 2024-07-22
TilesRenderer: Queue loads for all visible tiles

gkjohnson:
Related to #632 

In progress PR for queuing up all tiles needed for rendering so we can make sure the jobs are saturated as consistently as possible. Currently tiles can seem slower but it's possible this is because the queue sort is prioritizing tiles that were not previously added to the queue, preventing some parents from loading sooner.

**TODO**
- Check if raising the queue size on tiles has a significant improvement before / after
- Measure time to load completion
- Ideate on queue sort improvements

### DIFF
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -495,7 +495,7 @@ export class TilesRendererBase {
 		const downloadQueue = this.downloadQueue;
 		const parseQueue = this.parseQueue;
 		const isExternalTileSet = tile.__externalTileSet;
-		lruCache.add( tile, t => {
+		const addedSuccessfully = lruCache.add( tile, t => {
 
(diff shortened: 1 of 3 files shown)


gkjohnson:
Generally this seems like an improvement for loading google tiles in the "aerial" example by a couple hundred ms when the parse and download queue max job count has been raised to 1000. Without the increased job cap there is little to no difference.

**Before**
~ 1969ms

**After**
~1725ms

We're potentially at risk of filling up the lru cache too quickly causing one branch of the tree to render and not others. Ideally we'd only push items onto the queue if it's an improvement over existing tiles.

gkjohnson:
Closing for now - the impact isn't significant in common cases to be worth it.

PR #634, opened 2024-07-23
Add schedulePerFrame option

gkjohnson:
Related to #632 

Adds an option for limiting the amount of jobs started per frame while allowing for more parallel jobs.

Timing before:
~2486 ms

Timing after with higher maxJobs = 1000:
~2400 ms

The improvement seems minimal if it's even consistent.


gkjohnson:
Improvement is too small only in extreme cases to be worth the change. Related to #633

Issue #608, opened 2024-07-11
Support of Implicit Tiling


AnthonyGlt:
[Implicit Tiling](https://docs.ogc.org/cs/22-025r4/22-025r4.html#toc31)

I wish to implement the support of Implicit Tiling (introduced in 3D Tiles 1.1). It has been discussed in different issues but is there already a plan or ideas how to do it ?

I've seen that in Cesium, the implicit tiling is directly managed from [Cesium3DTileset](https://github.com/CesiumGS/cesium/blob/7b93161da1cc03bdc796b204e7aa51fb7acebf04/packages/engine/Source/Scene/Cesium3DTileset.js#L2259), which is kind of the equivalent of `TilesRendererBase` here. Do you think I should do a similar implementation ?

I've also noticed the recent addition of plugins, will it be a better idea to implement a ImplicitTilingPlugin with the necessary hooks ?
What is your take on this topic ?

Do you have any tips/warnings to help me in this new implicit tiling support in 3DTilesRendererJS ?

I'll first begin by the parsing of the .subtree files.

Thank you

gkjohnson:
Hello! A contribution to support implicit tiling would be great - there hasn't been thought put into this feature yet, though, so your help here would be appreciated. As far as I understand the feature requires support for something akin to "virtual" tile children that have implicitly calculated bounding volumes and content URLs. Is that a reasonable summary?

> Do you think I should do a similar implementation ?

I'm less familiar with the Cesium3DTileset implementation for this. Do you mind giving a brief description of what a similar implementation would look like?

> Do you have any tips/warnings to help me in this new implicit tiling support in 3DTilesRendererJS ?

I think one of the larger things that will need to be considered here is how to handle these "virtual" child tiles since there's a lot of intermediate information stored on the tile object for the sake of traversing and loading data. I was originally thinking we'd want to avoid generating child tile representations for each implicit child but that may be an okay place to start since we'll ultimately have to create an object to store the intermediate data, anyway - whether it's stored as a child or somewhere else. An optimization here could be deleting all child objects to free up memory as soon as a tile is unloaded as the camera moves.

I'm curious as how Cesium is handling this already if you know, though.

> I've also noticed the recent addition of plugins, will it be a better idea to implement a ImplicitTilingPlugin with the necessary hooks ?
> What is your take on this topic ?

[...]

PR #650, opened 2024-07-25
feat(ImplicitTiling): add implicit tiling support

AnthonyGlt:
A partial implementation of [Implicit Tiling](https://docs.ogc.org/cs/22-025r4/22-025r4.html#toc31)

This MR is resulting from this [issue](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/608)

**NOTE: The implemenation is very similar to the one from CESIUM. I have even kept names & comments.**
I don't know if it is an issue ?

*************************************************************

- In the current code we don't have a Tile class. The tiles' structure is created in the `preprocessNode` of the `TilesRenderer` even though it doesn't really depends on it. To keep the current architecture, I had to expose  `preprocessNode`, so I made it static.

- The flow is the following:
     1. We read the tileset.json, if the key `implicitTiling` is presents, then I swap the content uri with the one in the `implicitTiling` object. The target uri will end by `.subtree`
     2. We load the `.subtree`, and from it we collect the bitstreams
     3. We create all the children from the subtree, we add the BB, geometricerror and the content uri if there is content. These children will be traversed and displayed as usual.
     4. For the last row of children, we check if they link to other subtrees and put the corresponding uri if necessary
     5. Once a new subtree is loaded, we first check if the root has content, if so, we create a children which is cloned from the parent, with the content uri. By doing this, the new content will be correctly loaded

- The parser returns nothing

*******************************************

**TODOs:**

- [ ] OCTREE support. Doesn't seem to difficult
- [ ] Sphere bounding volume support. Not to difficult. I only did the Region & Box
- [ ] [Subtree JSON Format](https://docs.ogc.org/cs/22-025r4/22-025r4.html#toc38) support. I only did the Binary format
- [ ] Metadata support
- [ ] External support (e.g The buffer could be stored in another file)
- [ ] Extension `3DTILES_implicit_tiling` support. For now the tileset must have the object implicitTiling as follows:
```
"implicitTiling": {
      "subdivisionScheme": "QUADTREE",
      "availableLevels": 21,
      "subtreeLevels": 7,
      "subtrees": {
        "uri": "subtrees/{level}/{x}/{y}.json"
      }
    }

```
- [ ] Multiple Contents support. Only one content is supported for now
- [ ] ImplicitTiling defined in children. Only the one in root is supported


AnthonyGlt:
I'll implement the support of the box to be able to test the data from
https://github.com/CesiumGS/3d-tiles-samples/

AnthonyGlt:
I've implemented the support of bounding box
you can now try with the sample from 
https://github.com/CesiumGS/3d-tiles-samples/tree/main/1.1/SparseImplicitQuadtree

Note: 
If nothing appears, try to zoom.
I've loaded the tileset using the example code from `example/index.js`

src/index.js
```diff
@@ -5,6 +5,7 @@ export { B3DMLoader } from './three/loaders/B3DMLoader.js';
 export { PNTSLoader } from './three/loaders/PNTSLoader.js';
 export { I3DMLoader } from './three/loaders/I3DMLoader.js';
 export { CMPTLoader } from './three/loaders/CMPTLoader.js';
+export { SUBTREELoader } from './three/loaders/SUBTREELoader.js';
```
gkjohnson, line 8: There shouldn't be any reason to export this publicaly, I don't think - the only reason PNTS, I3DM etc loaders were exported was incase users wanted to to use these formats separately. With all those formats being deprecated, though, it's possible there's less of a reason to export those, as well, but we can wait on that.

src/three/TilesRenderer.js
```diff
@@ -647,6 +589,18 @@ export class TilesRenderer extends TilesRendererBase {
...
+			case 'subt': {
+
+				const loader = new SUBTREELoader( manager, tile, this.root );
+				loader.workingPath = workingPath;
+				loader.fetchOptions = fetchOptions;
+				loader.parse( buffer );
+				promise = Promise.resolve( null );
+				break;
+
+			}
```
gkjohnson, line 602: Any subtree parsing should happen in the `TilesRendererBase` class since there's nothing three.js specific about it. The goal in the TilesRenderer and TilesRendererBase separation was to keep any no-three.js, 3d-tiles-specific logic in the "Base" class so it's easier to follow. See how external tile sets are loaded in [requestTileConents](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/base/TilesRendererBase.js#L611).

src/three/loaders/SUBTREELoader.js
```diff
@@ -0,0 +1,694 @@
...
+			const leafTile = subtreeLocator.tile;
+			let subtreeTile = this.deriveChildTile(
+				null,
+				leafTile,
+				null,
+				subtreeLocator.childMortonIndex
+			)
+			//	Assign subtree uri as content
+			subtreeTile.content = {uri: parseImplicitURI(subtreeTile, this.rootTile.__basePath, this.rootTile.__subtreeUri)};
+			TilesRenderer.preprocessNode(subtreeTile, this.rootTile.__basePath, leafTile);
```
gkjohnson, line 428: There's no need to call preprocessNode here - the children are preprocessed during traversal to avoid the expensive up front cost:

https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/b3206619ea2a9b20f8aa40c6db8842e69ef89349/src/base/TilesRendererBase.js#L403-L419

src/three/loaders/SUBTREELoader.js
```diff
@@ -0,0 +1,694 @@
+/**
+ * Structure almost identical to Cesium, also the comments and the names are kept
+ * https://github.com/CesiumGS/cesium
+ */
+import {DefaultLoadingManager, Matrix3, Matrix4, Vector3} from 'three';
```
gkjohnson, line 5: I still have to dig into this file a bit more and understand the spec but it looks as though like this basically expands implicit tile nodes into a full hierarchy of children. Is that right?

If that's I think ideally this wouldn't rely on three.js. The tiles can be expanded into in a general structure as they typically are in the tileset.json files. And then the existing node preprocessing parse any necessary fields into three.js constructs used during traversal.

src/base/TilesRendererBase.js
```diff
@@ -308,17 +300,39 @@ export class TilesRendererBase {
...
+			//	Store the infos from the tileset
+			tile.availableLevels = tile.implicitTiling.availableLevels;
+			tile.subdivisionScheme = tile.implicitTiling.subdivisionScheme;
+			tile.__subtreeLevels = tile.implicitTiling.subtreeLevels;
+			tile.subtrees = tile.implicitTiling.subtrees;
+			tile.__subtreeDivider = tile.subdivisionScheme === "QUADTREE" ? 4 : 8;
+			tile.__subtreeUri = tile.implicitTiling.subtrees.uri;
+			tile.__contentUri = tile.content?.uri;
+
+			//	Declare some properties
```
gkjohnson, line 321: nit - can we use spaces rather than tabs after `//` like the rest of the comments

src/base/SubtreeTile.js
```diff
@@ -0,0 +1,45 @@
...
+			return {x: 0, y: 0}
+		}
+
+		const x = 2 * parentTile.__x + (tile.__subtreeIdx % 2);
+		const y = 2 * parentTile.__y + (Math.floor(tile.__subtreeIdx / 2) % 2);
+
+
+		return {x, y};
+
+	};
```
gkjohnson, line 39: Is there a reason we need a special class with a unique member function rather than a assigning these onto the tile object?

gkjohnson (commented):
Thanks for the addition! This is great. I've taken a first pass at things and after looking at it I'm starting to think this might be better suited as a Plugin. I've converted a bit of code to plugins over the last week and in my opinion it helps compartmentalize code and keeps the core of the project smaller to load.

From looking at the code it looks like this would mean adding a plugin hook for `prepreocessNode` and `requestTileContents` to handle expanding the child nodes.

What do you think?

src/three/loaders/SUBTREELoader.js
AnthonyGlt: The issue I've encountered is that we need to have some attributes defined when `preprocessNode` is called.
It looks like `children` needed to be defined, I can easily correct this by adding a children attribute in `SubtreeTile`.
Also, I'm using the `__depth` attribute, so I had to defined it in  `SubtreeTile`. By doing this, it will skip the call you're pointing at, because we will always fall in the condition and break.
I don't really know how to handle this gracefully... In the incoming commit, I introduce a "__level" attribute which is a duplicate from "__depth" :/ and patch this issue

gkjohnson: > It looks like children needed to be defined, I can easily correct this by adding a children attribute in SubtreeTile.

Is there a specific error you're seeing? As far as I can tell the `preprocessNode` function is resilient [to children not being set](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/4870d4f7e6195748638047422ebe0bf75f22da53/src/base/TilesRendererBase.js#L320) - as it needs to be since children isn't required by the [Tile schema](https://github.com/CesiumGS/3d-tiles/blob/main/specification/schema/tile.schema.json).

> I don't really know how to handle this gracefully... In the incoming commit, I introduce a "__level" attribute which is a duplicate from "__depth" :/ and patch this issue

I see this value is being used to calculate the implicit size of the child bounds. Lets use `__level` as an internal field to the SubTreeLoader for now and remove references to `__depth` and we can reevaluate later. Ideally the logic in SubTreeLoader should be self contained and not rely on side effects or methods from the TilesRenderer.

It's also the case that implicit tiling can be added to _any_ child in the tree in which case `__depth` will not start at 0 at that tile so it's probably not the right value to be relying on, anyway. From the spec:

> An implicitTiling object may be added to any tile in the tileset JSON.

AnthonyGlt: > Is there a specific error you're seeing? As far as I can tell the preprocessNode function is resilient [to children not being set](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/4870d4f7e6195748638047422ebe0bf75f22da53/src/base/TilesRendererBase.js#L320) - as it needs to be since children isn't required by the [Tile schema](https://github.com/CesiumGS/3d-tiles/blob/main/specification/schema/tile.schema.json).

I need my tiles to have children (for example [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/650/files#diff-24e9f963ffe77ddc637c982e940d4d8f021e51bf86ab27d170afac4a9599e38aR479) ), they are not preprocessed yet

> I see this value is being used to calculate the implicit size of the child bounds. Lets use __level as an internal field to the SubTreeLoader for now and remove references to __depth and we can reevaluate later. Ideally the logic in SubTreeLoader should be self contained and not rely on side effects or methods from the TilesRenderer.

Yes exactly, the geometricError and the BoundingBox depends on the depth/level and parent value

> It's also the case that implicit tiling can be added to any child in the tree in which case __depth will not start at 0 at that tile so it's probably not the right value to be relying on, anyway. From the spec:

This case hasn't been implemented. I'm only considering the data of the `root` object in the tileset, not the possible children.
My implementation is like a first draft to have a structure of implicit tiling, i've listed the current limitation. 
My idea is that, the support of implicit tiling is a big chunk, that should be sliced in smaller chunks. This MR could be the base to build a complete support of implicit tiling but it is not a full support :building_construction:

src/base/SubtreeTile.js
AnthonyGlt: It was to assign some attributes. Kind of overkill, I agree
Would you rather have a function in the subtreeloader adding these attributes ?

AnthonyGlt:
> Thanks for the addition! This is great. I've taken a first pass at things and after looking at it I'm starting to think this might be better suited as a Plugin. I've converted a bit of code to plugins over the last week and in my opinion it helps compartmentalize code and keeps the core of the project smaller to load.
From looking at the code it looks like this would mean adding a plugin hook for prepreocessNode and requestTileContents to handle expanding the child nodes.
What do you think?

I think it could be done with these hooks. I'm going to try to use your plugin architecture for the implicittiling

src/base/SubtreeTile.js
gkjohnson: I think it's okay to just create a bare javascript object with these fields pre-computed - or ideally compute them as-needed in the associated functions if it doesn't result in too much redundancy. I think ideally we'd be returning a tile hierarchy that shaped as much like the native json [tile schema](https://github.com/CesiumGS/3d-tiles/blob/main/specification/schema/tile.schema.json) from the loader as possible. Then we can rely on the rest of the tiles rendering to just work.

src/three/TilesRenderer.js
AnthonyGlt: Do you mean that I should do 
```
if(isExternalTileSet){
...
} else if(tile.isSubtree){

SAME CODE AS THE CASE !isExternalTileSet
Except I call : 
const loader = new SUBTREELoader( manager, tile, this.root );
loader.workingPath = workingPath;
loader.fetchOptions = fetchOptions;
loader.parse( buffer );

Instead of parseTile

}else{
... 
}

```
Is it correct ?
Could you draft a piece of script so its easier for me to understand ? 

Thank you

gkjohnson: Yeah let's start with that, for now, and we can dedupe some of the load logic later on. Here's an example of what I'm imagining with some abbreviated function contents. Here "SubTreeLoader" is returning a parsed root with children that is shaped like any other 3d tiles tile from the schema:

```js
if ( isExternalTileSet ) {

    // ...

} else if ( /subt$/i.test( tile.content.uri ) ) {

    downloadQueue.add( tile, tileCb => {

        // ...

        return fetch( processedUrl, Object.assign( { signal }, this.fetchOptions ) );

    } )
        .then( buffer => {

            // ...

            return res.arrayBuffer();

        } )
        .then( buffer => {

            // ...

            // handle parsing of subtree data and assignment of tile children
            const loader = new SubTreeLoader();
            const subTreeRoot = loader.parse( buffer );
            tile.children.push( subTreeRoot );

        } )
        .catch( errorCallback );

} else {

    // ...

}
```

Let me know if that answers things

AnthonyGlt:
@gkjohnson I've commited a beginning of a plugin implem.
Could you check it out ? [Draft commit](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/650/commits/949f77dd4fe69795df00904e35aa73908d6c2b3d) 

Some things I wonder:
1. the hook is at the end ‎of TilesRendererBase.preprocessNode :
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/949f77dd4fe69795df00904e35aa73908d6c2b3d/src/base/TilesRendererBase.js#L392-L396
Is it an issue ?

2. we don't use the magictype, only the file extension
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/949f77dd4fe69795df00904e35aa73908d6c2b3d/src/base/TilesRendererBase.js#L662

4. isn't better to have a hook for `parseTile` ?
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/949f77dd4fe69795df00904e35aa73908d6c2b3d/src/base/TilesRendererBase.js#L824

Instead of the
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/949f77dd4fe69795df00904e35aa73908d6c2b3d/src/base/TilesRendererBase.js#L662

5. I was wondering if i could squeeze `parseImplicitURI` in `preprocessURL` of my plugin but it require to have `tile`

WDYT ?

For the SUBTREELoader, does the plugin structure has an impact on it or I can leave it untouch, as it is ? 
Thank you

gkjohnson:
> 1. the hook is at the end ‎of TilesRendererBase.preprocessNode :
> ...
> Is it an issue ?

This looks right.

> 2. we don't use the magictype, only the file extension

Not every file type has a set of magic bytes, unfortunately, and we're relying on file extensions everywhere else. The least we can do is check and [assert the magic bytes](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/7d9ffd29be49ae91e8b063d7fff11905845b2de7/src/base/loaders/B3DMLoaderBase.js#L19-L21) for now. If this somehow becomes an issue in the future because tile sets aren't using matching extensions we can revisit this for all the file types.

> 3. isn't better to have a hook for parseTile ?

That's not a bad idea 🤔 I agree I think we can use `invokeOnePlugin( ... )`, instead, with a `parseTile` hook.

> 5. I was wondering if i could squeeze parseImplicitURI in preprocessURL of my plugin but it require to have tile

Maybe I'm misunderstanding but why does `parseImplicitURI` need to be handled in `preprocessNode` or a plugin function? Can't it be expanded for the children already in `SUBTREELoader`? Ie by the time a child is generated and returned the `content.uri` value be something like `0/1/2/3.glb` from the implicit tiles `{ level }/{ x }/{ y }/{ z }.glb` uri? Then it will be automatically concatenated with `__basePath` in `preprocessNode`.

> For the SUBTREELoader, does the plugin structure has an impact on it or I can leave it untouch, as it is ?

- Lets remove the need for the `SubTreeTile` class or at least remove the function member since all other tiles are just defined as bare objects without functions.
- Lets merge SUBTREELoader and SUBTREELoaderBase into one class in the plugins folder. Ideally this class would be rendering engine agnostic  and not require three.js classes but I see that some matrix math is required so we can leave that for now and I'll take a look at it later. But let's at least remove the `adjustmentTransform` and `manager` fields since they aren't used.

AnthonyGlt:
The files are in src/three/plugins for now
Should we add a plugin directory in src ?
Currently, as you said I use threejs for the matrix and vector.
We could do these calculus by hand, element by element if we want to get rid of threejs.

Concerning the [TODOs](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/650#issue-2429584772), we could treat them separately in others MR once this one will be merged, WDYT?

src/base/TilesRendererBase.js
```diff
@@ -302,13 +302,6 @@ export class TilesRendererBase {
 
 			}
 
-			if ( tile.content.uri ) {
-
-				// tile content uri has to be interpreted relative to the tileset.json
-				tile.content.uri = new URL( tile.content.uri, tileSetDir + '/' ).toString();
-
-			}
-
```
gkjohnson, line 311: What's the reason for this having to move?

src/index.d.ts
```diff
@@ -6,6 +6,7 @@ export { B3DMLoader } from './three/loaders/B3DMLoader';
 export { I3DMLoader } from './three/loaders/I3DMLoader';
 export { PNTSLoader } from './three/loaders/PNTSLoader';
 export { CMPTLoader } from './three/loaders/CMPTLoader';
+export { SUBTREELoader } from './three/plugins/SUBTREELoader';
```
gkjohnson, line 9: This isn't exported in index.js

src/three/TilesRenderer.js
```diff
@@ -505,6 +505,7 @@ export class TilesRenderer extends TilesRendererBase {
 
 		super.preprocessNode( tile, tileSetDir, parentTile );
 
+
```
gkjohnson, line 508: nit: Can we remove some of these unnecessary whitespace changes in the PR?

src/three/TilesRenderer.js
```diff
@@ -668,13 +672,17 @@ export class TilesRenderer extends TilesRendererBase {
 		// get the scene data
 		let scene;
 		let metadata;
-		if ( result.isObject3D ) {
+
+		if(!result){
+
+			return;
+
+		} else if ( result.isObject3D ) {
```
gkjohnson, line 680: This condition should no longer be needed

src/three/TilesRenderer.js
```diff
@@ -865,6 +873,9 @@ export class TilesRenderer extends TilesRendererBase {
 	setTileVisible( tile, visible ) {
 
 		const scene = tile.cached.scene;
+		if(!scene){
+			return;
+		}
```
gkjohnson, line 878: This condition shouldn't be needed

src/three/plugins/ImplicitTilingPlugin.d.ts
```diff
@@ -0,0 +1,3 @@
+export class ImplicitTilingPlugin {
+	constructor();
+}
```
gkjohnson, line 3: Is it necessary to provide a constructor definition here? I'm not a typescript expert so I'm not sure. Is this enough?

```ts
export class ImplicitTilingPlugin {}
```

src/three/plugins/ImplicitTilingPlugin.js
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	init(tiles) {
+
+		this.tiles = tiles;
+
+	}
+
+	preprocessNode(tile, uri) {
+		if (tile.implicitTiling) {	//only for root
```
gkjohnson, line 16: When we say "root" - does this mean the root of the tileset? Or the root of an implicit subtree, which could be 5 tiles deep, for example?

src/three/plugins/ImplicitTilingPlugin.js
```diff
@@ -0,0 +1,56 @@
...
+
+			// Store the infos from the tileset
+			tile.__availableLevels = tile.implicitTiling.availableLevels;
+			tile.__subdivisionScheme = tile.implicitTiling.subdivisionScheme;
+			tile.__subtreeLevels = tile.implicitTiling.subtreeLevels;
+			tile.__subtrees = tile.implicitTiling.subtrees;
+			tile.__subtreeDivider = tile.__subdivisionScheme === "QUADTREE" ? 4 : 8;
+			tile.__subtreeUri = tile.implicitTiling.subtrees.uri;
+			tile.__contentUri = uri ?? tile.content?.uri;
```
gkjohnson, line 25: What's the reason for duplicating this information? It looks like it's already available on the tile in the implicitTiling object - why not use those fields?

src/three/plugins/SUBTREELoader.d.ts
```diff
@@ -0,0 +1,9 @@
+import { LoadingManager } from 'three';
+
+export class SUBTREELoader {
+
+	constructor( manager : LoadingManager, tile, rootTile );
+	load( url : String ) : void ;
+	parse( buffer : ArrayBuffer ) : void ;
+
+}
```
gkjohnson, line 9: We're not exporting the SUBTREELoader so we don't need a d.ts file. I've only been adding these for "public" files to avoid adding more than necessary to maintain.

src/three/plugins/SUBTREELoader.js
```diff
@@ -0,0 +1,747 @@
+/**
+ * Structure almost identical to Cesium, also the comments and the names are kept
+ * https://github.com/CesiumGS/cesium
```
gkjohnson, line 3: Is there a specific file this is from? How much of this code is copy-pasted? I'd recommend taking a look at the Cesium license to make sure we're following what's expected under their Apache-2 license:

https://github.com/CesiumGS/cesium?tab=Apache-2.0-1-ov-file#readme

src/three/plugins/SUBTREELoader.js
```diff
@@ -0,0 +1,747 @@
...
+
+	/**
+	 * A helper object for storing the two parts of the subtree binary
+	 *
+	 * @typedef {object} Subtree
+	 * @property {number} version
+	 * @property {JSON} subtreeJson
+	 * @property {ArrayBuffer} subtreeByte
+	 * @private
+	 */
```
gkjohnson, line 29: Is this a comment from a deleted function?

src/three/raycastTraverse.js
```diff
@@ -26,7 +26,7 @@ function intersectTileScene( scene, raycaster, intersects ) {
 		} );
 		_hitArray.sort( distanceSort );
 
-	} else {
+	} else if( scene ) {
```
gkjohnson, line 29: This shouldn't be needed

src/three/plugins/SUBTREELoader.js
```diff
@@ -0,0 +1,747 @@
...
+	getTileBoundingVolume(tile) {
+		let boundingVolume;
+		if (this.rootTile.boundingVolume.region) {
+			boundingVolume = [...this.rootTile.boundingVolume.region];
+			const minX = boundingVolume[0];
+			const maxX = boundingVolume[2];
+			const minY = boundingVolume[1];
+			const maxY = boundingVolume[3];
+			const sizeX = (maxX - minX) / (Math.pow(2, tile.__depth ?? tile.__level));
+			const sizeY = (maxY - minY) / (Math.pow(2, tile.__depth ?? tile.__level));
```
gkjohnson, line 620: I think I mentioned this before but there shouldn't be any reason to use `__depth` here. `__level` is the more correct value to use, no?

src/three/plugins/SUBTREELoader.js
```diff
@@ -0,0 +1,747 @@
...
+
+
+	parseImplicitURI(tile, basePath, uri) {
+		uri = uri.replace("{level}", (tile.__depth ?? tile.__level) ?? 0);
+		uri = uri.replace("{x}", tile.__x ?? 0);
+		uri = uri.replace("{y}", tile.__y ?? 0);
+		uri = uri.replace("{z}", tile.__z ?? 0);
+		uri = new URL(uri, basePath + '/').toString();
+		return uri;
+	}
```
gkjohnson, line 745: We shouldn't need to pass the base path or anything here. Lets just replace the "level", "x", "y", and "z" values. Then the url will be expanded to a full uri in the `preprocessNode` function.

src/base/TilesRendererBase.js
```diff
@@ -716,7 +719,7 @@ export class TilesRendererBase {
 						const uri = parseTile.content.uri;
 						const extension = getUrlExtension( uri );
 
-						return this.parseTile( buffer, parseTile, extension );
+						return this.invokeOnePlugin( plugin => plugin.parseTile && plugin.parseTile( buffer, parseTile, extension ));
```
gkjohnson, line 722: We need to make sure that any child tiles created and added to the tile set are disposed of correctly when this tile is evicted from the cache. Effectively any members and child tiles added to the tile when "parsing" it should be disposed so when this tile is loaded again we can re load and regenerate the implicit children.

src/three/plugins/ImplicitTilingPlugin.js
gkjohnson, line 16: This function should set `tile.__hasUnrenderableContent` to `true` so the rest of the tile traversal logic can work as expected. This also means that the external tile set detection condition in `requestTileContents` will have to be modified to check for the `/json$/` extension rather than checking the above flag

gkjohnson (commented):
Added some more comments - I still have to take a look at SUBTREELoader to make sure I understand exactly what's happening but overall the structure looks good.

src/base/TilesRendererBase.js
AnthonyGlt: I need to keep the original content uri.
I've reverted some changes, but I had to move up the `uri` declaration.

src/three/TilesRenderer.js
AnthonyGlt: It is necessary to keep it because the first parsing of the root subtree doesn't return 3d object so no scene has been initialized.

src/three/raycastTraverse.js
AnthonyGlt: Same reason as above

src/three/TilesRenderer.js
gkjohnson: Loading external tile sets also don't produce 3d geometry but this isn't needed for those cases. It's likely related to https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/650#discussion_r1706242631 - if the unrenderable / renderable content flags are set correctly then `setVisible` etc should never be called for geometry that doesn't have 3d geometry.

src/base/TilesRendererBase.js
AnthonyGlt: Looks to me that it got disposed as any "classic" children.
Only the .subtree files will be loaded from the plugin, they don't have geometry or anything. Do you think will be issues with this ?

src/three/TilesRenderer.js
AnthonyGlt: Sorry I've removed my comment...
The comment was a misconception on my side.
(Old comment: ~~It is necessary to keep it because the first parsing of the root subtree doesn't return 3d object so no scene has been initialized or am I missing something here ?~~)

The `preprocessNode` method of my plugin handles only the tileset root, that's why the flags weren't well defined for the children. Thank you :+1:

AnthonyGlt: I've fixed it, thank you.
I'm not proud of my condition though. I need something to detect that it's a subtree, without content or with an other subtree as content ...
```js	
	}else if (/.subtree$/i.test(tile.content?.uri) || (tile.__subtreeIdx !== undefined && !uri)){
			tile.__hasUnrenderableContent = true;
			tile.__hasRenderableContent = false;
			
```

src/three/plugins/ImplicitTilingPlugin.js
AnthonyGlt: It's for the root of the tileset :+1:

src/base/TilesRendererBase.js
```diff
@@ -390,6 +392,9 @@ export class TilesRendererBase {
 
 		tile.__lastFrameVisited = - 1;
 
+		this.invokeAllPlugins( plugin => {
+			plugin !== this && plugin.preprocessNode && plugin.preprocessNode(tile, uri);
```
gkjohnson, line 396: We should pass the parent tile in to this function the same way it is in the main function.

src/three/TilesRenderer.js
gkjohnson: I'm not fully following how or why the `uri` argument could be null or undefined here. This should be the `tileSetDir` as is passed into the main `preprocessNode` function, no? And why do we need to adjust the content setting based on the `__subTreeIndex` field?

It should be entirely based on the `content` value. If a tile has no content uri then both "has renderable content" and "has unrenderable content" should be false. If it has content that is renderable geometry then only "has renderable content" should be true. And if it has content that points to a file that is used to expand children or somehow load something that _isn't_ renderable (json file, subtree file) then only "has unrenderable content" should be true.

src/three/plugins/SUBTREELoader.js
AnthonyGlt: No it's a typedef to define the output of `parseBuffer`. Isn't really useful but it can help for the understanding of the code in the IDE

src/three/plugins/ImplicitTilingPlugin.js
AnthonyGlt: It was to avoid the longer call `tile.implicitTiling.XXX`.
I'll remove some. The ones I keep seem to be relevant but could also be removed if you deem it necessary

src/base/TilesRendererBase.js
gkjohnson: What do you mean by "classic" children? As in children with geometry?

When a tile is no longer meeting the error target or in the frustum it's queued up for disposal if it has any kind of content that needed to be loaded. This includes [external tile sets](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/7f95332ea3f31a97c1b226e3ea38b8ed720183ab/src/base/TilesRendererBase.js#L538-L542). The children that are loaded (or generated in this case) still take up CPU memory and with enough loaded that could be substantial.

For now the subtree children are only being expanded for the root tile but those children should be disposed of to be consistent with the existing behavior of the TilesRenderer and will be more necessary once deeper child subtrees are supported.

src/three/plugins/SUBTREELoader.js
AnthonyGlt: https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Scene/Implicit3DTileContent.js

The structure is the same. Parsing/reading the json/binary -> Create children -> Filling the uri if a content/subtree file is available. I've kept the names and comment for clarity.
The license is the same as this repo, which is very permissive.

AnthonyGlt:
> I'm not fully following how or why the uri argument could be null or undefined here. This should be the tileSetDir as is passed into the main preprocessNode function, no? And why do we need to adjust the content setting based on the __subTreeIndex field?
It should be entirely based on the content value. If a tile has no content uri then both "has renderable content" and "has unrenderable content" should be false. If it has content that is renderable geometry then only "has renderable content" should be true. And if it has content that points to a file that is used to expand children or somehow load something that isn't renderable (json file, subtree file) then only "has unrenderable content" should be true.

Absolutely, I've must have messed up my tests. Looks like it's working fine :+1:

src/base/TilesRendererBase.js
AnthonyGlt: Yes, although i'm also passing the `uri` instead of the `tileSetDir`. Maybe this hook could be cleaner :/

src/base/TilesRendererBase.js
AnthonyGlt: Could it be enough to check it via the flag `__hasRenderableContent`, WDYT? As following:
```js
} else if ( isExternalTileSet || !t.__hasRenderableContent) {
    t.children.length = 0;
}
```

gkjohnson: I think I'd prefer the plugin be responsible for cleaning up at after itself since it adds the children instead of having the creation and clean up happen in two separate places.

src/base/TilesRendererBase.js
gkjohnson: Oh I see, now. Hm. It should just take the same arguments at the parent function but I see that you need to keep the  original content uri definition since it's used to expand the content uris. Maybe it's best if we don't modify the content uri in the preprocessNode function - but lets deal with that in another PR and fix the arguments passed here.

src/three/plugins/ImplicitTilingPlugin.js
```diff
@@ -0,0 +1,63 @@
...
+			// Coords of the tile
+			tile.__x = 0;
+			tile.__y = 0;
+			tile.__z = 0;
+
+			let implicitUri = tile.__subtreeUri.replace("{level}", (tile.__depth ?? tile.__level) ?? 0);
+			implicitUri = implicitUri.replace("{x}", "0");
+			implicitUri = implicitUri.replace("{y}", "0");
+			implicitUri = implicitUri.replace("{z}", "0");
+			tile.content.uri = new URL(implicitUri, tile.__basePath + '/').toString();
```
gkjohnson, line 43: Is it necessary to assign this? The tile is marked as unrenderable so the content uri will never be used. Or is it the case that this tile _is_ renderable?

src/three/plugins/ImplicitTilingPlugin.js
gkjohnson: Perhaps I'm misunderstanding but from reading the spec it looks like this is _not_ only for the root of the tile set. It's a bit confusing but the spec [makes a distinction](https://github.com/CesiumGS/3d-tiles/tree/main/specification/ImplicitTiling#implicit-root-tile) between the root of the tile set and the root of an implicit subtree. Note that it says the "implicitTiling" object can be added to any tile in the tile set:

> An implicitTiling object may be added to ***any tile*** in the tileset JSON. Such a tile is called an implicit root tile, to distinguish it from the root tile of the tileset JSON.

src/three/plugins/ImplicitTilingPlugin.js
```diff
@@ -0,0 +1,63 @@
...
+			tile.__hasRenderableContent = false;
+		}
+
+	}
+
+	parseTile(buffer, parseTile, extension) {
+
+		//todo use extension instead ?
+		if (/.subtree$/i.test(parseTile.content.uri)) {
+			const loader = new SUBTREELoader(parseTile, this.tiles.root);
```
gkjohnson, line 56: Given https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/650#discussion_r1710874576 it doesn't seem correct to pass the tile set root tile into this object. Rather this should be the implicit tiling root that expanded this tile.

src/three/plugins/SUBTREELoader.js
gkjohnson: Thanks - I still need to read through the SUBTREELoader to make sure I understand everything and am reading through the spec to make sure everything lines up, as well. But we should add a link using the latest commit hash to this file in the comment so it's clear what was referenced.

Thanks for bearing with me on the questions and changes. This is a pretty complex addition.

src/three/plugins/ImplicitTilingPlugin.js
AnthonyGlt: Yes you're right, a child declaring implicitTiling object should meet this condition. The comment is misleading. As I said before, I didn't test the case where a child is declaring an implicitTiling object. I've only implemented the most straightforward configuration, e.g with the implicitTiling object in the root of the tileset.

src/three/plugins/SUBTREELoader.js
AnthonyGlt: I'll add the permalink
https://github.com/CesiumGS/cesium/blob/0a69f67b393ba194eefb7254600811c4b712ddc0/packages/engine/Source/Scene/Implicit3DTileContent.js

> Thanks for bearing with me on the questions and changes. This is a pretty complex addition.

It's my first contribution to your lib, I'm not used to its code, so I was expecting this back and forth :+1:

src/three/plugins/ImplicitTilingPlugin.js
AnthonyGlt: In this line, we replace the content uri:
```js
  "content": {
      "uri": "content/{level}_{x}_{y}.glb"
    },

```
by 
```js
  "content": {
      "uri": "content/0_0_0.subtree"
    },

```
This will be the root subtree, so it needs to be fetch and parse, even though the file .subtree itself doesn't contains 3D object but only information on content availability.

These flags
``` js
tile.__hasUnrenderableContent = true;
tile.__hasRenderableContent = false;
```
seem to still allow the fetch and parse. Is that a bug ?
What''s the meaning of the 2 RenderableContent flags ? Renderable doesn't mean 3D data to be render on the screen ?
Am I not using it correctly ?

src/three/plugins/ImplicitTilingPlugin.js
AnthonyGlt: Absolutely.
Then I need to find a way to keep this this expanding tile.
As I've commented on the previous comment, I didn't test this case :/

AnthonyGlt:
@gkjohnson I've added data that you can use to test this MR:

https://raw.githubusercontent.com/AnthonyGlt/3DTilesRendererJS/i3dm/example/batiments_rieux_glb/tileset.json

Should look like this:

image

src/base/TilesRendererBase.js
AnthonyGlt: You mean like this : 
```js
	if ( t.__loadingState === LOADING ) {

		t.__loadAbort.abort();
		t.__loadAbort = null;

	} else if ( isExternalTileSet ) {

		t.children.length = 0;

	} else {

		this.invokeAllPlugins( plugin => {

			plugin.disposeTile && plugin.disposeTile(t);

		} );
	}
```

With inside `ImplicitTilingPlugin.js` : 

```js
disposeTile(tile){
  if (/.subtree$/i.test(tile.content?.uri)){
	  tile.children.length = 0;
  }
}
```

src/three/plugins/ImplicitTilingPlugin.js
gkjohnson: Okay I see now - I was under the impression that there was a separate of the spec that handled child subtrees differently that wasn't implemented. Since this is implemented generically the logic for parsing shouldn't be any different - especially since we're using __level now instead of __depth. We should just pass the implicit tiling root in, instead. The associated implicit root can be found by either climbing the tree until the implicit root is found or by saving a reference to it on all child tiles.

src/base/TilesRendererBase.js
gkjohnson: Yup! Like that.

src/three/plugins/ImplicitTilingPlugin.js
gkjohnson: > In this line, we replace the content uri:
> ...
>      "uri": "content/{level}_{x}_{y}.glb"
> ...
> by
> ...
>      "uri": "content/0_0_0.subtree"

OH. Sorry I missed that the original glb extension was replaced with the subtree uri.

> What''s the meaning of the 2 RenderableContent flags ? Renderable doesn't mean 3D data to be render on the screen ?

There are three content flags:

- `__hasContent` mean that the tile has content of any kind that has to be loaded - whether that be 3d geometry or data that should be expanded to child tiles.
- `__hasUnrenderableContent` means that the tile has content like an external tile set but has no geometry to render.
- `__hasRenderableContent` means the tile has content that is renderable like 3d geometry or points.

Given the above extension mixup, you're using this correctly, sorry about that.

src/three/plugins/ImplicitTilingPlugin.js
gkjohnson: This should be a minor change, I think - see https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/650#discussion_r1713001941.

src/three/plugins/ImplicitTilingPlugin.js
```diff
@@ -0,0 +1,63 @@
...
+			tile.__contentUri = uri ?? tile.content?.uri;
+
+			// Declare some properties
+			tile.__subtreeIdx = 0;	// Idx of the tile in its subtree
+
+			// Coords of the tile
+			tile.__x = 0;
+			tile.__y = 0;
+			tile.__z = 0;
+			tile.__level = tile.__depth;
```
gkjohnson, line 38: Since this is _always_ the implicit root of sub tree should the level no always start off at 0? From the spec:

> level is `0` for the implicit root tile. This tile’s children are at level 1, and so on.

src/three/plugins/SUBTREELoader.js
```diff
@@ -0,0 +1,730 @@
...
+				// Create a child holding the content uri, this child is similar to its parent and doesn't have any children
+				let contentTile = {
+					content: {uri: this.parseImplicitURI(subtreeRoot, this.rootTile.__contentUri)},
+					boundingVolume: subtreeRoot.boundingVolume,
+					geometricError: subtreeRoot.geometricError
+				}
+				subtreeRoot.children.push(contentTile);
+				break;
+			}
+		}
```
gkjohnson, line 448: Can you explain this? It looks like this child is added in order to support geometry for the level 0 tile but I'm not seeing how the next level children are added to this tile. From [transcodeSubtreeTiles](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/650/files#diff-e1ca0fc4e2fdac866a35e569dee4e3e37a9545e4fd24c395eebbd5ca998a0b1cR489) it looks like all the next level 1 children are all also added as children to the subtree root meaning the level 0 tile and level 1 tiles would be siblings. Is that correct? In order for the structure of the tile set to be correct and for tile traversal to work correctly the level 1 tiles need to be children of the level 0 content tile.

It's okay to _always_ add another level 0 content tile (even if it's empty) here and then expand the children into that to handle this case.

gkjohnson:
Made some comments for the last things that I think need to be addressed. I think after these things are handled we can leave the rest to future PRs.

src/three/plugins/ImplicitTilingPlugin.js
AnthonyGlt: I'm now storing the Implicit Root Tile as `__implicitRoot`and it's carried by each children. It means the Implicit Root Tile is storing itself (circular reference).

AnthonyGlt:
@gkjohnson in this commit eef3c4eaee24dccd1c8af7570d065031f53cff2f , I had to continue the loop because during my tests in adding in the tileset children who have implicitTiling object, I've noticed that we will first preprocess all the children from the tileset, and then we will add the newly created subtree children.
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/eef3c4eaee24dccd1c8af7570d065031f53cff2f/src/base/TilesRendererBase.js#L428-L435

This means that the children defined in the tileset will have the `__depth` attribute and break the loop, even though the new children haven't be preprocessed.
We are now losing in performance, maybe we could think to replace this condition by another one.

src/three/plugins/SUBTREELoader.js
AnthonyGlt: You're correct, it's exactly what happens here.
I add the level 0 tile with content as a children and I add all the children to the "original" level 0 tile (the one with subtree uri). So as you said the level 0 tile will be siblings with the children tiles :/
I cannot replace the level 0 tile subtree by the level 0 tile content, they both need to exist, we agree on this.

> It's okay to always add another level 0 content tile (even if it's empty) here and then expand the children into that to handle this case.

1. You mean I should use the one I've just created and call `transcodeSubtreeTiles` with it ? Like this:

```js
let contentTile;
for ( let i = 0; subtree && i < subtree._contentAvailabilityBitstreams.length; i ++ ) {

	if ( subtree && this.getBit( subtree._contentAvailabilityBitstreams[ i ], 0 ) ) {

		// Create a child holding the content uri, this child is similar to its parent and doesn't have any children
		contentTile = {
			content: { uri: this.parseImplicitURI( subtreeRoot, this.rootTile.__contentUri ) },
			boundingVolume: subtreeRoot.boundingVolume,
			geometricError: subtreeRoot.geometricError
		};
		subtreeRoot.children.push( contentTile );
		break;

	}

}


// Creating each leaf inside the current subtree
const bottomRow = this.transcodeSubtreeTiles(
	contentTile || subtreeRoot,  //HERE I'M USING THE NEW TILE AS PARENT
	subtree
);

```

2. Or I should add it the parentTile:

```js

for ( let i = 0; subtree && i < subtree._contentAvailabilityBitstreams.length; i ++ ) {

	if ( subtree && this.getBit( subtree._contentAvailabilityBitstreams[ i ], 0 ) ) {

		// Create a child holding the content uri, this child is similar to its parent and doesn't have any children
		const contentTile = {
			content: { uri: this.parseImplicitURI( subtreeRoot, this.rootTile.__contentUri ) },
			boundingVolume: subtreeRoot.boundingVolume,
			geometricError: subtreeRoot.geometricError
		};
		subtreeRoot.parent.children.push( contentTile ); // HERE I add it to the parent
		break;

	}

}

```

3. Or I combine both ?


The last one (`subtreeRoot.parent.children.push`) was causing issue with the break the `__depth` condition on the child in `ensureChildrenArePreprocessed`.
But if we replace `break` by `continue`, there is no issue (cf https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/650#issuecomment-2283609240) . Except in the case of a subtree root tile who also has content, because it won't have parent... So maybe I should just keep the 1st implem (use `contentTile` as parent for the other children)

gkjohnson:
> This means that the children defined in the tileset will have the __depth attribute and break the loop, even though the new children haven't be preprocessed.

Can you explain the exact scenario in which some children will have been processed and some children have not? From looking at the code it doesn't look like there should ever be a situation in which new some siblings are added to a parent that already has children. They should all be processed at once.

src/three/plugins/ImplicitTilingPlugin.js
gkjohnson: Great! Regarding https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/608#issuecomment-2283364305 - does it make sense to change this line item:

> ImplicitTiling defined in children. Only the one in root is supported

to

> Test implicit tiling defined in children

It _should_ be supported since we've implemented what's needed against the spec. It just hasn't been tested.

src/three/plugins/SUBTREELoader.js
AnthonyGlt: My first implementation isn't completely correct though.
As you said, I should always add the empty level 0 tile but to make it work I need to have some attributes.
I don't see how to handle this properly. I need to copy the attributes from the original tile know that the SubtreeTile class did bother you.
To make it more comprehensible, you can check my latest commit 87e9c3193677c14f0a51e39042fbd12ed6576112
This static function doesn't satisfy me

gkjohnson: It's okay - it could just be a function that returns the cloned element but the static functions works. The only thing that sticks out to me is this line:

```
copyTile.parent = tile.parent;
```

The "copied tiles" parent should not be set to the same as the parent of it's _actual_ parent (the tile it's a child of). This will get overwritten in preprocess node, though, so there's no reason to set the parent here or in the SubTreeTile constructor.

---

Ultimately this is an artifact of the fact that "multiple contents" isn't support right now. If multiple contents were supported then this tile could contain both renderable and un renderable content. We can reevaluate this in the future. Maybe worth just adding a comment inline to note that.

AnthonyGlt:
> > This means that the children defined in the tileset will have the __depth attribute and break the loop, even though the new children haven't be preprocessed.
> 
> Can you explain the exact scenario in which some children will have been processed and some children have not? From looking at the code it doesn't look like there should ever be a situation in which new some siblings are added to a parent that already has children. They should all be add and then later processed at once.
> 
> I'm also not seeing this situation when testing.

To try the case of the children defined in the tileset, I've used the data from https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/650#issuecomment-2277638571 and replace the tileset by:

```json

{
  "geometricError": 2000.0,
  "root": {
    "transform": [
      1.0,
      0.0,
      0.0,
      0.0,
      0.0,
      1.0,
      0.0,
      0.0,
      0.0,
      0.0,
      1.0,
      0.0,
      4625478.5,
      118434.4296875,
      4375219.0,
      1.0
    ],
    "geometricError": 2000.0,
    "refine": "ADD",
    "boundingVolume": {
      "region": [
        0.02542394815546297,
        0.7606475623375515,
        0.02577445511321351,
        0.7609346533601059,
        145.974731,
        196.594
      ]
    },
    "content": {
      "uri": "content/{level}_{x}_{y}.glb"
    },
    "implicitTiling": {
      "availableLevels": 6,
      "subdivisionScheme": "QUADTREE",
      "subtreeLevels": 3,
      "subtrees": {
        "uri": "subtrees/{level}_{x}_{y}.subtree"
      }
    },
	  "children":[{
		  "boundingVolume": {
			  "region": [
				  0.02542394815546297,
				  0.7606475623375515,
				  0.02577445511321351,
				  0.7609346533601059,
				  145.974731,
				  196.594
			  ]
		  },
		  "content": {
			  "uri": "content/{level}_{x}_{y}.glb"
		  },
		  "implicitTiling": {
			  "availableLevels": 6,
			  "subdivisionScheme": "QUADTREE",
			  "subtreeLevels": 3,
			  "subtrees": {
				  "uri": "subtrees/{level}_{x}_{y}.subtree"
			  }
		  }
	  }]
  },
  "asset": {
    "version": "1.1"
  }
}


```

I've just copy/paste the implicitTiling object in a child, for testing.
By doing this, I've noticed that the children from the tileset will be preprocess and then the attribute __depth will be added to it. 
Then we will add other children from the subtree loader.
But since the first child has already been preprocessed, the condition 
```js
 if ( '__depth' in child ) {
```
will be true and loop will stop. Meaning that the other children won't be preprocessed.

Isn't it the case ?
Maybe we should have a condition on the parent instead of being on the child, WDYT ?

gkjohnson:
That's not a valid tile set, though. From the section on [the implicit tiling root in the spec](https://github.com/CesiumGS/3d-tiles/tree/main/specification/ImplicitTiling#implicit-root-tile):

> The tile shall omit the children property

So there are not allowed to be any children in the implicit root. It might be worth reviewing the spec again in detail so we can make sure we're not being overly defensive in cases that don't require it which might cause issues later on.

AnthonyGlt:
> That's not a valid tile set, though. From the section on [the implicit tiling root in the spec](https://github.com/CesiumGS/3d-tiles/tree/main/specification/ImplicitTiling#implicit-root-tile):
> 
> > The tile shall omit the children property
> 
> So there are not allowed to be any children in the implicit root. It might be worth reviewing the spec again in detail so we can make sure we're not being overly defensive in cases that don't require it which might cause issues later on.

Mmmh
The tileset could be like this one then : 
```json
{
  "geometricError": 2000.0,
  "root": {
    "transform": [
      1.0,
      0.0,
      0.0,
      0.0,
      0.0,
      1.0,
      0.0,
      0.0,
      0.0,
      0.0,
      1.0,
      0.0,
      4625478.5,
      118434.4296875,
      4375219.0,
      1.0
    ],
    "geometricError": 2000.0,
    "refine": "ADD",
    "boundingVolume": {
      "region": [
        0.02542394815546297,
        0.7606475623375515,
        0.02577445511321351,
        0.7609346533601059,
        145.974731,
        196.594
      ]
    },
    "content": {
		"uri": "content/2_1_0.glb"
    },
	  "children":[{
		  "boundingVolume": {
			  "region": [
				  0.02542394815546297,
				  0.7606475623375515,
				  0.02577445511321351,
				  0.7609346533601059,
				  145.974731,
				  196.594
			  ]
		  },
		  "content": {
			  "uri": "content/{level}_{x}_{y}.glb"
		  },
		  "implicitTiling": {
			  "availableLevels": 6,
			  "subdivisionScheme": "QUADTREE",
			  "subtreeLevels": 3,
			  "subtrees": {
				  "uri": "subtrees/{level}_{x}_{y}.subtree"
			  }
		  }
	  }]
  },
  "asset": {
    "version": "1.1"
  }
}

```
Which seems to work without modifying the __depth condition, so I can revert this change

gkjohnson:
Excellent! This is great, thank you. This is a large addition and we'll can continue to improve it over time.

Issue #663, opened 2024-08-02
Cesium OSM tile set with "ADD" refinement loads very slowly


gkjohnson:
Related to #662 

The ion asset with id "96188" loads open street map buildings using the "additive" refinement all the way down. However it takes an extremely long time to load everything and requires setting the lru cache to a size of 6000 to load everything. It's possible that traversal of the additive refinement is being handled incorrectly. Setting `loadSiblings` to false results in the lru cache only being filled with ~600 tiles, instead. It's possible we need to handle the additive refinement case with different logic to avoid forcing all siblings to load in this situation.

![image](https://github.com/user-attachments/assets/37ea80bc-3701-4ffb-92a4-80270e2e4723)

The raycast performance is also very poor.

Issue #662, opened 2024-08-01
Cesium OSM Buildings from Cesium ion are not displayed 


jailln:
1. Go to https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/ionExample.html
2. Add your access token and the following asset id: 96188
3. See that nothing gets displayed.

There is no specific error. When looking at the network pane, Cesium ion connection is ok and the tileset and root tile are correctly retrieved so it doesn't seem related to cesium ion specifically. Bounding volumes are regions, could it be related to #624 ?

gkjohnson:
The tile set you're loading is the size of the globe which the ion example page is not designed for. If you load it with the appropriate camera controls that account for the WGS84 ellipsoid the tiles can load in, though very slowly:

<img width="651" alt="image" src="https://github.com/user-attachments/assets/7861f5da-51be-426a-92d8-2745b4b43ded">

I'll create another issue for that.

gkjohnson:
Closing in favor of #663

gkjohnson:
> If you load it with the appropriate camera controls that account for the WGS84 ellipsoid the tiles can load in, though very slowly

For more context on this I added a new TilesRenderer that loads the OSM data as a sibling tile set in the google maps globe example.

jailln:
> The tile set you're loading is the size of the globe which the ion example page is not designed for. If you load it with the appropriate camera controls that account for the WGS84 ellipsoid the tiles can load in, though very slowly.

Thanks! Sorry I missed that, I tested it in itowns first and then created a quick reproduction on 3DTilesRendererJS without paying enough attention. Turns out there was an issue anyway, thanks for the quick fixes :)

> For more context on this I added a new TilesRenderer that loads the OSM data as a sibling tile set in the google maps globe example the lru cache max size was increased to ~6000.

Just to be sure, the cache size doesn't have an impact on whether more tiles are rendered or not, does it? It's just that newly loaded and rendered tiles are not cached?

gkjohnson:
> Just to be sure, the cache size doesn't have an impact on whether more tiles are rendered or not, does it? It's just that newly loaded and rendered tiles are not cached?

[...]

Issue #142, opened 2020-12-24
Loaded tiles seem to reset when turning the camera with loadSiblings === false


gkjohnson:
- Disable `loadSiblings`
- Set `errorThreshold` to `Infinity`
- Load large dataset
- Get camera close to surface
- Let data load
- Turn the camera

gkjohnson:
I suspect it's because of this check which stops traversal and displays the parent when the relevant children aren't loaded yet. When turning the camera a new (unloaded) child comes into view and suddenly the parent doesn't traverse the children because it wants to wait for the child to load. What should the behavior here be?

https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/d462193f7d1a20fd6b7724039e3fa38c9382a92e/src/base/traverseFunctions.js#L226-L247

gkjohnson:
Perhaps we should continue to traverse in this case if any children were visible?

dbuck:
Had you tested this theory out at all? I recently shut off loadSiblings to try and reduce some of the request load, and this blinking behavior started showing up very frequently.

gkjohnson:
> Had you tested this theory out at all?

I haven't revisited this since opening the issue since we're not adjusting the loadSiblings option at the moment. But looking at [Cesiums use of their loadSiblings option](https://github.com/CesiumGS/cesium/blob/dacc86c6a32f1ece62d66ac20b6ea5ff6f040ad0/Source/Scene/Cesium3DTilesetTraversal.js#L479) it looks like it only applies to tiles with the ADD refinement. You can see [here](https://github.com/CesiumGS/cesium/blob/dacc86c6a32f1ece62d66ac20b6ea5ff6f040ad0/Source/Scene/Cesium3DTilesetTraversal.js#L459-L460) that if the refinement is set to REPLACE (and tiles are not being skipped) then children and siblings are loaded regardless of the `loadSiblings` flag. Given that it seems like the flag would be of little benefit unless you have a lot of ADD refinement tiles in your tileset.

> I recently shut off loadSiblings to try and reduce some of the request load

What specifically are you trying to cut down on? Is it load times? Or hiccups when loading tiles? I wonder if there are improvements that could be made to the tile skipping logic that could be made to help.

gkjohnson:
Load siblings has been removed in #667 since the behavior seemed broken and did not match the behavior from Cesium for the same setting, either. #480 outlines possible options for replicating the "skip traversal" behavior from Cesium.

PR #667, opened 2024-08-03
Remove support for loadSiblings, stopAtEmptyTiles options

gkjohnson:
Related to #142
Related to #662 
Related to #663 

This support removes the "loadSiblings" flag and adopts the "true" behavior for "REPLACE" refine and the "false" behavior for "ADD". Now "ADD" will no longer force sibling geometry to load while "REPLACE" will.

The general goal is to ensure that when refine === "REPLACE" that we always have coherent sibling tiles loaded such that there is no "pop" when turning the camera and other tiles haven't been loaded. This is done by ensuring that any children are marked as "used" down to the next tile with content so we can be resilient to content "gaps" in the tile set from parent tiles with no geometry.

This enables the OSM tile set from #662 to load a lot faster but doesn't address the France tile set from https://github.com/iTowns/itowns/issues/2335. It's likely that the sibling behavior isn't correct - ie we could be considering tiles with no geometry as a "valid" sibling rather than traversing to children the next content-ful child (assuming the empty parent isn't an external tile set json). This would render the "loadSiblings" options and refinement difference irrelevant, I think.

OSM buildings marked in red:
image

cc @jailln 

_edit The Google globe tile set includes some intermediate "empty" tiles that are not external tile sets but they all use a geometric error of 1e+100 which may provide some hint as to how this should be handled._

<details>
<summary>Needed Traversal Improvements</summary>

- "REPLACE" parent tiles should only trigger loads for immediate children - or at least only up to calculated error requirements - rather than [loading up to the next set of visible children](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/base/traverseFunctions.js#L58), which may be incredibly broad and deep.
- If an in-frustum REPLACE parent tile has no children in the frustum, then the parent tile should not be considered in the frustum.
- When triggering all "REPLACE" parent children to load we can calculate the SSE even though they may be out of the frustum (see [executeTraversal](https://github.com/CesiumGS/cesium/blob/0a69f67b393ba194eefb7254600811c4b712ddc0/packages/engine/Source/Scene/Cesium3DTilesetBaseTraversal.js#L212-L214) > [updateAndPushChildren](https://github.com/CesiumGS/cesium/blob/0a69f67b393ba194eefb7254600811c4b712ddc0/packages/engine/Source/Scene/Cesium3DTilesetBaseTraversal.js#L99-L101) > [updateTile](https://github.com/CesiumGS/cesium/blob/0a69f67b393ba194eefb7254600811c4b712ddc0/packages/engine/Source/Scene/Cesium3DTilesetTraversal.js#L193) > [updateTileVisibility](https://github.com/CesiumGS/cesium/blob/0a69f67b393ba194eefb7254600811c4b712ddc0/packages/engine/Source/Scene/Cesium3DTilesetTraversal.js#L211) > [Tile.updateVisibility](https://github.com/CesiumGS/cesium/blob/0a69f67b393ba194eefb7254600811c4b712ddc0/packages/engine/Source/Scene/Cesium3DTile.js#L1041)). This information is then used in [executeEmptyTraversal](https://github.com/CesiumGS/cesium/blob/0a69f67b393ba194eefb7254600811c4b712ddc0/packages/engine/Source/Scene/Cesium3DTilesetBaseTraversal.js#L141) to load tiles up to the next valid [error value](https://github.com/CesiumGS/cesium/blob/0a69f67b393ba194eefb7254600811c4b712ddc0/packages/engine/Source/Scene/Cesium3DTilesetBaseTraversal.js#L273).
- Remove check for in-frustum in error calculation

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,14 +352,6 @@ maxDepth = Infinity : Number
 
 The max depth to which tiles will be loaded and rendered. Setting it to `1` will only render the root tile. If the tile at depth `maxDepth` is an empty tile then the next set of visible children will be rendered.
 
-### .loadSiblings
-
-```js
-loadSiblings = true : Boolean
-```
-
-If true then all sibling tiles will be loaded, as well, to ensure coherence when moving the camera. If false then only currently viewed tiles will be loaded.
-
 ### .displayActiveTiles
 
 ```js
--- a/TESTCASES.md
+++ b/TESTCASES.md
@@ -260,52 +260,6 @@ Verify no errors are logged.
 
 Verify that the stats display 125 geometries, 126 textures, and 1 programs.
 
-## Verify stopAtEmptyTiles works
-
-#### steps
-
-1. Open the kitchen sink example by navigating [here](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/).
-1. Set `maxDepth` to 1.
-1. Verify that only 1 tile is visible.
-1. Open the kitchen sink example with the no root content tileset by navigating [here](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/#../data/tileset-no-root-content.json).
-1. Set `maxDepth` to 1.
-1. Verify that all tiles disappear.
-
-#### expected
-
-No tiles are visible when the active leafs have no content.
-
-## Verify maxDepth limit does not stop at empty tiles when stopAtEmptyTiles = false
-
-1. Open the kitchen sink example by navigating [here](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/).
-1. Set `maxDepth` to 1.
-1. Verify that only 1 tile is visible.
-1. Open the kitchen sink example with the no root content tileset by navigating [here](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/#../data/tileset-no-root-content.json).
-1. Disable `stopAtEmptyTiles`.
-1. Set `maxDepth` to 1.
-1. Verify that 4 tiles are visible.
-
-#### expected
-
-The next shallowest tiles are visible past the `maxDepth` cutoff.
-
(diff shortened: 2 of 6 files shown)


Nmzik:
In my opinion, loadSiblings should remain a configurable option for the user.

> The general goal is to ensure that when refine === "REPLACE", we always have coherent sibling tiles **loaded** so that there is no "pop" when turning the camera and other tiles haven't been loaded

 That’s the point; for example, I am limited by the number of tiles **loaded** in 3DTilesRendererJS LRUCache. I don’t care about the pop-in effect when turning the camera and I want to minimize RAM, VRAM, bandwidth usage etc...

gkjohnson:
> In my opinion, loadSiblings should remain a configurable option for the user.

Are you actually using `loadSiblings` for anything? When it's set to `false` the experience feels completely broken with flashing tiles everywhere and gaps in the tile set.

> CesiumJS does provide this option.

This isn't true. Cesium has an option called `loadSiblings` but it defaults to `false` and behaves differently. This confusion is another reason why this option needs to be removed. 

> That’s the point; for example, I am limited by the number of tiles loaded in 3DTilesRendererJS LRUCache. I don’t care about the pop-in effect when turning the camera and I want to minimize RAM, VRAM, bandwidth usage etc...

There are other solutions to these issues, as well. I'd rather discuss the bottlenecks people are facing rather than hold on to a broken setting. If something similar to the current behavior (but fixed in some ways, hopefully) of `loadSiblings` needs to be brought back we can do so with a different name that doesn't conflict with Cesium's naming.

gkjohnson:
Gonna merge this, for now. @Nmzik if you want to make a new issue to discuss any memory or performance issues you may be having I'm happy to talk about alternatives or (if really needed) how to bring something like this flag back once some of this traversal simplification is finished.

PR #668, opened 2024-08-04
Fix 3d tiles traversal logic

gkjohnson:
Based on #667 (merge first)

Refactors some of the traversal functions to have more consistent function signatures and adjusts the traversal logic to account for error when loading REPLACE refinement tile children.

**TODO**
- Check for "ADD" refinement when traversing "REPLACE" parent children
- Test performance compared to master branch
- Test OSM, french tile sets from #667 
- If a tile has no renderable parents then don't trigger loads for all children and wait for them to load


gkjohnson:
This branch is a little slower (0.7ms -> 0.8ms for update) - likely due to always updating the sse metrics. Further optimizations can be made once the load issues are fixed.

PR #680, opened 2024-08-08
feat(i3dm): add RTC_CENTER support

AnthonyGlt:
Add support of RTC_CENTER

Test Files from TOULOUSE METROPOLE:
https://github.com/AnthonyGlt/3DTilesRendererJS/tree/i3dm/example/arbres_rieux_cmpt_sample

image

### DIFF
--- a/src/three/loaders/I3DMLoader.js
+++ b/src/three/loaders/I3DMLoader.js
@@ -76,6 +76,7 @@ export class I3DMLoader extends I3DMLoaderBase {
 						const NORMAL_RIGHT = featureTable.getData( 'NORMAL_RIGHT', INSTANCES_LENGTH, 'FLOAT', 'VEC3' );
 						const SCALE_NON_UNIFORM = featureTable.getData( 'SCALE_NON_UNIFORM', INSTANCES_LENGTH, 'FLOAT', 'VEC3' );
 						const SCALE = featureTable.getData( 'SCALE', INSTANCES_LENGTH, 'FLOAT', 'SCALAR' );
+						const RTC_CENTER = featureTable.getData( 'RTC_CENTER' );
 
 						[
 							'QUANTIZED_VOLUME_OFFSET',
@@ -93,7 +94,6 @@ export class I3DMLoader extends I3DMLoaderBase {
 							}
 
 						} );
-						const rtcCenter = featureTable.getData( 'RTC_CENTER' );
 
 						// get the average vector center so we can avoid floating point error due to lower
 						// precision transformation calculations on the GPU
@@ -110,13 +110,6 @@ export class I3DMLoader extends I3DMLoaderBase {
 						const instances = [];
 						const meshes = [];
 						model.scene.updateMatrixWorld();
-						if ( rtcCenter ) {
-
-							model.scene.position.x += rtcCenter[ 0 ];
-							model.scene.position.y += rtcCenter[ 1 ];
-							model.scene.position.z += rtcCenter[ 2 ];
-
-						}
 
 						model.scene.traverse( child => {
 
@@ -127,6 +120,15 @@ export class I3DMLoader extends I3DMLoaderBase {
 								const { geometry, material } = child;
 								const instancedMesh = new InstancedMesh( geometry, material, INSTANCES_LENGTH );
 								instancedMesh.position.copy( averageVector );
+
+								if ( RTC_CENTER ) {
+
+									instancedMesh.position.x += RTC_CENTER[ 0 ];
+									instancedMesh.position.y += RTC_CENTER[ 1 ];
+									instancedMesh.position.z += RTC_CENTER[ 2 ];
+
+								}
+
 								instances.push( instancedMesh );
 
 							}
--- a/src/three/loaders/I3DMLoader.js
+++ b/src/three/loaders/I3DMLoader.js
@@ -168,11 +168,9 @@ export class I3DMLoader extends I3DMLoaderBase {
 							}
 
 							// scale
-							if ( SCALE ) {
+							tempSca.set( 1, 1, 1 );
 
-								tempSca.setScalar( SCALE[ i ] );
-
-							} else if ( SCALE_NON_UNIFORM ) {
+							if ( SCALE_NON_UNIFORM ) {
 
 								tempSca.set(
 									SCALE_NON_UNIFORM[ i * 3 + 0 ],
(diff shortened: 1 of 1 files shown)


src/three/loaders/I3DMLoader.js
```diff
@@ -110,6 +110,14 @@ export class I3DMLoader extends I3DMLoaderBase {
 						const instances = [];
 						const meshes = [];
 						model.scene.updateMatrixWorld();
+						if ( rtcCenter ) {
+
+							model.scene.position.x += rtcCenter[ 0 ];
+							model.scene.position.y += rtcCenter[ 1 ];
+							model.scene.position.z += rtcCenter[ 2 ];
+
+						}
```
gkjohnson, line 119: Lets add this to each `InstancedMesh` in the traverse loop below, instead, so the scene object is set to the origin as every other model result is set to:

```js
instancedMesh.position.copy( averageVector );

if ( RTC_CENTER ) {

  instancedMesh.position.x += RTC_CENTER[ 0 ];
  // ...

}
```

src/three/loaders/I3DMLoader.js
```diff
@@ -94,6 +93,7 @@ export class I3DMLoader extends I3DMLoaderBase {
 							}
 
 						} );
+						const rtcCenter = featureTable.getData( 'RTC_CENTER' );
```
gkjohnson, line 96: Lets define this as `RTC_CENTER` as the rest of the feature table values are and move it up above with the rest of the definitions to be consistent.

gkjohnson (commented):
A couple small comments but otherwise this looks good!

gkjohnson:
Excellent! Thank you!

gkjohnson:
Regarding the EAST_NORTH_UP extension - it looks like it's applied [here](https://github.com/CesiumGS/cesium/blob/be2aad4d443c276978e2b381011b31b8bfd34af9/packages/engine/Source/Scene/Model/I3dmLoader.js#L803-L809) in Cesium.

The [Ellipsoid.constructLatLonFrame](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/93d811699244ffc04450a9607a7f429dd55c2540/src/three/math/Ellipsoid.js#L51-L63) (perhaps can be renamed to `getEastNorthUpFrame`) can be used construct an east-north-up frame from the derived latitude and longitude of an instance.

AnthonyGlt:
> Regarding the EAST_NORTH_UP extension - it looks like it's applied [here](https://github.com/CesiumGS/cesium/blob/be2aad4d443c276978e2b381011b31b8bfd34af9/packages/engine/Source/Scene/Model/I3dmLoader.js#L803-L809) in Cesium.
> 
> The [Ellipsoid.constructLatLonFrame](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/93d811699244ffc04450a9607a7f429dd55c2540/src/three/math/Ellipsoid.js#L51-L63) (perhaps can be renamed to `getEastNorthUpFrame`) can be used construct an east-north-up frame from the derived latitude and longitude of an instance.

Should it be something like this ?

```js

if ( NORMAL_UP ) {

		....

	} else if (EAST_NORTH_UP) {
		let pos = {};
		pos = WGS84_ELLIPSOID.getPositionToCartographic( tempPos, pos ) ;

		rotMat = WGS84_ELLIPSOID.getEastNorthUpFrame( pos.lat, pos.lon, tempMat )

		tempQuat.setFromRotationMatrix( tempMat );
	}
	if(!(NORMAL_UP || EAST_NORTH_UP)){

		tempQuat.set( 0, 0, 0, 1 );

	}


```

gkjohnson:
From looking at the code it's not completely clear to me how it's supposed to work so I'll have to leave it to some investigation. The section of code I linked and the docs seem to indicate that the NORMAL_UP, NORMAL_UP_OCT32P, and EAST_NORTH_UP are mutually exclusive but in your example geometry includes both NORMAL_UP and EAST_NORTH_UP semantics so the ENU logic shouldn't be applying. It's not exactly clear to me what's happening in Cesium in this case. I assume this means that EAST_NORTH_UP isn't having any effect?

But either way I do expect that the east north up frame needs to be calculated based on the instances final position - which happens [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/f7a7ca4a30b2c7068bec50dc67ac7a4c94da1022/src/three/loaders/I3DMLoader.js#L204-L207).

AnthonyGlt:
That's what I've noticed. In Cesium, it's mutually exclusive (`else if`):
https://github.com/CesiumGS/cesium/blob/009a66d04221f74f3f957fa7d79fe0c1b393997c/packages/engine/Source/Scene/Model/I3dmLoader.js#L803

I think my data are wrong, by following what's done in cesium, the ENU shouldn't be apply because there is already a NORMAL_UP...

AnthonyGlt:
> But either way I do expect that the east north up frame needs to be calculated based on the instances final position - which happens [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/f7a7ca4a30b2c7068bec50dc67ac7a4c94da1022/src/three/loaders/I3DMLoader.js#L204-L207).

Does it mean I should calculate the matrix for every instance like this:

```js
		//THIS COMMENTED
	// compose the instance matrix
	// tempMat.compose( tempPos, tempQuat, tempSca ).multiply( adjustmentTransform );

	// multiple in the original meshes world transform
	for ( let j = 0, l = instances.length; j < l; j ++ ) {

		const instance = instances[ j ];
		
		//EDITED HERE
		if (EAST_NORTH_UP) {
			let pos = {};
			let inst_pos = instance.position;
			pos = WGS84_ELLIPSOID.getPositionToCartographic( inst_pos, pos ) ;
			let rotMat = new Matrix4();

			rotMat = WGS84_ELLIPSOID.getEastNorthUpFrame( pos.lat, pos.lon, rotMat )

			tempQuat.setFromRotationMatrix( rotMat );

		}
		tempMat.compose( tempPos, tempQuat, tempSca ).multiply( adjustmentTransform );
		
		//END EDITED

		const mesh = meshes[ j ];
		tempMat2.multiplyMatrices( tempMat, mesh.matrixWorld );
		instance.setMatrixAt( i, tempMat2 );


	}



```

Looks like its working but is there a better way to do it ?
By the way I'm using this example 
https://raw.githubusercontent.com/CesiumGS/3d-tiles-samples/main/1.0/TilesetWithTreeBillboards/tileset.json
Before:
Screenshot from 2024-09-06 16-58-47

After:
Screenshot from 2024-09-06 16-58-19

gkjohnson:
The position you're using isn't taking into account the world transform since it's happening before this line:

> ```js
> tempMat2.multiplyMatrices( tempMat, mesh.matrixWorld );
> ```

I expect you'll want to extract the position from this transform and then calculate the orientation. This transform could be one (and likely is) one that offsets it to the surface of the globe.

PR #683, opened 2024-08-09
LRUCache: unload unused tiles even when not calling update continuously

### LINKED ISSUE #673: LRUCache does not continue to unload unless "update" is called
This is an artifact of the design intending to support multiple tilesets calling "scheduleUnload" and implicitly marking all tiles as unused after that. Ideally once the unload step is complete the LRU cache would automatically schedule the next frame for unload until there are no tiles to unload.

gkjohnson:
Fix #673


gkjohnson:
This will cause issues if calling "update" only on one tile set of two or so when sharing an LRUCache.

PR #687, opened 2024-08-14
ImplicitTilesPlugin: Remove three.js dependency, simplify box bounds update

gkjohnson:
Related to #608 

Removes the dependency on three.js by computing the box bounds without matrix transformations. Also updates the `getTileBoundingVolume` logic to support output multiple bounds types of present.

At some point we can probably move this to the "base" folder since three.js is no longer required.

cc @AnthonyGlt

### DIFF
--- a/src/three/plugins/SUBTREELoader.js
+++ b/src/three/plugins/SUBTREELoader.js
@@ -703,22 +703,25 @@ export class SUBTREELoader extends LoaderBase {
 			const cellSteps = 2 ** tile.__level - 1;
 			const scale = Math.pow( 2, - tile.__level );
 
-			// iterate over the three obb axes
-			for ( let i = 0; i < 3; i ++ ) {
+			// Iterate over the three obb axes. Skip the z axis since octree
+			// implicit bounds are not supported.
+			for ( let i = 0; i < 2; i ++ ) {
 
-				// multiply x and y axes by the scale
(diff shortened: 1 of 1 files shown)


src/three/plugins/SUBTREELoader.js
```diff
@@ -659,87 +658,74 @@ export class SUBTREELoader extends LoaderBase {
...
-				scaleFactors.z = scale;
+			// 0-2: center of the box
+			// 3-5: x axis direction and half length
+			// 6-8: y axis direction and half length
+			// 9-11: z axis direction and half length
+
+			const box = [ ...this.rootTile.boundingVolume.box ];
+			const cellSteps = 2 ** tile.__level - 1;
+			const scale = Math.pow( 2, - tile.__level );
+			for ( let i = 0; i < 3; i ++ ) {
```
AnthonyGlt, line 705: Should it not  stop at 2 and leave Z untouched ?
```suggestion
			for ( let i = 0; i < 2; i ++ ) {
```

AnthonyGlt (commented):
Thank you for these improvements

src/three/plugins/SUBTREELoader.js
gkjohnson: Thanks for the catch - I was ignoring the z component of each axis rather than the z-axis of the bounding box.

AnthonyGlt (approved):

PR #689, opened 2024-08-14
Implict Tiling: Remove unneeded tile members

gkjohnson:
Related to #608 

- Removes the need for `__subtreeDivider` by calculating it as-needed.
- Removes `__subtreeUri` since it's already present on the object.
- Remove unnecessary assignment of `__implicitRoot`.

cc @AnthonyGlt


AnthonyGlt (approved):
:+1:

PR #709, opened 2024-08-21
Fix `TileBoundingVolume.intersectsRay` (typo)

pmi-gbs:
I think the wrong function was called? This lead to an exception for me when raycasting the scene.


gkjohnson:
Good catch - thank you!

PR #712, opened 2024-08-22
Rename threejs `REVISION_165` to `REVISION_LESS/GEQ_165` to clearify its usage

Philipp-M:
I think `REVISION` definition is wrong currently, preventing a raycast optimization?
(Also renamed the variable to be more clear that it's greater equals that version)

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -20,7 +20,7 @@ import { ExtendedFrustum } from './math/ExtendedFrustum.js';
 import { estimateBytesUsed } from './utilities.js';
 
 // In three.js r165 and higher raycast traversal can be ended early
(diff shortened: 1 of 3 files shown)


gkjohnson:
Hello! Thanks for the contribution - the only thing that should be wrong / misleading here is the name of the variables. All of the logic and conditions are correct. So if you change the operation from `<` to `>=` you'll have to invert the usage of the variables, as well.

Philipp-M:
Oh well, I guess I wasn't thorough enough, I think I was confused by the comment:

```
// In three.js r165 and higher raycast traversal can be ended early
```
and the same name (`REVISION_165`) in the other file. Ok for this PR to have a sense at all, I just made the naming clearer...

gkjohnson:
Great, thanks!

PR #729, opened 2024-08-29
Add rudimentary JsDoc for `EnvironmentControls`

pmi-gbs:
This adds a few JsDoc definitions for better editor support and potentially typescript definitions.
Speaking of, I noticed, that the typescript definitions are quite out of sync.
E.g. I only have seen the `EnvironmentControls` as I went over all the examples.
When using the typescript language server (in typescript), it tells me, that there aren't even the `EnvironmentControls` to import (as that isn't defined in `index.d.ts`).

So I wonder whether it would make sense to generally use JsDoc instead of these typescript definitions and maybe generate the `.d.ts` bindings as e.g. described [here](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html), so that all of this would be more in sync while also better documenting the js files.

### DIFF
--- a/src/three/controls/EnvironmentControls.js
+++ b/src/three/controls/EnvironmentControls.js
@@ -8,6 +8,9 @@ import {
 	EventDispatcher,
 	MathUtils,
 	Clock,
+	PerspectiveCamera,
+	OrthographicCamera,
+	Object3D,
 } from 'three';
 import { PivotPointMesh } from './PivotPointMesh.js';
 import { PointerTracker } from './PointerTracker.js';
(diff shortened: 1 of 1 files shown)


gkjohnson:
Hey! Thanks for adding this. JSDoc isn't really used in this project - I think adding a d.ts file for Environment controls would be better and align with the rest of the files.

> So I wonder whether it would make sense to generally use JsDoc instead of these typescript definitions and maybe generate the .d.ts bindings as e.g. described [here](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html),

JSDoc and typescript bindings generally very easily get out of sync and both require a maintenance burden. I've chosen to use d.ts files in the repo to provide types for other project for the public APIs but generally I need to rely on contributions to fix up or add files if there are any mistakes.

pmi-gbs:
> JSDoc and typescript bindings generally very easily get out of sync and both require a maintenance burden.

That's the reason why I suggested JSDoc, as IME typescript bindings run into this issue even more, while I think JSDoc has the advantage, of providing better editor/IDE support also inside this project (than pure js without doc-comments) and directly serves as (inline) API documentation.

But as said, it was just a suggestion. I guess I'll close this PR then, when `d.ts` files are the preferred way.

gkjohnson:
I've used JSDoc before and find it much more verbose to document and can also become out of date. I understand and see the value but adding new overhead using processes to generate something like docs and typescript definitions from JSDoc comments isn't something I have the bandwidth to maintain with the projects I work on. If someone would like to commit to figuring that out and maintaining that process then perhaps there's more possibility. Otherwise I have to stick with processes I know and work for me.

Generally the typescript definitions should be fairly up to date but when something is added to the project, as GlobeControls and EnvironmentControls have been, sometimes the d.ts definitions are missed. I appreciate it when people can make contributions when they're found so if you'd like to add definition files for the new controls I'd be happy to get that PR merged.

Issue #657, opened 2024-07-25
Deprecation of `BatchTable#getData`


Desplandis:
**Describe the bug**

The latest release and PR #627 deprecates the method `getData` of `BatchTable`.
We could be however want to retrieve a whole named buffer. Some of the use-cases includes the fact that we could want to generate a specific VBO (e.g. `color`) from an entry of the batch table. This is our case with `py3dtiles` which generates 3D Tiles point clouds from LAS and encode the `classification` property in the batch table, we then use this data to color our point cloud.

gkjohnson:
We can add it back if it's needed but I think the name should be changed to better reflect what's being retrieved - like `getBuffer` or `getPropertyBuffer` or something.

woondossier:
This will also cause issues for the Dutch 3D BAG implementation, the batchtable includes an attributes entry with info about the building based on batch id index so we get that by using .getData("attributes") and then use that data to display building information.

So basically without this function, we can't update anymore because it breaks existing functionality. So if it could be brought back under a different name it would be greatly appreciated.

The company that creates these datasets also has an [viewer](https://3dbag.nl/en/viewer) that shows this as well 
[Sourcecode](https://github.com/3DBAG/3dbag-viewer/blob/2317b9a3710e7dce692b0af825b8f5ee35ca814d/src/components/ThreeViewer.vue#L904) showing the function

gkjohnson:
> So basically without this function, we can't update anymore because it breaks existing functionality.

Can you explain why the newly added `getDataFromId` in #627 does not work for your use case? It provides all attributes associated with an id:

```js
const attributes = JSON.parse( batchTable. getDataFromId( batch_id ).attributes );
```

woondossier:
Ah yes, that specific use case would work, and I actually agree it's an improvement. 

The main issue that I'm running into is that we allow people to "delete" buildings from the tiles for urban planning purposes and when a tile is loaded we loop through the attributes and check if a identifier is in the list of deleted buildings and "remove" it from the tile based on the index. 

This is done in reverse order because there's no ID yet as these change with every release of a new dataset, so we store the unique identifier instead, so I suppose the real issue is we do attributes array -> id instead of id -> attributes in this particular case.

So if there is a way of getting all the IDs of a BatchTable, it would also solve the issue I'm having right now but I couldn't find it.

gkjohnson:
You can still iterate over all ids in the BatchTable with this function and check the deletion state:

```js
[...]

PR #731, opened 2024-09-04
add BatchTable.getPropertyArray

jailln:
Add `BatchTable.getPropertyArray` to get an array of data for a given property key.

See #657.

### DIFF
--- a/README.md
+++ b/README.md
@@ -675,19 +675,22 @@ getKeys() : Array<String>
 
 Returns the keys of all the data in the batch table.
 
-### .getData
+### .getDataFromId
 
 ```js
-getData(
-	key : String,
-	defaultComponentType = null : String | null,
-	defaultType = null : String | null,
-) : Array | TypedArray | null
+getDataFromId( id: Number, target?: Object ) : Object;
 ```
 
-Returns the data associated with the `key` passed into the function. If the component and type are specified in the batch table contents then those values are used otherwise the values in `defaultComponentType` and `defaultType` are used. Returns null if the key is not in the table.
+Returns an object definition for all properties of the batch table and its extensions for a given `id`.
+A `target` object can be specified to store the result. Throws an error if the id is out of the batch table bounds.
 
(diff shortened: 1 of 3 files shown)


src/utilities/BatchTable.js
```diff
@@ -71,4 +72,11 @@ export class BatchTable extends FeatureTable {
 
 	}
 
+	getPropertyArray( key, componentType = null, type = null ) {
```
gkjohnson, line 75: Do we need the `componentType` and `type` fields?

jailln: I kept them since they were in `getData` but I was wondering the same thing. Personally I don't see any use cases. I will remove them.

README.md
```diff
@@ -675,19 +675,27 @@ getKeys() : Array<String>
...
 ```js
-getData(
-	key : String,
-	defaultComponentType = null : String | null,
-	defaultType = null : String | null,
-) : Array | TypedArray | null
+	getDataFromId(
+		id: Number,
+		target?: Object
+	) : Object;
```
gkjohnson, line 684: nit: indentation and can we collapse this onto one line since it's not all that long any more

README.md
```diff
@@ -675,19 +675,27 @@ getKeys() : Array<String>
...
-Returns the data associated with the `key` passed into the function. If the component and type are specified in the batch table contents then those values are used otherwise the values in `defaultComponentType` and `defaultType` are used. Returns null if the key is not in the table.
+Returns an object definition for all properties of the batch table and its extensions for a given `id`.
+A `target` object can be specified to store the result. Throws an error if the id is out of the batch table bounds.
+
+### .getPropertyArray
+
+```js
+getPropertyArray(
+	key : String
+) : Array | TypedArray | null
```
gkjohnson, line 695: nit: can we collapse this onto one line since it's not all that long any more

gkjohnson (commented):
Last couple comments - otherwise looks great. Thanks!

gkjohnson:
Great thanks!

PR #734, opened 2024-09-06
fix: load i3dm with external gtlf

AnthonyGlt:
Load correctly the i3dm files when the gltf is external and doesn't share the same base url

This adaptation is made to comply with this part of the [documentation](https://portal.ogc.org/files/102132#tileformats-instanced3dmodel-instanced-3d-model):
>When the glTF field contains a URI, then this URI may point to a [relative external reference (RFC3986)](https://tools.ietf.org/html/rfc3986#section-4.2). When the URI is relative, its base is always relative to the referring .i3dm file. Client implementations are required to support relative external references. **Optionally, client implementations may support other schemes (such as http://). All URIs shall be valid and resolvable.**

We should be able to load i3dm containing an URL starting by "http"

### DIFF
--- a/src/base/loaders/I3DMLoaderBase.js
+++ b/src/base/loaders/I3DMLoaderBase.js
@@ -77,6 +77,7 @@ export class I3DMLoaderBase extends LoaderBase {
 
 		let glbBytes = null;
 		let promise = null;
+		let gltfWorkingPath = null;
 		if ( gltfFormat ) {
 
 			glbBytes = bodyBytes;
@@ -84,13 +85,19 @@ export class I3DMLoaderBase extends LoaderBase {
 
 		} else {
 
-			this.workingPath = this.resolveExternalURL( arrayToString( bodyBytes ) );
-			promise = fetch( this.workingPath, this.fetchOptions )
+			const externalUri = this.resolveExternalURL( arrayToString( bodyBytes ) );
+
+			//Store the gltf working path
+			const uriSplits = externalUri.split( /[\\/]/g );
+			uriSplits.pop();
+			gltfWorkingPath = uriSplits.join( '/' );
+
+			promise = fetch( externalUri, this.fetchOptions )
 				.then( res => {
 
 					if ( ! res.ok ) {
 
-						throw new Error( `I3DMLoaderBase : Failed to load file "${ this.workingPath }" with status ${ res.status } : ${ res.statusText }` );
+						throw new Error( `I3DMLoaderBase : Failed to load file "${ externalUri }" with status ${ res.status } : ${ res.statusText }` );
 
 					}
 
@@ -112,7 +119,7 @@ export class I3DMLoaderBase extends LoaderBase {
 				featureTable,
 				batchTable,
 				glbBytes,
-
+				gltfWorkingPath
 			};
 
 		} );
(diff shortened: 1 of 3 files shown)


src/base/loaders/I3DMLoaderBase.js
```diff
@@ -84,13 +84,13 @@ export class I3DMLoaderBase extends LoaderBase {
 
 		} else {
 
-			const externalUri = this.resolveExternalURL( arrayToString( bodyBytes ) );
-			promise = fetch( externalUri, this.fetchOptions )
+			this.workingPath = this.resolveExternalURL( arrayToString( bodyBytes ) );
+			promise = fetch( this.workingPath, this.fetchOptions )
```
gkjohnson, line 88: Working path shouldn't be set to the file URL - it should be set to the path root. Ie we should remove the file name from the path if we're going to call it a working path. Then the GLTFLoader can use it directly.

src/base/loaders/I3DMLoaderBase.js
```diff
@@ -112,6 +112,7 @@ export class I3DMLoaderBase extends LoaderBase {
 				featureTable,
 				batchTable,
 				glbBytes,
+
```
gkjohnson, line 115: nit: unneeded white space addition.

src/base/loaders/I3DMLoaderBase.js
```diff
@@ -84,13 +84,13 @@ export class I3DMLoaderBase extends LoaderBase {
 
 		} else {
 
-			const externalUri = this.resolveExternalURL( arrayToString( bodyBytes ) );
-			promise = fetch( externalUri, this.fetchOptions )
+			this.workingPath = this.resolveExternalURL( arrayToString( bodyBytes ) );
```
gkjohnson, line 87: I'd prefer not to overwrite the existing working path for the I3DMLoader since this is intended to be the working path for the I3DM path being loaded, not the GLTF. I think we can pass the GLTF working path in via the `parse` constructor.

src/base/loaders/I3DMLoaderBase.js
AnthonyGlt: If I'm following this 
https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/734#discussion_r1749055695

Should I just revert this part and come back to `externalUri` ?

src/base/loaders/I3DMLoaderBase.js
AnthonyGlt: I have difficulties to follow, do you mean this `parse` ?
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/564fb9faa096292dec3b7d2ca00dceb75e7af9b1/src/three/loaders/I3DMLoader.js#L73

Should I return it  here ? 
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/564fb9faa096292dec3b7d2ca00dceb75e7af9b1/src/base/loaders/I3DMLoaderBase.js#L110-L116

src/base/loaders/I3DMLoaderBase.js
gkjohnson: Yeah - hopefully the answer in https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/734/files#r1751888696 clarifies things. Let me know if I need to rephrase it.

gkjohnson:
Thank you!

PR #744, opened 2024-09-14
Auth Plugins: automatically refresh the token

### LINKED ISSUE #732: GoogleCloudAuthPlugin, CesiumIonAuthPlugin: Optionally auto refresh token when it expires
After a certain amount of time (hours) the Google Cloud and Cesium tokens expire and a new one must be requested once it fails. We need to understand how to procedurally determine if the token has expired.

From [Cesium Ion](https://cesium.com/learn/ion/rest-api/#operation/getAssetEndpoint) API:

> This token is different than the access token used throughout the rest of the REST API. It provides access to the asset for approximately one hour. To continue using the asset after the allotted time, **simple re-request the endpoint for a new token**.

For [Google Photorealistic Tiles](https://developers.google.com/maps/documentation/tile/3d-tiles#getting_3d_tiles):

> The render can make at least three hours of tile requests from a single root tileset request. After reaching this limit, **you must make another root tileset request**.

Behavior should be dis-ableable with a "autoRefreshToken" setting.

gkjohnson:
Fix #732 

**TODO**
- Find way to verify token refresh behavior

### DIFF
--- a/src/three/plugins/CesiumIonAuthPlugin.js
+++ b/src/three/plugins/CesiumIonAuthPlugin.js
@@ -31,6 +31,7 @@ export class CesiumIonAuthPlugin {
 
 	loadRootTileSet( rootUrl ) {
 
+		// TODO: see if we can get rid of this "load root tile set" function
 		if ( this._tokenState === UNLOADED ) {
 
 			this._tokenState = LOADING;
--- a/src/three/plugins/CesiumIonAuthPlugin.js
+++ b/src/three/plugins/CesiumIonAuthPlugin.js
@@ -16,7 +16,7 @@ export class CesiumIonAuthPlugin {
 		this.tiles = null;
 		this._tileSetVersion = - 1;
 		this._tokenState = UNLOADED;
-		this._loadPromise = null;
+		this._rootEndpointPromise = null;
 		this._tokenRefreshPromise = null;
 		this._endpointURL = null;
 		this._deferredToPlugin = false;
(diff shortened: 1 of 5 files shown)


src/three/plugins/GoogleCloudAuthPlugin.js
```diff
@@ -63,4 +77,38 @@ export class GoogleCloudAuthPlugin {
 
 	}
 
+	async fetchData( uri, options ) {
+
+		// wait for the token to refresh if loading
+		if ( this._tokenRefreshPromise !== null ) {
+
+			await this._tokenRefreshPromise;
```
gkjohnson, line 85: The session token will likely be modified by this point so we'll need to reassign it like below

src/three/plugins/CesiumIonAuthPlugin.js
```diff
@@ -1,122 +1,141 @@
...
 
-					} else {
+			return Promise.resolve().then( async () => {
 
-						return Promise.reject( new Error( `${ res.status } : ${ res.statusText }` ) );
+				// wait for the token to refresh if loading
+				if ( this._tokenRefreshPromise !== null ) {
 
-					}
+					await this._tokenRefreshPromise;
```
gkjohnson, line 68: See https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/744/files?w=1#r1768600939

gkjohnson:
> - Ensure loading will not cycle trying for new tokens if a refresh fails
> - Find way to verify token refresh behavior

Wait 1 / 4 hours respectively to ensure the token is refreshed as expected with a log. Likewise the looping behavior can be checked by setting a global flag to trigger the appropriate code path.

PR #750, opened 2024-09-17
Fix visible pivot mesh and intertia in `EnvironmentControls` when controls are disabled

pmi-gbs:
When clicking/dragging while the controls are disabled, the pivot mesh is still visible, this should fix that issue.
The `drag`/`rotationInertia` is also reset when disabling the controls, to avoid a "fling" when enabling the controls again.
(+ removal of `multiplyScalar` which doesn't seem to be necessary)


gkjohnson:
Great, thank you!

Issue #738, opened 2024-09-12
Testing maxByteSize does not work


zysyz:
I first tested the TileCompressionPlugin plugin and found that there was indeed a loss in model quality, but the memory usage did not decrease.
Then I tested maxByteSize, and no matter how much the value was adjusted, I did not find any changes in the memory usage. May I ask if my code was written incorrectly or if there are any other issues.Here is a partial code of mine:

```
import { TileCompressionPlugin } from './ TileCompressionPlugin.js'
const tilesRenderer = new TilesRenderer( './ model/tileset.json' ); 
……
tilesRenderer.registerPlugin( new TileCompressionPlugin() );

tilesRenderer.fetchOptions.mode = 'cors';
tilesRenderer.manager.addHandler(  /\. (gltf|glb)$/g, loader );
tilesRenderer.setCamera( camera );
tilesRenderer.setResolutionFromRenderer( camera, renderer );

tilesRenderer.errorTarget = 35;
tilesRenderer.lruCache.maxByteSize = 0.3 * 2**30;
tilesRenderer.lruCache.minByteSize = 0.2 * 2**30;
```

gkjohnson:
> there was indeed a loss in model quality

Can you provide some screenshots showing the loss in model quality? It's not unexpected and it will impact different models with different severity but it's good to get a sense of how affected different tile sets are.

> but the memory usage did not decrease.

How are you measuring this? The GPU memory will be what's affected impacted due to removing mip maps. Measuring memory usage differences using any CPU profiling tools will likely not show much difference.

zysyz:
> 你能提供一些屏幕截图显示模型质量的损失吗？这并不意外，它会对不同的模型产生不同的严重程度的影响，但最好了解不同的瓷砖集受到的影响程度。

The following is a comparison of the border of a billboard before and after using tilesRendering. registrant Plugin (new TileCompression Plugin ()).
![17d03ccc-2795-4232-b3e4-6e2a5d6b626e](https://github.com/user-attachments/assets/aadff78e-24a1-4fd6-9e5e-d5a0c27650d5)
![a33786b4-b20b-4553-aff3-5a377511908f](https://github.com/user-attachments/assets/886897fd-4657-443f-967a-3c344f889c05)



> How are you measuring this? The GPU memory will be what's affected impacted due to removing mip maps. Measuring memory usage differences using any CPU profiling tools will likely not show much difference.

I checked the GPU memory usage in the task manager.

[...]

PR #753, opened 2024-09-18
Update LRUCache doc and question issue template

jailln:
See https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/738#issuecomment-2356715561


gkjohnson:
Great, thank you!

PR #755, opened 2024-09-20
feat(i3dm): add EAST_NORTH_UP support

AnthonyGlt:
Adding the EAST_NORTH_UP support as discussed here https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/680

### DIFF
--- a/src/three/loaders/I3DMLoader.js
+++ b/src/three/loaders/I3DMLoader.js
@@ -1,6 +1,7 @@
 import { I3DMLoaderBase } from '../../base/loaders/I3DMLoaderBase.js';
 import { DefaultLoadingManager, Matrix4, InstancedMesh, Vector3, Quaternion } from 'three';
 import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader.js';
+import {WGS84_ELLIPSOID} from "../math/GeoConstants.js";
 
 const tempFwd = new Vector3();
 const tempUp = new Vector3();
@@ -77,6 +78,8 @@ export class I3DMLoader extends I3DMLoaderBase {
 						const SCALE_NON_UNIFORM = featureTable.getData( 'SCALE_NON_UNIFORM', INSTANCES_LENGTH, 'FLOAT', 'VEC3' );
 						const SCALE = featureTable.getData( 'SCALE', INSTANCES_LENGTH, 'FLOAT', 'SCALAR' );
 						const RTC_CENTER = featureTable.getData( 'RTC_CENTER' );
+						const EAST_NORTH_UP = featureTable.getData( 'EAST_NORTH_UP' );
+						let modifiedRot = false;
 
 						[
 							'QUANTIZED_VOLUME_OFFSET',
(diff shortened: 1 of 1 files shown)


src/three/loaders/I3DMLoader.js
```diff
@@ -194,18 +207,36 @@ export class I3DMLoader extends I3DMLoaderBase {
...
 
 							// multiple in the original meshes world transform
 							for ( let j = 0, l = instances.length; j < l; j ++ ) {
 
 								const instance = instances[ j ];
+
+								//EDITED HERE
+								if (EAST_NORTH_UP) {
+									let pos = {};
+									let inst_pos = instance.position;
```
gkjohnson, line 222: This instance pos doesn't account for the translation from `mesh.matrixWorld` or the translation already in `tempMat`. You'll probably need to extract the position from `tempMat2` below or multiply both of those translations into this vector before computing the orientation.

AnthonyGlt: I've tried by using:
`tempMat2.decompose( inst_pos, tempQuat2, tempSca2 )` or `inst_pos.setFromMatrixPosition( tempMat2 );`
Then how do I handle the resulting `rotMat` ? :grimacing: 
Directly `tempMat2.multiply(rotMat);` or `tempMat2.compose(...)` neither worked. I must be missing something
Thank you

gkjohnson: The rotation calculated will be in the "global" frame so it will need to be transform into the local frame of the InstancedMesh object before being assigned to the instance itself. The most straight forward way to do this is probably to work with rotation matrices to transform into local space rather than the quaternions.

Hopefully that makes sense. There's a lot of matrix transformation going on here and it can be hard to keep track of. It might be worth drawing out the hierarchy of transformations being applied here so it's easier to visualize.

Also, how are you testing this?

AnthonyGlt: Ok thank you,
I'm testing it using this [sample data](https://raw.githubusercontent.com/CesiumGS/3d-tiles-samples/main/1.0/TilesetWithTreeBillboards/tileset.json) if it's what you meant

AnthonyGlt: @gkjohnson I didn't succeed to make it work correctly. Would it be possible for you to check it out please ? Looks like not that much is left to do.
I'm still using this [sample data](https://raw.githubusercontent.com/CesiumGS/3d-tiles-samples/main/1.0/TilesetWithTreeBillboards/tileset.json) 
There is 2 models charging. To switch from one to another, I have to decrease the `errorTarget` value (using the slider).
The 2 models should be perpendicular to the biggest surface of the Bounding Region, like:

Screenshot from 2024-10-10 10-45-31


image

gkjohnson: > The 2 models should be perpendicular to the biggest surface of the Bounding Region, like:

Do you have an image of what it looks like now for you? I can take a look when I have time.

AnthonyGlt: Yes, with the code from this bra,ch, I'm obtaining the right orientation. The thing is, I'm not able to do this:

> This instance pos doesn't account for the translation from `mesh.matrixWorld` or the translation already in `tempMat` since it's just the local translation. You'll probably need to extract the position from `tempMat2` below or multiply both of those translations into this vector before computing the orientation.

Without losing the orientation.

Thank you for checking it out :+1:

gkjohnson: Just made some changes that I _think_ should fix it. It would be good to try more test data if it's available, though. Let me know what you think -

AnthonyGlt: Looks perfect to me :+1:  Thank you !

PR #762, opened 2024-09-25
Fix memory cache

### LINKED ISSUE #761: [REGRESSION] Implicit Tiling Repeating request
**Description**

Not exactly the same as https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/728
I've noticed that after moving the camera around, we start doing the same requests again and again, infinitely.

Could be link to the latest changes concerning the cache management, WDYT ?

**To Reproduce**

1. Load the tileset from https://github.com/AnthonyGlt/data/tree/main/arbres_rieux_glb
2. Rotate / Move the camera around
3. Check the network request

[Screencast from 23-09-2024 17:59:56.webm](https://github.com/user-attachments/assets/adc05089-b3fc-403f-937c-5fd184d9e2c3)

gkjohnson:
Fix #761

### DIFF
--- a/src/utilities/LRUCache.js
+++ b/src/utilities/LRUCache.js
@@ -297,7 +297,8 @@ class LRUCache {
 				const bytes = bytesMap.get( item ) || 0;
 				if (
 					usedSet.has( item ) && loadedSet.has( item ) ||
-					this.cachedBytes - removedBytes - bytes < maxBytesSize
+					this.cachedBytes - removedBytes - bytes < maxBytesSize &&
+					itemList.length - removedNodes <= maxSize
 				) {
 
 					break;
@@ -309,7 +310,8 @@ class LRUCache {
 
 			}
 
-			// evict up to the min size
+			// evict up to the min node or bytes size, keeping one more item over the min bytes limit
+			// so we're meeting it
 			while (
 				removedBytes < bytesToUnload ||
 				removedNodes < nodesToUnload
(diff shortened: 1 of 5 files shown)


gkjohnson:
cc @AnthonyGlt - this should be ready if you want to test it a bit. I just have to write some more tests / fix the existing ones.

In simpler terms, the changes are:
- Split the "max size" eviction and the "min size" eviction while loops so the logic is easier to follow.
- Add a concept of "fully loaded" items. If an item isn't full loaded then it can be evicted from the cache if it's above the "max" size even if its marked as "used". This lets us stop any downloading or yet-to-be-parsed tile geometry if the cache has suddenly become full.
- The loading logic now checks the if the geometry memory usage has been assigned before determining if tile data should be evicted after loading.

gkjohnson:
PR should be ready, now.

src/utilities/LRUCache.js
```diff
@@ -135,6 +145,28 @@ class LRUCache {
 
 	}
 
+	// Marks whether tiles in the cache have been completely loaded or not. Tiles that have not been completely
+	// loaded are subject to being disposed early even if the are marked as used if the cache is full above its
+	// max size limits.
```
AnthonyGlt, line 150: nit There is a word missing here

AnthonyGlt: the test check is a good addition :+1:

gkjohnson: Thanks - just reworded it. Let me know once you think everything looks good and we can get this merged and a new version published.

AnthonyGlt (approved):
Looks good to me :+1: 
We got to juggle with the notions of `used` and `loaded` but at the end, it makes the cache management more subtle.
Well done :D

gkjohnson:
Thanks - yeah nomenclature could probably use some work but I think the functionality is good.

PR #768, opened 2024-10-01
draft-PR WIP: create component for R3f 3dtilesrendererjs

jo-chemla:
First shot at publishing an r3f component which wraps 3DTilesRendererJS with @alouis-jpg - draft to gather feedback and eventually rework a polished component and example. 
 - component stored at `/src/r3f/R3F3DTilesRenderer.jsx`
 - example created with [installation instructions](https://r3f.docs.pmnd.rs/getting-started/installation) from r3f doc with vite bundler, which calls the component at `/example/r3f-3dtilesrenderer/src/TilesComponents.jsx`
 - Only wraps standard TilesRenderer and DebugTilesRenderer for now, not Google or cesium ion renderers. 
 - needs to use `setLatLonToYUp` or define a matrixTransform matrix by the user - or set the camera to the location of your tileset so camera-controls (equivalent to orbit controls) works ok with y up. Uses `resetTransform` or passed as prop `matrixTransform` applied to root.

PR #769, opened 2024-10-02
 draft-PR WIP: create component for R3f 3dtilesrendererjs

jo-chemla:
Closing #768 to compare against r3f-dev branch

First shot at publishing an r3f component which wraps 3DTilesRendererJS with @alouis-jpg - draft to gather feedback and eventually rework a polished component and example. 
 - component stored at `/src/r3f/R3F3DTilesRenderer.jsx`
 - example created with [installation instructions](https://r3f.docs.pmnd.rs/getting-started/installation) from r3f doc with vite bundler, which calls the component at `/example/r3f-3dtilesrenderer/src/TilesComponents.jsx`
 - Only wraps standard TilesRenderer and DebugTilesRenderer for now, not Google or cesium ion renderers. 
 - needs to use `setLatLonToYUp` or define a matrixTransform matrix by the user - or set the camera to the location of your tileset so camera-controls (equivalent to orbit controls) works ok with y up. Uses `resetTransform` or passed as prop `matrixTransform` applied to root.


gkjohnson:
Amazing, thank you! I took a look and I was thinking that component could be more directly pass-through rather than handling anything like rotation or unique constructor instantiation. Then anything else can be handled via a plugin or a ref - something like this:

```jsx
<TilesRendererComponent
    url={ url }

    lruCache-minBytesSize={ 0.25 * 1e6 }
    lruCache-maxBytesSize={ 0.5 * 1e6 }
    downloadQueue-maxJobs={ 10 }
    errorTarget={ 40 }
    fetchOptions={ { mode: 'cors' } }

    /* event registration */
    onTileSetLoad={ e => {} }
    onModelLoad={ e => {} }
>
    { /* Plugin Wrapper */ }
    <TilesPluginComponent plugin={ DebugTilesPlugin } displayBoxBounds={ true } /* ...other options */ />

    { /* Position children */ }
    <EastNorthUpFrame lat={ lat } lon={ lon } height={ 0 } /* az, el, roll */>{ children }</EastNorthUpFrame>
    <TileRoot>{ children }</TileRoot>
</TilesRendererComponent>
```

Any options would be implicitly assigned to the TilesRenderer instance. I'll make a PR into this branch with the rough concept. Let me know what you think!

While we figure out the component structure - it would be nice to fix up build so we don't have to have a separate installation folder under `/examples` as well as getting linting working correctly for JSX.

gkjohnson:
Made the PR here: https://github.com/jo-chemla/3DTilesRendererJS/pull/1

Also in the future I'm thinking I'll be removing all sub classes of `TilesRenderer` (`CesiumIonRenderer`, `DebugTilesRenderer`, etc) and relying more on plugins for these things so I think it would be best to avoid using those classes in any new components.

jo-chemla:
Thanks for the feedback! I made a similar edit a few minutes ago, trying to extract code from the component and be a bit more generic, 
 - eg with plugins where I now pass a prop `rendererPlugins` which is an array of plugin js objects, defined within the root component that calls the renderer component. Your take on making these plugins components as well is indeed a nice design. By using the debug plugin, I removed the need to rely on `DebugTilesRenderer` but still use Google and CesiumIon renderers
 - and I also pass to the component a prop `addRendererEventListeners` which does add the `load-model`, and could proibably be used to pass additional variables like lruCache props. 

I'll look into your suggestion and try to integrate them. 

On my side, I still have no textures loading, with the following warnings:, still investigating. 
```
WebGL warning: texSubImage: Alpha-premult and y-flip are deprecated for non-DOM-Element uploads.
WebGL warning: drawElementsInstanced: Tex image TEXTURE_2D level 0 is incurring lazy initialization.
// And less important: 
TilesRenderer: tiles versions at 1.1 or higher have limited support. Some new extensions and features may not be supported. [TilesRendererBase.js:496:11](http://192.168.10.150:5173/@fs/C:/Dev/Iconem/3DTilesRendererJS/src/base/TilesRendererBase.js)
ThreeTilesRenderer : Non uniform scale used for tile which may cause issues when calculating screen space error. 3 
```

jo-chemla:
Hi again just had a look at your edits, thanks a lot! 
 - Really nice recursion for parsing the options path via  your `getPath (split on dashes) & getValueAtPath (recursively get attribute hierarchy`! This looks indeed way more generic, and more react-friendly for managing all tiles/tilesRenderers properties and options. 
 - Also, the useContext to get the tiles object for the `TilesPluginComponent` is also a nice way of doing it, with the depsArray to listen to props modification.

gkjohnson:
> with plugins where I now pass a prop rendererPlugins which is an array of plugin js objects, defined within the root component that calls the renderer component

One issue here is that if the tiles renderer needs to be recreated (eg the component url changes or something to display a new tile set) then these plugins can't be removed and readded to a new renderer. I suppose that could be changed but the plugins were designed with the idea that they would only ever be added to one tiles renderer and only get initialized once. I think maintenance and testing gets complicated otherwise. The components I added let them be recreated as-needed when the tiles renderer changes.

> On my side, I still have no textures loading, with the following warnings:, still investigating.

If you have a tile set you can share - even privately or just a single tile geometry file - I can take a look.

> Hi again just had a look at your edits, thanks a lot!

Thanks! If you're happy with the TilesRendererComponent maybe we can work from that, then, and get anything else we need added.

gkjohnson:
Sounds good - I'll merge these two PRs into the r3f-dev branch, then, and make an issue to track what else needs to be supported.

gkjohnson:
I've just made an issue here outlining some of the other work that will be needed. Feel free to add more in a comment! The items around linting and Vite set up at the root I could probably use the most help with since I'm less familiar with the set up those for react. https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/770

jo-chemla:
The utility classes for each renderer were pretty useful, having a constructor which took access-keys as args, but indeed the auth plugins and the associated components also make it easier to set them up within a new scene. 

Also thanks for the outline of the work in https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/770 that still needs to be addressed, very useful. Just saw you also moved the ellipsoid to the TilesRenderer but `setLatLonToYUp` is for now still in GoogleTilesRenderer

I'll merge your updates from your `r3f-dev` branch to test them out and see if there are remaining issues. I've noted your comment about linting and Vite set up, will need to investigate whether there are still multiple versions of three advertised.

gkjohnson:
> The utility classes for each renderer were pretty useful, having a constructor which took access-keys as args, 

Yeah I think this is just a code style preference but the plugin system if significantly more flexible, I feel, and enables things that couldn't really be done previously (the Cesium plugin needs to defer to the Google plugin if using Google's tiles). If users want the fully wrapped class it's very easy to do:

```js
function GoogleTilesComponent( { children, apiToken, ...rest } ) {
  return (
    <TilesRendererComponent { ...rest }>
      <TilesPluginComponent plugin={ GoogleCloudAuthPlugin } args={ { apiToken } } />
      { children }
    </TilesRendererComponent>
  );
}
```

> Just saw you also moved the ellipsoid to the TilesRenderer but setLatLonToYUp is for now still in GoogleTilesRenderer

Yeah I'll probably move this function to the main TilesRenderer for backwards compatibility but I'm thinking in the long run this kind of "automatic" placement of tiles for convenience can be moved to a plugin (eg scale to a size, center the tile set, correct the "up" orientation, etc)

jo-chemla:
Thanks for that sample again, that's very helpful and informative! I've tried using both that Google component as well as a tileset of ours in EPSG:4978 CRS, along the `EastNorthUpFrame` with given latlon, and it does not seem to apply the matrix transform "the right way" - nothing shows up, but some requests seem to go through. Am I using the components the way I should - example below?  

Moreover, would it also be possible to add an option, either via the `ENUFrame` or a `CustomFrame` to use a user-defined matrixTransform to apply to the root? This way the user would be in full control of the transform and local coordinate system mapping. Thanks again for all the help!

```
<group rotation-x={ Math.PI / 2 }>
  <TilesRendererComponent 
    url={ url} 
  >
    {fade && <TilesPluginComponent plugin={ TilesFadePlugin } fadeDuration={500} />}
    <EastNorthUpFrame lat = {41.902149} lon = {12.455084} height = {100} > </EastNorthUpFrame>
    {/* <TileSetRoot></TileSetRoot> */}
  </TilesRendererComponent>
</group>
```

gkjohnson:
> `<EastNorthUpFrame lat = {41.902149} lon = {12.455084} height = {100} > </EastNorthUpFrame>`

All of the lat and lon values in `Ellipsoid` are passed as radians so for consistency they passed as radians here, too, meaning you'll have to convert them before passing them in.

> Moreover, would it also be possible to add an option, either via the ENUFrame or a CustomFrame to use a user-defined matrixTransform to apply to the root? This way the user would be in full control of the transform and local coordinate system mapping. Thanks again for all the help!

Which root? If you want to transform the whole tile set you can do this:

```html
<group position={ ... } rotation={ ... }>
  <TilesRendererComponent ... />
</group>
```

If you want to transform children in the local frame of the root of the tile set you can do this:

```html
<TilesRendererComponent ...>
  <group position={ ... } rotation={ ... } />
</TilesRendererComponent>
```

jo-chemla:
Thanks for that feedback. I tried re-running the example with the ENUFrame after pulling the commits to r3f-dev from this weekend, but I'm getting error `tiles is null in EastNorthUpFrame TilesRenderer.jsx:49` with the following component: 

```

          <group rotation-x={ Math.PI / 2 }>
            <GoogleTilesComponent apiToken={googleApiKey} url={'https://tile.googleapis.com/v1/3dtiles/root.json'}>
                <EastNorthUpFrame
                    lat = {41.902149 * Math.PI / 180}
                    lon = {12.455084 * Math.PI / 180}
                    height = {100}
                    az = {0}
                    el = {0}
                    roll = {0}
                  > </EastNorthUpFrame>
            </GoogleTilesComponent>
            </group>```

Regarding applying a matrix-transform rather than position/rotation/scale to the tileset root or all leaves, this is because the matrix4 transform we apply - to get perfect registration/alignement for datasets represented in different coordinate systems - has not exactly uniform scale, and therefore cannot be decomposed via threejs standard methods. Below is a warning I added to threejs doc on [Matrix-transformations](https://threejs.org/docs/#manual/en/introduction/Matrix-transformations) which was slightly edited in latest version of the doc

> An object can be transformed via [page:Object3D.applyMatrix4].
> Note: Under-the-hood, this method relies on [page:Matrix4.decompose], and not all matrices are decomposable in this way.
For example, if an object has a non-uniformly scaled parent, then the object's world matrix may not be decomposable, and this method may not be appropriate.
> In this case, it might be safer to use the second method `object.matrix.premultiply(transform);`
> Since we also wan't to avoid matrix4 recomputation, use instead applyMatrix4_matrixVersion

**Comment by gkjohnson** [comment]
> I'm getting error tiles is null in EastNorthUpFrame TilesRenderer.jsx:49 with the following componen

Looks like this is caused because we're not checking for `null` when reading the ellipsoid from the tile set (see [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/224e1bcaa025e01c348d217e45b9e838d8fb2310/src/r3f/TilesRenderer.jsx#L44-L55)). It should be an easy fix if you'd like to make a PR with the change to check before running the effect function.

> PS: I also tried using GLTFExtensionsPlugin with the GoogleTilesComponent, but receive an error

It should work with `args` since that's how I tested it but I'm not sure why you're approach isn't work. Would you be able to look into it?

> to get perfect registration/alignement for datasets represented in different coordinate systems - has not exactly uniform scale

If you mean you're applying skew transformations - then yeah those can't be represented with a pos, quat, and scale. But you can still assign whatever custom matrices you want to the parent object:

```jsx
const callback = useMemo( () => o => {
  o.matrix.copy( skewMatrix );
  o.matrixAutoUpdate = false;
} } );

return <group ref={ callback }><TilesRenderer /* ... */ /></group>;
```

Or any number of other ways. You can also create a plugin to assign this directly to the tiles group rather than add a parent component. I guess the question is whether or not you want this to skew transform to apply to all the "children" of the tile set, as well.

> It seems that using ReorientationPlugin instead of EastNorthUpFrame does load the right tiles

`EastNorthUpFrame` does not transform the tile set root while `ReorientationPlugin` _does_ transform the tile set. `EastNorthUpFrame`, instead, creates a frame on the surface of the globe that you can add children to if you, for example, want to create markers on the planet surface.

jo-chemla:
Thanks again for all the replies!
- Indeed, the whole implementation you've come up with in the last 5 days is a great design. I'll try to see the remaining bits of https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/770 once I've got a clean slate. 
- Thanks for the clarification between `ReorientationPlugin` (bring tileset coordinate back to origin) and `EastNorthUpFrame` (new coordinate system attached to user-provided origin for children), great that both do exist. also thanks for the sample to apply custom matrix - I thought about applying the matrix only to tileset root, which should propagate all the way through the hierarchy towards leaf meshes by composing matrixTransform of every parent along the way. PS that ENUFrame therefore looks like [react-three-map](https://github.com/RodrigoHamuy/react-three-map) which offers a way to mix se [react-three-fiber](https://github.com/pmndrs/react-three-fiber) inside [react-map-gl](https://github.com/visgl/react-map-gl).
- understood for ~removing original values restoration~ reverting original props values
- Indeed, passing the loaders to the `GLTFExtensionsPlugin` as args object rather than as props retrieved via `{...options}` does work! The `...options` is correctly deconstructed, but it seems that `useShallowOptions( instance, options )` hook does not apply them to the loader. It appears that `instance` is undefined on that call - and when using several plugins, the options are not always defined. After digging through I could ont find the explanation for this. 

```jsx
// Not working
<TilesPlugin plugin={ GLTFExtensionsPlugin } 
        dracoLoader={dracoLoader}
        ktxLoader={ktx2Loader}
        autoDispose={false}

      />
// ok
<TilesPlugin plugin={ GLTFExtensionsPlugin } 
        args = {{
          dracoLoader, ktxLoader:ktx2Loader
        }}
      />
```

gkjohnson:
> The ...options is correctly deconstructed, but it seems that useShallowOptions( instance, options ) hook does not apply them to the loader.

Looks like an order-of-operations issue - the plugin is created and then attached before any of the local options can be applied (that particular plugin needs the DRACOLoader available on init). I'll make a PR to address it. Though note that this won't handle the case where the draco loader is set initially and then changed later.

Issue #770, opened 2024-10-03
Add support for react-three-fiber


gkjohnson:
Related to #769 

- *Add a README to the r3f folder and link from the root readme
- ~Fix up vite so the root config supports jsx~
- ~Add an example~
- ~Add improved support for recursive dependency array checking~
- ~Fix up linting rules to support jsx~
- ~Fix EastNorthUpFrame~
- ~Decide if parameters should automatically revert to a previous state if removed (r3f does not)~
  - R3F actually does reset to the original value on removal of the field.
- ~Add plugin to make it easier to add draco, ktx2, other extension support for glTF.~
- ~Add plugin to facilitate easy reorientation of the tile set on load.~
- ~Add support for controls (EnvironmentControls, GlobeControls)~
- ~Add support for `EeastNorthUpFrame` component that will provide a frame at a point and orientation on the globe.~
- ~By default ensure children are added to the same frame as the tile set.~
- ~#766~
- ~Add `3d-tiles-renderer/r3f` export~
- ~Add support for "ref" to plugins, TilesRendererComponent~
- ~Add implicit support for credits display string (google tiles, ion)~
- ~Add a "credits display" component~

**Later**
- Add prop types?
- Make attributions component more flexible (ordering, styling, collapsing, children placement)

gkjohnson:
> Add plugin to facilitate easy reorientation of the tile set on load.

```js
const plugin = new TransformAdjustmentPlugin( {
  // whether to center the tile set at the origin
  center: true,
  
  // whether to reorient the tile set based on a derived lat / lon or region
  reorient: true,
  
  // the axis to fall back to when reorienting up
  upAxis: 'z',
  
  // if these are present the other parameters are ignored
  lat, lon, height,
} );

// or

plugin.setLatLonHeightToOrigin( lat, lon, height );
```

> Add plugin to make it easier to add draco, ktx2, other extension support for glTF.

```js
const plugin = new GLTFExtensionsPlugin( {
  // other extension plugins to register
  extensions: [ ... ],

  // extension fields
  metadata: true,
  rtcCenter: true,

  // loaders
  dracoLoader: dracoLoader,
  ktx2Loader: ktx2Loader,
  autoDispose: true,
} );
```

PR #771, opened 2024-10-03
R3F: Add transform roots, other updates

gkjohnson:
Related to #770 

- Add a transform for tiles renderer children to ensure they align with the tile root.
- Add EastNorthUpFrame component for positioning things on the globe surface.
- Remove trailing white space in example
- Add `3d-tiles-renderer/r3f` export.

cc @jo-chemla 

I was also trying to test with the google tile renderer but the tiles seemed to be loading in the wrong spot. I'm wondering if this has anything to do with the multiple three.js imports. We may have to fix the vite situation sooner rather than later.

### DIFF
--- a/example/r3f-3dtilesrenderer/src/DemoTiles.jsx
+++ b/example/r3f-3dtilesrenderer/src/DemoTiles.jsx
@@ -10,9 +10,9 @@ import { TransformControls, Grid, GizmoHelper, GizmoViewport } from "@react-thre
 import { useControls, folder } from 'leva'
 
 import {DebugTilesPlugin, NONE, SCREEN_ERROR, GEOMETRIC_ERROR, DISTANCE, DEPTH, RELATIVE_DEPTH, IS_LEAF, RANDOM_COLOR, RANDOM_NODE_COLOR, CUSTOM_COLOR, LOAD_ORDER} from '../../../src/three/plugins/DebugTilesPlugin'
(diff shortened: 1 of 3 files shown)


jo-chemla:
I'll open up a private dataset url for you to try out. In my case, the dataset we exported from RealityCapture over the Roma area overlaps perfectly with the Google one. However, I cannot manage to have the right matrix setup for the cesium ion 58587 new York dataset, I imagine they are shifted via another property (maybe need to use that RTC_CE'TER plug-in)

Both Iconem and Google 3d-cities seem to be expressed in CRS EPSG:4978. I've applied the same matrix that transforms axes around a user-selected origin to be y-up, similar to your setLatLonToYUp, and translate that coords towards the threejs scene origin so the whole setup is around origin. Great idea to have that ENUFrame to handle this via a component as well! 

The vite bundling bits I was mentioning are not present anymore with my setup, what are the issues you're experiencing?

gkjohnson:
> The vite bundling bits I was mentioning are not present anymore with my setup, what are the issues you're experiencing?

When positioning the camera at the surface of the globe I'm seeing this:

image

And this is logged when starting the page, meaning there's an issue with the build set up (ie we're using two separate node_modules folders):

image

And I'm seeing this, as well, but I'm not sure what it's from:

image

gkjohnson:
> When positioning the camera at the surface of the globe I'm seeing this:

Found that this was caused by using an incorrect matrix 🎉 

image

jo-chemla:
This is looking so good!

Issue #788, opened 2024-10-07
Add middle-click + drag listener for ZOOM on EnvironmentControls


jo-chemla:
**Is your feature request related to a problem? Please describe.**

The `EnvironmentControls` parses left and right clicks + mouse-wheel events, (Quick note, the `SHIFT + LEFT_CLICK` did not trigger rotate during my tests). It could be nice if these controls could also parse Middle click for zoom - the awy it is done in [googleMapsAerial](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/googleMapsAerial.html) within [OrbitControls](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/4246f179ae16f992fb9bd0b1de17b48c7ad7a087/example/src/lib/OrbitControls.js#L678)

**Describe the solution you'd like**

[Web/API/Pointer_events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) suggest adding to [PointerTracker.js](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/4246f179ae16f992fb9bd0b1de17b48c7ad7a087/src/three/controls/PointerTracker.js#L251)

```js
isMiddleClicked() {
  return Boolean( this.buttons & 4 );
}
```

And then this could probably resort to something like the following addition to [EnvironmentControls.js - pointerdownCallback ](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/4246f179ae16f992fb9bd0b1de17b48c7ad7a087/src/three/controls/EnvironmentControls.js#L302-L311)

```js
else if ( pointerTracker.isMiddleClicked() ) {

	// if the clicked point is coming from below the plane then don't perform the drag
	this.setState( ZOOM );
	this.zoomDirectionSet = false;
	this._updateZoomDirection();

        // The below could be shared on top if the if (hit) block
        this.pivotPoint.copy( hit.point );
        this.pivotMesh.position.copy( hit.point );
        this.pivotMesh.updateMatrixWorld();
        this.scene.add( this.pivotMesh );
}
```

**Describe alternatives you've considered**

Use another controls, like Pivot controls instead of EnvironmentControls to handle middle-mouse click

gkjohnson:
I'm open to adding the feature if you like to make a PR!

jo-chemla:
Just did a try, seems like this is working! 
https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/789

[...]

PR #789, opened 2024-10-07
Handle Middle Click Drag for Environment Control

jo-chemla:
Following issue https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/788
Handle Middle Click Drag for Environment Control

### DIFF
--- a/src/three/controls/EnvironmentControls.js
+++ b/src/three/controls/EnvironmentControls.js
@@ -284,6 +284,11 @@ export class EnvironmentControls extends EventDispatcher {
 			const hit = this._raycast( raycaster );
 			if ( hit ) {
 
+				this.pivotPoint.copy( hit.point );
+				this.pivotMesh.position.copy( hit.point );
+				this.pivotMesh.updateMatrixWorld();
+				this.scene.add( this.pivotMesh );
+
 				// if two fingers, right click, or shift click are being used then we trigger
 				// a rotation action to begin
 				if (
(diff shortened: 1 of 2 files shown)


src/three/controls/EnvironmentControls.js
```diff
@@ -399,6 +400,17 @@ export class EnvironmentControls extends EventDispatcher {
...
 			}
 
+			// Add mouse click handle in pointerMove
+			if ( this.state === ZOOM ) {
+
+				const delta = e.movementY;
+				// use LOG to scale the scroll delta and hopefully normalize them across platforms
+				const deltaSign = Math.sign( delta );
+				const normalizedDelta = Math.log( Math.abs( delta ) + 1 );
+				this.zoomDelta -= 3 * deltaSign * normalizedDelta * 1;
```
gkjohnson, line 410: Log scale shouldn't be needed if we're moving in screen space. It's used because the reported wheel "scroll" has pretty severely different magnitudes depending on platform.

gkjohnson:
Just got back home and tested it with a mouse - as you allude to in https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/788, I'm seeing that the drag point shifts as you move the mouse around. This is probably because the "zoom direction" is getting reset in different conditions. Like when the [pointer moves](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/controls/EnvironmentControls.js#L324) and the [ellipsoid frame has been rotated](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/b87fc5c6ec8e8808b6ff0027f593ce0716682f4d/src/three/controls/EnvironmentControls.js#L1190).

These are both cases that are valid when using pinch-to-zoom on mobile, though (since you may move your fingers around to zoom to different spots) so we'll need to be careful to not break that case when addressing regenerating the zoom direction.

gkjohnson:
Closing this for now - happy to get a version merged if we can get one going that doesn't break some of the existing zoom behavior.

Issue #730, opened 2024-08-30
DebugTilesPlugin: Add support for displaying parent bounds


gkjohnson:
Can be added efficiently without significant performance impact by traversing the ancestors when making the tile visible, adding all parent bounds, and then incrementing a reference counter per parent tile. When a tile is hidden the reference counter for each ancestor can be decremented and if it reaches 0 then the parent bounds can be removed.

gkjohnson:
cc @sguimmara I'm not sure if this debug feature is interesting for you since you're working in that plugin.

sguimmara:
@gkjohnson  I could look into it but I don't really understand what the issue is wanting to do...

gkjohnson:
Sorry it's unclear - right now only "leaf tile" bounds are displayed when toggling them:

<img width="450" alt="image" src="https://github.com/user-attachments/assets/c2e24523-adfe-4a2c-bb51-1de75c549fe3" />

Of course the parent tiles also have bounds that encapsulate all the child bounds that must be in the frustum before the leaves are displayed. So this issue is about adding an option to display the parent bounds of all the visible leaf tiles in addition to the leaf bounds. Hopefully that's more clear.

sguimmara:
> Sorry it's unclear - right now only "leaf tile" bounds are displayed when toggling them:
> 
[...]

PR #791, opened 2024-10-07
feat(ImplicitTiling): add octree support

AnthonyGlt:
Add octree support for implicitTiling.

TODO:
- add a method to get directly the type of subdivision or use getBoundsDivider ?
- test on region octree tileset. I don't have a tileset to try on

### DIFF
--- a/src/base/plugins/SUBTREELoader.js
+++ b/src/base/plugins/SUBTREELoader.js
@@ -16,7 +16,7 @@ function isOctreeSubdivision( tile ) {
 
 function getBoundsDivider( tile ) {
 
-	return isOctreeSubdivision(tile) ? 8 : 4;
+	return isOctreeSubdivision( tile ) ? 8 : 4;
 
 }
 
@@ -31,7 +31,7 @@ function getSubtreeCoordinates( tile, parentTile ) {
 
 	const x = 2 * parentTile.__x + ( tile.__subtreeIdx % 2 );
 	const y = 2 * parentTile.__y + ( Math.floor( tile.__subtreeIdx / 2 ) % 2 );
-	const z = isOctreeSubdivision(tile) ?
+	const z = isOctreeSubdivision( tile ) ?
 		2 * parentTile.__z + ( Math.floor( tile.__subtreeIdx / 4 ) % 2 ) : 0;
 
 	return [ x, y, z ];
@@ -754,7 +754,7 @@ export class SUBTREELoader extends LoaderBase {
 			}
 
 			//Also divide the height in the case of octree.
-			if ( isOctreeSubdivision(tile) ) {
+			if ( isOctreeSubdivision( tile ) ) {
 
 				const minZ = region[ 4 ];
 				const maxZ = region[ 5 ];
@@ -780,7 +780,7 @@ export class SUBTREELoader extends LoaderBase {
 			const box = [ ...this.rootTile.boundingVolume.box ];
 			const cellSteps = 2 ** tile.__level - 1;
 			const scale = Math.pow( 2, - tile.__level );
-			const axisNumber = isOctreeSubdivision(tile) ? 3 : 2;
+			const axisNumber = isOctreeSubdivision( tile ) ? 3 : 2;
 
 
 			for ( let i = 0; i < axisNumber; i ++ ) {
--- a/src/base/plugins/SUBTREELoader.js
+++ b/src/base/plugins/SUBTREELoader.js
@@ -6,11 +6,20 @@ import { LoaderBase } from '../loaders/LoaderBase.js';
 import { readMagicBytes } from '../../utilities/readMagicBytes.js';
 import { arrayToString } from '../../utilities/arrayToString.js';
 
-function getBoundsDivider( tile ) {
 
-	return tile.implicitTiling.subdivisionScheme === 'QUADTREE' ? 4 : 8;
+
+function isOctreeSubdivision( tile ) {
+
+	return tile.__implicitRoot.implicitTiling.subdivisionScheme === 'OCTREE';
 
 }
+
+function getBoundsDivider( tile ) {
+
+	return isOctreeSubdivision(tile) ? 8 : 4;
+
+}
+
 
 function getSubtreeCoordinates( tile, parentTile ) {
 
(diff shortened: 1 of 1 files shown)


AnthonyGlt:
@gkjohnson 
Is there really a thing to add here ?
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/727a44021ce975fea3d79769a7ac62e5213019d4/src/base/plugins/SUBTREELoader.js#L773-L774

gkjohnson:
> add a method to get directly the type of subdivision or use getBoundsDivider ?

What do you mean by this?

> test on region octree tileset. I don't have a tileset to try on

Looks like `3d-tiles-samples/1.1/SparseImplicitOctree/tileset.json` works for this, right?

> Is there really a thing to add in the case of a box ?

The loop referenced adjusts the bounding box center and half-axes to describe the specific subdivided child. Note that the array for a bounding box is [defined as](https://github.com/CesiumGS/3d-tiles/tree/main/specification#box) 3 values for the center, and 3 values for each half-vector on each axis.

Currently the [loop for adjusting the bounds](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/727a44021ce975fea3d79769a7ac62e5213019d4/src/base/plugins/SUBTREELoader.js#L775-L793) does two things - it adjusts the [half axes](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/727a44021ce975fea3d79769a7ac62e5213019d4/src/base/plugins/SUBTREELoader.js#L778-L780) and adjusts the [bounding box center](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/727a44021ce975fea3d79769a7ac62e5213019d4/src/base/plugins/SUBTREELoader.js#L788-L791) by each necessary axis. Currently the loop only iterates over two axes to adjust their scale and the center because a quad tree only splits along X and Y.

So two things need to happen:
- The loop needs to iterate over all the half-width vectors (ie changing 2 to 3 when using an OCTREE split).
- And accommodate the z-axis (ie `i===2`) when computing the `axisOffset` value so the center can be offset in the z-half-width direction in addition to the x and y offsets currently applied.

Hopefully that makes sense - let me know if anything is still unclear

AnthonyGlt:
Thank you for the answer :D

>
    add a method to get directly the type of subdivision or use getBoundsDivider ?

> What do you mean by this?

- I meant that I'm using this condition `tile.__implicitRoot.implicitTiling.subdivisionScheme === 'OCTREE' `, could be smarter to use an other method

>The loop needs to iterate over all the half-width vectors (ie changing 2 to 3 when using an OCTREE split).

- Yes, I must have had a brainfreeze to miss that :grimacing: 


>
    test on region octree tileset. I don't have a tileset to try on

>Looks like 3d-tiles-samples/1.1/SparseImplicitOctree/tileset.json works for this, right?

It's bounding box, I'll modify it to have region then :)

AnthonyGlt:
Screenshot from 2024-10-08 10-14-38

gkjohnson:
> I meant that I'm using this condition tile.__implicitRoot.implicitTiling.subdivisionScheme === 'OCTREE' , could be smarter to use an other method

Yeah it's a little long - maybe a function called "isOctreeSubdivision" that returns true or false would make it more readable.

> 

Looks great! It would be nice to get parent bounds rendering at some point so we can see the full hierarchy a la #730.

src/base/plugins/SUBTREELoader.js
```diff
@@ -773,7 +785,7 @@ export class SUBTREELoader extends LoaderBase {
 				const z = box[ 3 + i * 3 + 2 ];
 
 				// adjust the center by the x and y axes
-				const axisOffset = i === 0 ? tile.__x : tile.__y;
+				const axisOffset = i === 0 ? tile.__x : ( i === 1 ? tile.__y : tile.__z );
```
gkjohnson, line 788: Looks like the comment above can be updated.

src/base/plugins/SUBTREELoader.js
```diff
@@ -743,6 +744,17 @@ export class SUBTREELoader extends LoaderBase {
...
+			if ( tile.__implicitRoot.implicitTiling.subdivisionScheme === 'OCTREE' ) {
+
+				const minZ = region[ 4 ];
+				const maxZ = region[ 5 ];
+
+				const sizeZ = ( maxZ - minZ ) / Math.pow( 2, tile.__level );
+				region[ 4 ] += tile.__z * sizeZ;
+				region[ 5 ] = region[ 4 ] + sizeZ;
+
+			}
```
gkjohnson, line 756: Would be good to add some comments here, I think, regarding what this is adjusting and how.

src/base/plugins/SUBTREELoader.js
```diff
@@ -16,15 +16,16 @@ function getSubtreeCoordinates( tile, parentTile ) {
...
-		return [ 0, 0 ];
+		return [ 0, 0, 0 ];
 
 	}
 
 	const x = 2 * parentTile.__x + ( tile.__subtreeIdx % 2 );
 	const y = 2 * parentTile.__y + ( Math.floor( tile.__subtreeIdx / 2 ) % 2 );
-	//TODO z coord
+	const z = tile.__implicitRoot.implicitTiling.subdivisionScheme === 'OCTREE' ?
+		2 * parentTile.__z + ( Math.floor( tile.__subtreeIdx / 4 ) % 2 ) : 0;
```
gkjohnson, line 26: Interestingly just setting z to "0" here would make all the octtree / quadtree logic "just work" even if we don't have conditions for the two tree types 🤔. Maybe worth making a comment for later

AnthonyGlt: I don't think that the `0` value even matters, it could be whatever. In the case of a quadtree, the Z value is not used in any calculus and is called only at the url parsing, but there wont be a "z" coordinate in a quadtree url

gkjohnson:
Looks great, thanks!

phoenixbf:
Hi all, I join the discussion since cesium ion service removed the tiler switch (1.0 / 1.1): generated tilesets (now only v1.1) seem to throw the same error (`TilesRenderer : Failed to load tile at url "tiles/{level}/{x}/{y}/{z}.glb"`). I suppose it is related to implicit tiling support.

Let me know if some help is needed, or it's something already planned in the next release (?)

gkjohnson:
Hi @phoenixbf - support for implicit tiling is added through a [plugin](https://github.com/NASA-AMMOS/3DTilesRendererJS/tree/master/src/plugins#implicittilingplugin) for now which was done to reduce imported file sizes and keep the logic compartmentalized. It would be nice to automatically dynamically import the plugins needed by a specific tile set but unfortunately that's not well supported across build processes.

You can use it like so:

```js
import { ImplicitTilingPlugin } from '3d-tiles-renderer/plugins';

// ...

const tiles = new TilesRenderer( url );
tiles.registerPlugin( new ImplicitTilingPlugin() );
```

Happy to take a PR that improves the documentation to make things more clear.

phoenixbf:
Thank you so much Garrett!
Yeah I missed the plugin bit, it works now with v1.1 ion-generated tilesets 👍 

Indeed an on-demand loading would be nice in the future depending on ts version (even if I suppose the 1.0 is being completely dropped in the future).
I'll follow the checklist features and hopefully contribute where possible, also a dedicated example could be helpful for newcomers.

Issue #778, opened 2024-10-04
Deprecate and remove the GooglePhotorealisticTilesRenderer


gkjohnson:
Open questions:
- Where should the other default settings be assigned? In the plugin?
- Should the plugin also handle the tiles URL assignment?

Related to #770

jo-chemla:
Just saw that thread. I was cleaning an example based on the r3f components, and did notice the CesiumIonAuthPlugin assigns endpointUrl on init based on assetId, while GoogleAuthPlugin does not. I edited it to include a fallback google endpoint url on init as well 'https://tile.googleapis.com/v1/3dtiles/root.json'. Will do a PR with these and a few other minor fixes asap.

PR #793, opened 2024-10-08
r3f-dev minor fixes

jo-chemla:
Related to #778, #770

- add default tileset root URL to GoogleCloudAuthPlugin the same way CesiumIonAuthPlugin builds one from assetId
- catch tiles === null in TilesRenderer
- edit CesiumIon doc accessToken  -> apiToken
- add @react-three/fiber dependency to root package.json
- minor fix to old r3f componentR3F3DTilesRenderer


jo-chemla:
Also simplified App.js to have more granular components for Staging, TilesRendererDemo, etc. 

Note I wanted to add a Leva control which would allow the user to laod another Cesium Ion asset Id or standard tileset url, but it seems that unmounting the components yields error `TilesRendererBase: A plugin can only be registered to a single tile set in registerPlugin TilesRendererBase.js:165`. I tried attributing a key to the plugin component so it re-renders when assetId changes, but it did not resolve this error. 

Note also I've got a large number of warnings `Material: onBeforeRender() has been removed.` when the tilesrenderer plugins are used.

package.json
```diff
@@ -61,5 +61,8 @@
   },
   "peerDependencies": {
     "three": ">=0.123.0"
+  },
+  "dependencies": {
+    "@react-three/fiber": "^8.17.9"
```
gkjohnson, line 66: This should be added under [optionalDependencies](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#optionaldependencies) in addition to `react` and `react-dom` since those are used in the project.

src/r3f/R3F3DTilesRenderer.jsx
```diff
@@ -57,7 +57,7 @@ function R3F3DTilesRenderer(props) {
     }
 
     if (props.rendererPlugins) 
-      for (plugin of props.rendererPlugins) 
+      for (const plugin of props.rendererPlugins)
```
gkjohnson, line 60: I think we should look into removing this and the "utils.jsx" files since they've been superseded by the TilesRenderer component. Unless there's something else you're thinking?

src/r3f/TilesRenderer.jsx
```diff
@@ -165,6 +171,12 @@ export const TilesRenderer = forwardRef( ( props, ref ) => {
 	// add the camera
 	useEffect( () => {
 
+		if ( tiles === null ) {
+
+			return;
+
+		}
+
```
gkjohnson, line 179: This looks redundant with the `tiles === null` condition below

src/three/plugins/GoogleCloudAuthPlugin.js
```diff
@@ -46,6 +47,15 @@ export class GoogleCloudAuthPlugin {
...
 	init( tiles ) {
 
+		if ( this.endpointURL === null ) {
+
+			this.endpointURL = 'https://tile.googleapis.com/v1/3dtiles/root.json';
+
+		}
+
+		tiles.rootURL = this.endpointURL;
+
```
gkjohnson, line 57: I think we should only assign this if the user hasn't set root url when creating `TilesRenderer` - I don't think there's a lot of value in taking a full URL through the tiles constructor _and_ the plugin options. The plugin should just assign a default if the user hasn't provided one.

gkjohnson (commented):
Thank you!

> I've got a large number of warnings Material: onBeforeRender() has been removed. when the tilesrenderer plugins are used - either Google, cesium, or standard with the sample msl-dingo-gap tileset.

I expect this and the multiple three.js instances are due to the nested package.json - we should remove that and set up the root vite instance to support building for react, as well.

> I had to add a key prop to the CesiumIonAuthPlugin to remove this error.

This sounds like a bug we should fix rather than work around.

> but TilesFadePlugin yields error this.onBeforeCompile is null i customProgramCacheKey getParameters

Can you submit a separate issue for this.

> It seems that changing the standard tileset url does yield a reload of the TilesRenderer component with the new tileset url, and I had to add a key prop to the CesiumIonAuthPlugin to remove this error.

The attribution overlay has been working for me - though there is one case I've seen that results in the attributions disappearing but it's more rare and I'll fix it, soon. I'll have to take a look.

> when using drei CameraControls and a custom PerspectiveCamera to edit its far clip plane with the google tileset, it seems that tiles in the forefround are loading after tiles in the background, and this become more noticeable with higher far values (15000m in my case).

This is something that will need a more dedicated repro case - so you can just load the page and see the issue. I think this deserves a separate issue.

---

Regarding the demo - I think it's best to keep these kinds of things dead simple without a lot of indirection in custom components or toggle-able options so users can easily see how to use this stuff. I'm thinking one demo with a mars tile set and controls and another one with the google tiles or lunar tiles and globe controls. Maybe we can add some kind of iconic drei component, as well, so it's more clear this is rendering in react.

jo-chemla:
Thanks again, will review your suggestions!
Just a quick heads-up, I've "fixed" the Google attribution by adding a string attribution object that encompasses the whole attribution string in the getAttribution method: 

```js

	getAttributions( target ) {

		target.push( {
			value: this._attributionsManager.toString(),
			type: 'string',
			required: true,
		} );

	}
```

src/r3f/R3F3DTilesRenderer.jsx
jo-chemla: Yes you're definitely right, these were just lying around while I wanted to test the applyMatrix4 to the tileset to see I can replicate a perfect registration between one of our iconem tileset and google 3d cities. Probably DemoTiles.jsx will vanish in the end as well, since indeed the demo should probably be as light as possible.

src/r3f/TilesRenderer.jsx
jo-chemla: sorry about that

src/three/plugins/GoogleCloudAuthPlugin.js
jo-chemla: Agreed. 

Will replace with 
```js

	init( tiles ) {

		if ( tiles.url === null ) {

			tiles.url = 'https://tile.googleapis.com/v1/3dtiles/root.json';

		}
        //...
```

gkjohnson:
> Just a quick heads-up, I've "fixed" the Google attribution by adding a string attribution object that encompasses the whole attribution string in the getAttribution method.

There were some fixes in the master branch that hadn't made it to r3f-dev - I just updated the branch if you want to merge it in. That should fix it.

Let me know when you're ready for another review!

gkjohnson:
> Just a quick heads-up, I've "fixed" the Google attribution by adding a string attribution object that encompasses the whole attribution string in the getAttribution method.

There were some fixes in the master branch that hadn't made it to r3f-dev - I just updated the branch if you want to merge it in. That should fix it.

Let me know when you're ready for another review!

src/three/plugins/GoogleCloudAuthPlugin.js
```diff
@@ -46,6 +47,15 @@ export class GoogleCloudAuthPlugin {
 
 	init( tiles ) {
 
+		if ( this.endpointURL === null ) {
+
+			this.endpointURL = 'https://tile.googleapis.com/v1/3dtiles/root.json';
+
+		}
+
+		tiles.rootURL = this.endpointURL;
```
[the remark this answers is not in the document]
jo-chemla: Done conditionnally, if an arg is passed to the GoogleCloudAuthPlugin constructor, when `useRecommendedSettings = true (default)`

jo-chemla:
Thanks again for the detailed feedback.

> Re Google attribution: There were some fixes in the master branch that hadn't made it to r3f-dev - I just updated the branch if you want to merge it in. That should fix it.

 I had to remove a check for google attribution to work, which was `if ( this._attribution.value ) {` since _attribution had value null. The `TilesAttributionOverlay` now works as it should with google credits in latest commit! 



> large number of warnings `Material: onBeforeRender() has been removed`: I expect this and the multiple three.js instances are due to the nested package.json - we should remove that and set up the root vite instance to support building for react, as well.

Will investigate, not sure I can do much on that front

> I had to add a key prop to the CesiumIonAuthPlugin to remove this error: This sounds like a bug we should fix rather than work around.

Agreed, not sure where this error comes from. 

> but TilesFadePlugin yields error this.onBeforeCompile is null in customProgramCacheKey getParameters

Just submitted dedicated issue


Regarding repro case for tiles loading performance, I'll see if I can narrow it down and report back!

src/three/plugins/GoogleCloudAuthPlugin.js
```diff
@@ -51,9 +52,31 @@ export class GoogleCloudAuthPlugin {
...
+			tiles.parseQueue.maxJobs = 10;
+			tiles.downloadQueue.maxJobs = 30;
+			tiles.errorTarget = 40;
+
+		}
+
 		this.tiles = tiles;
 		this._onLoadCallback = ( { tileSet } ) => {
 
+
```
gkjohnson, line 79: nit: lets remove some of this unnecessary added whitespace.

src/three/plugins/GoogleCloudAuthPlugin.js
```diff
@@ -64,7 +87,10 @@ export class GoogleCloudAuthPlugin {
 
 		this._visibilityChangeCallback = ( { tile, visible } ) => {
 
+
 			const copyright = tile.cached.metadata.asset.copyright || '';
+
+
```
gkjohnson, line 93: nit: whitespace

src/three/plugins/GoogleCloudAuthPlugin.js
```diff
@@ -84,20 +110,16 @@ export class GoogleCloudAuthPlugin {
...
-
-			this._attribution.value = this._attributionsManager.toString();
-			target.push( this._attribution );
+			this._logoAttribution.value = this.logoUrl;
+			target.push( this._logoAttribution );
 
 		}
 
+		this._attribution.value = this._attributionsManager.toString();
+		target.push( this._attribution );
```
gkjohnson, line 121: It looks like there I miscommitted something previously - rather than checking the attribution string we should check the number of visible tiles - see [this condition in master](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/plugins/GoogleCloudAuthPlugin.js#L86). I'd prefer to keep it that way so we don't return any attributions if there are no tiles to display attributions for.

PLUGINS.md
```diff
@@ -327,14 +327,20 @@ The function used to map a [0, 1] value to a color for debug visualizations. By
 constructor( { accessToken : String, autoRefreshToken = false : Boolean, logoUrl = null : String | null } )
 ```
 
-Takes the Google Cloud access token. If `autoRefreshToken` is set to true then the plugin will automatically perform a new root tile request once the existing token has expired after four hours. The url in "logoUrl" will be used to create an image attribution required for use with Google Photorealistic Tiles.
+Takes the Google Cloud access token. If `autoRefreshToken` is set to true then the plugin will automatically perform a new root tile request once the existing token has expired after four hours.
+This plugin changes below values to be more efficient for the photorealistic tiles if `useRecommendedSettings = true (default)`:
```
gkjohnson, line 331: We should also add this field to the constructor above

gkjohnson (commented):
> I anticipated that r3f example could be in its own standalone directory/project, with its own package.json and vite bundler ... it might indeed be completely feasible to isolate a single html+js and let the root vite bundler handle it.

This is what I was thinking. Having two package.json, two node_module folders, and therefore two versions of three.js imported is going to cause issues and isn't something we should implicitly recommend to people.

> as well as to the TilesPlugin CesiumIonAuthPlugin to avoid this error TilesRendererBase: A plugin can only be registered to a single tile set in registerPlugin TilesRendererBase.js:165

The plugin should be disposed and recreated if a new tile renderer is created. It's not obvious why this is happening yet, though.

_edit: I'm not able to run the demo page in this PR (looks like an issue with Leva?) and it's not super clear to me what's happening with all the components from looking at it briefly, so I'll have to see a simpler case, I think._

jo-chemla:
Just added a readme to the last commit in the src/r3f directory. Should this be moved to the project root?

jo-chemla:
Finally managed to get a working, minimal html + js r3fDemo in the example directory, so I removed the more complex r3f-project directory 🎉 
This required adding the react plugin to the vite config, as well as reworking Leva controls a bit - which was using undefined env vars that are not picked-up correctly, `import.meta.env.VITE_GOOGLEAPIKEY and import.meta.env.VITE_IONACCESSTOKEN`. 
This example can eventually be left as-is and a stripped version created from it, or only the stripped-off version can be retained in the end. The readme added to the src/r3f dir redefines most of the components showcased in this example, hence we can probably simplify the r3fDemo.js a lot with no regrets. I also added drei to optionalDependencies for showcasing some of the r3f components. Hope this helps getting closer to merging in master!

.gitignore
```diff
@@ -62,3 +62,4 @@ typings/
 
 example/local-data
 example/dev-bundle
+example/bundle
```
gkjohnson, line 65: We should remove this - this is necessary for the examples build process since this will get committed and pushed to a separate branch.

vite.config.js
```diff
@@ -1,11 +1,13 @@
 import { searchForWorkspaceRoot } from 'vite';
 import fs from 'fs';
+import react from '@vitejs/plugin-react';
 
 export default {
 
 	root: './example/',
 	base: '',
 	build: {
+		minify: false,
```
gkjohnson, line 10: Why are we disabling minification?

vite.config.js
```diff
@@ -21,6 +23,11 @@ export default {
 				searchForWorkspaceRoot( process.cwd() ),
 			],
 		},
-	}
+	},
+	plugins: [
+		react( {
+			jsxRuntime: 'automatic',
+		} )
```
gkjohnson, line 30: It looks like 'automatic' is the default so we can remove the option.

example/r3fDemo.jsx
```diff
@@ -0,0 +1,286 @@
...
+      // compressUvs={false} uvType={Int8Array}
+      // compressPosition={false} positionType={Int16Array}
+    />
+  </>
+}
+
+const URL = 'https://raw.githubusercontent.com/NASA-AMMOS/3DTilesSampleData/master/msl-dingo-gap/0528_0260184_to_s64o256_colorize/0528_0260184_to_s64o256_colorize/0528_0260184_to_s64o256_colorize_tileset.json';
+
+const googleApiKey_ = import.meta.env.VITE_IONACCESSTOKEN || 'put-your-api-key-here' 
+const ionAccessToken_ = import.meta.env.VITE_IONACCESSTOKEN || 'put-your-api-key-here'
```
gkjohnson, line 204: I'd prefer to use local storage for storing these fields, as is done [in the other examples](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/example/ionExample.js#L18). Setting env variables is fairly opaque even for development, I think.

package.json
```diff
@@ -61,5 +61,12 @@
   },
   "peerDependencies": {
     "three": ">=0.123.0"
+  },
+  "optionalDependencies": {
+    "@react-three/drei": "^9.114.3",
```
gkjohnson, line 66: Drei is only used in the examples and isn't used in any of the published components so this can be moved to "devDependencies", instead.

gkjohnson (commented):
Thanks! A last couple changes and we can merge this. I'll make a few changes to the examples in another PR after this

vite.config.js
jo-chemla: Sorry this was to troubleshoot the lema controls error, caused by undefined value for access-keys

example/r3fDemo.jsx
jo-chemla: Leva has no [persistence](https://github.com/pmndrs/leva/issues/113) to localStorage yet - multiple threads on global store/state integration listed [here](https://github.com/pmndrs/leva/issues/295) - but I can use onChange to persist individual values to localStorage, and setStore initially. Will try that right now.

jo-chemla: Alright, just added localStorage support for google and cesium api keys! Note if we want automatic localStorage support with every leva controls (which is not part of this pr), I've also added a snippet [there](https://github.com/pmndrs/leva/issues/295#issuecomment-2404777836)

jo-chemla:
Alright thanks again for all this feedback! This is ready to merge from my point of view!

example/r3fDemo.jsx
gkjohnson: That's okay - the local storage saving is mostly useful for the access tokens for viewing the data.

gkjohnson:
Last couple things:
- The package-lock conflict needs to be resolved (merge in r3f-dev).
- It looks like the "leva" package isn't defined in the package.json causing `npm start` to fail.
- I'm seeing these errors in the console:

image

I don't see them in the main r3f-dev branch. It looks like the demo is trying to add the google and cesium ion plugins when they're not needed? I can't load any ion data sets or anything even when pasting my ion token in. Is there something I'm missing for using the example?

jo-chemla:
Alright, just bring another batch of changes: 
 - resolved the package-lock conflict by removing node-modules and reinstalling `npm i`
 - I added leva as a devDependency, not sure when I removed it but this was not intentional. 
 - indeed I was loading the Cesium and Google renderer components even when api-key were undefined or default-string `"put-your-api-key-here"`, so I modified the component to handle that case
 - Also, the leva control for api keys required another prop, `transient: false` in order for these api variables to not be undefined. 
 - The last error I saw was a last `TypeError: tiles is null`. I managed to get around it by adding one last `tiles === null` check in useFrame on update the resolution for the camera

jo-chemla:
Demo video

jo-chemla:
Hi again, just added a mini example with only one standard renderer and default r3f box mesh + a few drei components. Since the google and cesium ion renderers have been added to the r3f readme, they might stay only in the complex r3f example.

gkjohnson:
> PS: I'd like to release a small post on socials (probably twitter) to announce that r3f integration

You can go ahead and post something - I'll announce once everything is ready and merged into master.

> added a mini example with only one standard renderer and default r3f box mesh + a few drei components.

Thanks - I'll merge this and will probably reorganize a few things over the next week while getting it ready for release.

PR #803, opened 2024-10-13
Remove unused react component files

gkjohnson:
- Removes the `src/r3f/R3F3DTilesRenderer.jsx` and associated `utilities.js` files
- Move src components to `src/r3f/components` directory

cc @jo-chemla


jo-chemla:
Thanks! Sorry I forgot these when committing the mini example

PR #812, opened 2024-10-17
`TilesRendererBase` `url` constructor parameter should be optional

RodrigoHamuy:
I believe `TilesRendererBase` constructor `url` parameter should be optional, no?

Updated the Google examples so one doesn't get the `GooglePhotorealisticTilesRenderer: Class has been deprecated. Use "TilesRenderer" with "GoogleCloudAuthPlugin" instead.` warning.

That's how I noticed that it needs to be optional, as otherwise I can't follow the deprecation warning suggestions without getting TS errors because I didn't pass a `url` to `TileRenderer` constructor.


gkjohnson:
Looks right! Thank you

Issue #818, opened 2024-10-20
Consider an "3d-tiles-renderer/addons" or "3d-tiles-renderer/plugins"-style export for in-progress or non-core functionality


gkjohnson:
Ie this is how some GLTF extensions or plugins in the examples foldre can be distributed while keeping the core folders more organized.

**Addons**
- Controls
- CameraTransitionManager
- Debug helpers (instead of in plugins)?

OR just move CameraTranstionManager to core, for now

gkjohnson:
Perhaps other "utilities" like the camera transition manager can go here, as well?

PR #839, opened 2024-11-03
Rename imports for subpackages

gkjohnson:
Related to #818


gkjohnson:
Might need to wait until there are no circular exports

PR #845, opened 2024-11-20
Fix errors in the PNTS Draco decoding docs

jsulli:
The PNTS decoding docs reference the wrong decoder as well as the wrong loader. Spent a bit of time hair pulling and hunting down confusing errors till I realized this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,12 +170,13 @@ scene.add( tilesRenderer2.group );
 
 ## Adding DRACO Decompression Support
 
-Adding support for DRACO decompression within the GLTF files that are transported in B3DM and I3DM formats. The same approach can be used to add support for KTX2 and DDS textures. Note that GLTF decoding and PNTS decoding require separate decoders.
+Adding support for DRACO decompression within the GLTF files that are transported in B3DM and I3DM formats. The same approach can be used to add support for KTX2 and DDS textures.
 
 ```js
 
 // Note the DRACO compression files need to be supplied via an explicit source.
 // We use unpkg here but in practice should be provided by the application.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -170,7 +170,7 @@ scene.add( tilesRenderer2.group );
 
 ## Adding DRACO Decompression Support
 
-Adding support for DRACO decompression within the GLTF files that are transported in B3DM and I3DM formats. The same approach can be used to add support for KTX2 and DDS textures.
+Adding support for DRACO decompression within the GLTF files that are transported in B3DM and I3DM formats. The same approach can be used to add support for KTX2 and DDS textures. Note that GLTF decoding and PNTS decoding require separate decoders.
```
gkjohnson, line 173: I'm not sure what this added line means. Can it be removed or clarified?

gkjohnson (commented):
Thanks! Looks like primarily a typo in the PNTS example snippet.

README.md
jsulli: In the current docs, the selected draco decoder from unpkg uses the GLTF branch, which *does not work* for compressed PNTS files. This is a really easy detail to miss IMO, I added this line to clarify that. I'll reword things

https://github.com/mrdoob/three.js/tree/dev/examples/jsm/libs/draco
vs
https://github.com/mrdoob/three.js/tree/dev/examples/jsm/libs/draco/gltf

jsulli:
Hopefully this language is clearer

README.md
gkjohnson: Oh I see now... how annoying. I'll move this information down to the section below where the PNTS draco loader is being discussed.

PR #865, opened 2024-12-12
Typing events tile renderer

sguimmara:
This PR adds types for events raised by `TilesRenderer`.

Note: the various methods for the event dispatcher are copy-pasted from the `EventDispatcher` type in `@types/three`. If you know a way to avoid duplicating the methods, I'd be happy to use it instead.

Note 2: I also replaced the `Number` and `Boolean` types with their primitive equivalent `number` and `boolean` (lowercase), as [recommended by Typescript](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#the-primitives-string-number-and-boolean):

> The type names String, Number, and Boolean (starting with capital letters) are legal, but refer to some special built-in types that will very rarely appear in your code. Always use string, number, or boolean for types.

### DIFF
--- a/tsconfig.test.json
+++ b/tsconfig.test.json
@@ -1,0 +1,10 @@
+{
+	"extends": "./tsconfig.json",
+	"compilerOptions": {
+		"noEmit": true,
+	},
+    "include": [
+		// Type-based test files to check API
+        "test/types/"
+    ]
+}
--- a/jest.config.json
+++ b/jest.config.json
@@ -1,1 +1,3 @@
-{}
+{
+	"testPathIgnorePatterns": ["types"]
+}
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "start": "vite --config ./vite.config.js",
     "build-examples": "vite build --config ./vite.config.js",
     "lint": "eslint \"src/**/*.{js,jsx,ts}\" \"test/**/*.js\" \"example/*.{js,jsx}\" && tsc -p tsconfig.json --noEmit",
-    "test": "jest"
+    "test": "tsc --project tsconfig.test.json && jest"
   },
   "repository": {
     "type": "git",
--- a/src/plugins/three/ReorientationPlugin.d.ts
+++ b/src/plugins/three/ReorientationPlugin.d.ts
@@ -1,4 +1,4 @@
-export class GLTFExtensionsPlugin {
+export class ReorientationPlugin {
 
 	constructor( options: {
 		up: '+x' | '-x' | '+y' | '-y' | '+z' | '-z',
--- a/src/three/TilesRenderer.d.ts
+++ b/src/three/TilesRenderer.d.ts
@@ -1,10 +1,25 @@
-import { Box3, Camera, Vector2, Matrix4, WebGLRenderer, Object3D, LoadingManager, Sphere } from 'three';
+import { Box3, Camera, Vector2, Matrix4, WebGLRenderer, Object3D, LoadingManager, Sphere, EventListener, EventDispatcher, BaseEvent } from 'three';
 import { Tile } from '../base/Tile';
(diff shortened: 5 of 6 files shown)


gkjohnson:
Thanks again for the type additions - I'm not really familiar with typescript so these template types look _really_ complicated to me. If you have brief explanation that would be great but either way if types like this are going to be added to the project I'm going to have to ask for some commitment to helping to answer questions or maintain them. Is that something you or the Giro3D project can help with?

Also the event system isn't just limited to the events dispatched from within the class. It's possible for end users or plugins to add their own event types for their own uses:

```js
tiles.dispatchEvent( { type: 'custom-tiles-event' } );
```

The TilesFadePlugin does this, for example. I'm not sure if it's possible to add this kind of flexibility to the event system as well (basically allow for an "any" event).

sguimmara:
> Thanks again for the type additions - I'm not really familiar with typescript so these template types look _really_ complicated to me. If you have brief explanation that would be great but either way if types like this are going to be added to the project I'm going to have to ask for some commitment to helping to answer questions or maintain them. Is that something you or the Giro3D project can help with?
> 
> Also the event system isn't just limited to the events dispatched from within the class. It's possible for end users or plugins to add their own event types for their own uses:
> 
> ```js
> tiles.dispatchEvent( { type: 'custom-tiles-event' } );
> ```
> 
> The TilesFadePlugin does this, for example. I'm not sure if it's possible to add this kind of flexibility to the event system as well (basically allow for an "any" event).

You're right, there might be a way of doing it with arbitrary plugins, by unioning the event types of all plugins + the base object. I'll see if I can find a way that is ergonomic enough. Otherwise, I'll abandon this PR.

I'll also open a separate PR for the removal of wrapper types (i.e `Number`-> `number`), because it makes sense to do it once for all the typescript definition files rather than one file.

gkjohnson:
> You're right, there might be a way of doing it with arbitrary plugins, by unioning the event types of all plugins + the base object.

I don't necessarily mean that we need to include all the plugin events explicitly as much as the dispatch and event listen functions need to be flexible enough to allow for users to pass in event types that are not included in the d.ts file since end users may make their own custom events in their own plugins, as well.

I appreciate the work on this!

sguimmara:
@gkjohnson  I updated the PR with the following changes:
- added a way to test the API files (`*.d.ts`) without actually running test (just using the typechecker). By calling the typescript compiler on the test files, we can detect API breaking changes (for example, an event argument that was removed or renamed, or its type changed)
- fixed incorrect class name for the `ReorientationPlugin`

sguimmara:
> > You're right, there might be a way of doing it with arbitrary plugins, by unioning the event types of all plugins + the base object.
> 
> I don't necessarily mean that we need to include all the plugin events explicitly as much as the dispatch and event listen functions need to be flexible enough to allow for users to pass in event types that are not included in the d.ts file since end users may make their own custom events in their own plugins, as well.
> 
> I appreciate the work on this!

The event dispatcher supports both typed and unknown events, that is why there are two versions of event dispatcher methods: one for types that are known (and the type signature of their event handler known as well), and the events that are unknown, such those dispatched by plugins. For the latter, the event listener is not typechecked.

tsconfig.test.json
```diff
@@ -0,0 +1,10 @@
+{
+	"extends": "./tsconfig.json",
+	"compilerOptions": {
+		"noEmit": true,
+	},
+    "include": [
+		// Type-based test files to check API
+        "test/types/"
+    ]
```
gkjohnson, line 9: Nit: looks like this is invalid json and indentation looks off

gkjohnson (commented):
Okay, thanks! I'll merge this but I'll have to say I don't fully understand the typing here so as I've mentioned before I'll have to leave it to others to properly maintain some of these types.

PR #868, opened 2024-12-14
eslint: add the `no-wrapper-object-types` rule

sguimmara:
This [rule](https://typescript-eslint.io/rules/no-wrapper-object-types) disallow so-called wrapper types in favor of primitives types: `boolean` instead of `Boolean`, `number` instead of `Number`, etc.

The official Typescript documentation [says](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#the-primitives-string-number-and-boolean):

> The type names String, Number, and Boolean (starting with capital letters) are legal, but refer to some special built-in types that will very rarely appear in your code. Always use string, number, or boolean for types.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -43,7 +43,8 @@
 		"rules": {
 			"no-unused-vars": [ "error", { "args": "none" } ],
 			"indent": [ "error", "tab" ],
-			"no-dupe-class-members": "off"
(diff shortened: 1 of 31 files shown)


gkjohnson:
Great thank you! This is a good rule to have - I wasn't aware of the diffrence.

PR #869, opened 2024-12-14
docs: add missing tiles-load-start/end events documentation

lxxorz:
docs: add missing tiles-load-start/end events documentation

Add documentation for two missing events:
- tiles-load-start: fired when tiles start loading
- tiles-load-end: fired when all tiles finish loading

These events are implemented in the code but were not documented in the README.


lxxorz:
Related PR #602.

gkjohnson:
Great, thank you!

PR #871, opened 2024-12-16
fix(GLTFCesiumRTCExtension.d.ts): implements THREE.js's GLTFLoaderPlugin

sguimmara:
This typing is necessary to register this extension on the three.js GLTFLoader.

### DIFF
--- a/src/plugins/three/gltf/GLTFCesiumRTCExtension.d.ts
+++ b/src/plugins/three/gltf/GLTFCesiumRTCExtension.d.ts
@@ -1,4 +1,6 @@
-export class GLTFCesiumRTCExtension {
+import { GLTFLoaderPlugin } from 'three/examples/jsm/loaders/GLTFLoader.js';
+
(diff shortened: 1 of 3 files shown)


gkjohnson:
Thanks! Can you add this on the other GLTF Extensions, as well? (MeshFeaturesExtension, StructuralMetadataExtension)

sguimmara:
> StructuralMetadataExtension

Yes, updated the other files as well :)

While looking at the `GLTFLoaderPlugin` API, it seems very similar to what you did for the `TilesRenderer`: an object that exposes optional methods that are called by the invoke plugin mechanism. It might be a good idea to add types for those plugins as well, what do you think ? 

Something like that (truncated only for the `fetchData` method):

```ts
export interface TilesRendererPlugin {
        ...
	fetchData?: ( url: RequestInfo, options: RequestInit ) => Promise<Response>;
        ...
}

```

gkjohnson:
Thanks! I'll respond to the tiles plugin types proposal in the other PR.

PR #874, opened 2024-12-18
Add "UnloadTilesPlugin" to save GPU memory

### LINKED ISSUE #873: Test deleting geometry on visibility change rather than toggling visibility
This would require geometry and textures to reupload but would help save GPU resources.

For BatchedMesh this would limit the amount of slots needed.

### LINKED ISSUE #875: Plugins: Add a "priority" field to ensure plugins are sorted in a compatible way
Otherwise fall back to insertion order

gkjohnson:
Fix #873 
Fix #875 

- Add priority support to plugins
- TilesCompression and BatchedMesh plugin will always run first
- Prevent empty groups from being added for each tile when using BatchedMeshPlugin
- Add "UnloadTilesPlugin"
- Add support for "setTileActive" and "setTileVisible" plugin callbacks
- Add names to some plugins that were missing them

**TODO**
- Add "visibleTiles" and "activeTiles" to base tiles renderer class
- Consider removing "setTileActive" callback
- Expose byte counting function for plugin use

**Upcoming PR**
- Make this work with fade tiles
  - Override the setTileVisible to only fire when the tile fades in and out for the first time
- Add to demos, move to core plugins directory

### DIFF
--- a/example/src/plugins/UnloadTilesPlugin.js
+++ b/example/src/plugins/UnloadTilesPlugin.js
@@ -34,7 +34,7 @@ export class UnloadTilesPlugin {
 
 				if ( ! visible ) {
 
-					tiles.invokeOnePlugin( plugin => plugin.unloadTileFromGPU && plugin.unloadTileFromGPU( tile, scene ) );
+					tiles.invokeOnePlugin( plugin => plugin.unloadTileFromGPU && plugin.unloadTileFromGPU( scene, tile ) );
 
 				}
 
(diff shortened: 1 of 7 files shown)


sguimmara:
Would you consider adding some sort of configurable timer to unload the tiles after some time ? Often, a tile that has been hidden because the user quickly moved the camera can be re-displayed a few milliseconds/seconds after being hidden. It's useful that this tile can be displayed right away.

gkjohnson:
> Would you consider adding some sort of configurable timer to unload the tiles after some time ? 

It's a good idea - I'll make another issue for it and focus on getting the basics working in this PR. You could imagine other metrics, as well, such as only removing items from the GPU once a certain memory threshold is passed. Something like a smaller lru cache just for GPU memory.

Issue #484, opened 2024-02-19
DebugTilesPlugin: improve "init extremes" performance


gkjohnson:
When a new child tileset is loaded the whole hierarchy is loaded which can take 40+ ms in some cases especially with complex tilesets like google tiles, causing a jerkiness when new tiles are loaded in.

Options:
- Derive extremes only from currently visible tilesets
- Take the max of the current extreme and any newly added tileset

sguimmara:
I didn't see this issue, but it might be obsolete now with #885

gkjohnson:
I think it is!

PR #885, opened 2024-12-20
DebugTilesPlugin: optimize performance and make it toggleable

### LINKED ISSUE #647: DebugTilesPlugin: Add option for enabling / disabling plugin to toggle performance
The traversal and updates of the fields and objects in debug plugin can cause some hiccups and slowness. It would be nice to be able to disable the function of the plugin so it doesn't impact performance. Things like tile updates can be deferred until their needed in the update function which can be paused for performance.

sguimmara:
This PR optimizes the `DebugTilesPlugin` by reducing its overhead, an makes it toggleable.

fix #647 

## Tree traversal optimizations

This PR optimizes (some) tree traversals by using a stack-based approach, rather than a recursive approach, that reduces overhead of recursion.

`traverseSet()` is modified from a recursive approach to a stack-based approach. ~`toggleTiles()` now uses this version.~

Note: other functions in `traverseFunctions.js` might also benefit from using this approach, but the migration is not as easy as for `toggleTiles()`, as the recursion logic is entangled with the callback logic.

## Disable/Enable the `DebugTilesPlugin` dynamically

Add the `enabled` property on the plugin. When toggled on, the plugin initializes its state from the current tile hierarchy. When toggled off, the existing helpers are removed, and all event listeners do nothing (except for `'dispose-model'`).

### DIFF
--- a/src/base/traverseFunctions.js
+++ b/src/base/traverseFunctions.js
@@ -134,32 +134,57 @@ function canTraverse( tile, renderer ) {
 
 }
 
-// Helper function for recursively traversing a tile set. If `beforeCb` returns `true` then the
+// Helper function for traversing a tile set. If `beforeCb` returns `true` then the
 // traversal will end early.
-export function traverseSet( tile, beforeCb = null, afterCb = null, parent = null, depth = 0 ) {
+export function traverseSet( tile, beforeCb = null, afterCb = null ) {
 
-	if ( beforeCb && beforeCb( tile, parent, depth ) ) {
+	const stack = [];
+
+	// A stack-based, depth-first traversal, storing
+	// triplets (tile, parent, depth) in the stack array.
+
+	stack.push( tile );
+	stack.push( null );
+	stack.push( 0 );
+
+	while ( stack.length > 0 ) {
+
+		const depth = stack.pop();
+		const parent = stack.pop();
+		const tile = stack.pop();
+
+		if ( beforeCb && beforeCb( tile, parent, depth ) ) {
+
+			if ( afterCb ) {
+
+				afterCb( tile, parent, depth );
+
+			}
+
+			return;
+
+		}
+
+		const children = tile.children;
+
+		// Children might be undefined if the tile has not been preprocessed yet
+		if ( children ) {
+
+			for ( let i = children.length - 1; i >= 0; i -- ) {
+
+				stack.push( children[ i ] );
+				stack.push( tile );
+				stack.push( depth + 1 );
+
+			}
+
+		}
 
 		if ( afterCb ) {
 
 			afterCb( tile, parent, depth );
 
 		}
-
-		return;
-
-	}
-
-	const children = tile.children;
-	for ( let i = 0, l = children.length; i < l; i ++ ) {
-
-		traverseSet( children[ i ], beforeCb, afterCb, tile, depth + 1 );
-
-	}
-
-	if ( afterCb ) {
-
-		afterCb( tile, parent, depth );
 
 	}
 
--- a/src/plugins/three/DebugTilesPlugin.js
+++ b/src/plugins/three/DebugTilesPlugin.js
@@ -60,6 +60,8 @@ export class DebugTilesPlugin {
 		this.name = 'DEBUG_TILES_PLUGIN';
 		this.tiles = null;
 
+		this._enabled = true;
+
 		this.extremeDebugDepth = - 1;
 		this.extremeDebugError = - 1;
 		this.boxGroup = null;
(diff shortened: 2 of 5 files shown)


sguimmara:
Before continuing further, I would like your opinion on this @gkjohnson.

The biggest performance hit of `DebugTilesPlugin` is the tree traversal(s) in `_initExtremes()`, because the traversals preprocess all tiles in the tileset, which is not necessary for the needs of `_initExtremes()`.

In one of [Giro3D examples](https://giro3d.org/latest/examples/colorized_pointcloud.htm), we have a massive tileset with a lot of nested tilesets, which means that the preprocessing happens a lot, rather than once (which would be the case with a single root `tileset.json`).

sguimmara:
I actually ended up implementing the toggling itself, so it's ready for review

src/base/traverseFunctions.js
```diff
@@ -402,64 +427,62 @@ export function markVisibleTiles( tile, renderer ) {
 }
 
 // Final traverse to toggle tile visibility.
-export function toggleTiles( tile, renderer ) {
+export function toggleTiles( root, renderer ) {
```
gkjohnson, line 430: This change will result in a different and more complex traversal than the previous one. The current implementation in master [conditionally traverses the children](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/base/traverseFunctions.js#L408) so it will stop early in the hierarchy. These tile traversal functions are complex and can be a bit difficult to test and benchmark - I'm happy to make changes to them but it might be best to keep them in a dedicated PR so they're easier to analyze and test individually.

The stack-based `traverseSet` update looks like a good change, though!

test/traverseFunctions.test.js
```diff
@@ -0,0 +1,101 @@
+import { traverseSet } from '../src/base/traverseFunctions.js';
+
+describe( 'traverseSet', () => {
```
gkjohnson, line 3: Thank for adding this 😁

src/plugins/three/DebugTilesPlugin.js
```diff
@@ -108,31 +138,48 @@ export class DebugTilesPlugin {
...
+				this._onUpdateAfter();
+
+			}
 
 		};
 
 		this._onTileVisibilityChangeCB = ( { scene, tile, visible } ) => {
 
-			this._onTileVisibilityChange( tile, visible );
+			if ( this._enabled ) {
```
gkjohnson, line 178: nit: lets use the `enabled` getter here rather than the `_enabled` value.

src/plugins/three/DebugTilesPlugin.js
```diff
@@ -214,17 +268,19 @@ export class DebugTilesPlugin {
...
 
-		} );
+		}
 
+		// initialize the extreme values of the hierarchy
+		let maxDepth = - 1;
 		let maxError = - 1;
-		this.tiles.traverse( tile => {
 
+		traverseSet( this.tiles.root, null, ( tile, parent, depth ) => {
```
gkjohnson, line 281: Why use `traverseSet` rather than `tiles.traverse`?

src/plugins/three/DebugTilesPlugin.js
```diff
@@ -83,6 +86,33 @@ export class DebugTilesPlugin {
...
+			this._enabled = v;
+
+			if ( this._enabled ) {
+
+				this._initExtremes();
+				this._initializeExistingTiles();
+
+			} else {
+
+				this._removeHelpers();
```
gkjohnson, line 108: Material changes are done based on the `colorMode` in the `_onUpdateAfter` function so we'll also want to make sure we revert the material changes here. This is done in the `dispose` function like so:

```js
// reset all materials
this.colorMode = NONE;
this._onUpdateAfter();
```

Now that I'm looking at it I'm wondering if we should just call `this.dispose` when enabled is toggled to "false" and then call `init` again when it's toggled to "true". Then the plugin can clean up after itself in a consistent way, unhook any events so there's no added overhead. That may be a smaller change.

gkjohnson (commented):
Amazing! Thank you. I added a few comments but this should make the plugin a lot nicer to use.

src/plugins/three/DebugTilesPlugin.js
```diff
@@ -108,31 +141,48 @@ export class DebugTilesPlugin {
...
+
+				this._onLoadModel( scene, tile );
+
+			}
 
 		};
 
 		this._onDisposeModelCB = ( { tile } ) => {
 
+			// Note that we want to cleanup memory even if the plugin is disabled.
```
gkjohnson, line 164: This comment seems out of date - since we remove the events when the plugin is disabled this function will no longer be called.

gkjohnson (commented):
Thanks! Last couple comments and then this looks ready to merge.

sguimmara:
@gkjohnson  seems it also solves #484

gkjohnson:
Excellent, thank you! I'll update the docs to add the field.

PR #892, opened 2024-12-22
style(.eslintrc.json): disable the no-undef rule for typescript

sguimmara:
This rule conflicts with TypeScript's own typechecker in some cases, for example the RequestInit DOM type.

This rule should be [disabled for TypeScript files](https://eslint.org/docs/latest/rules/no-undef#handled_by_typescript):
> It is safe to disable this rule when using TypeScript because TypeScript's compiler enforces this check.


gkjohnson:
Thank you!

Issue #844, opened 2024-11-18
Implicit Tiling (v1.1) missing subtrees / wrong lookup?


phoenixbf:
Hi all
I'm testing tilesets using the new V1.1 tiler by cesium ion and implicit tiling plugin (octree).
Everything basically works ok, except towards octree leaves (higher-res tiles) there appears random misses (eg. `"TilesRenderer : Failed to load tile at url "subtrees/5/8/14/0.subtree"`) on *.subtree files, causing indeed some holes in the mesh on close inspection as they refine.

I initially thought it was something related to a corrupted zip downloaded from ion, but instead it seems the last bit of the subtree path is composed differently (e.g. instead of "2.subtree", a "0.subtree" is fetched, and so on... while the subfolders path is correct on both cesium viewer and 3DTilesRenderer).
As if the last part of the path has wrong subtree filename computed (maybe morton codes)?

This bug can be replicated with any sufficiently complex obj converted into tileset using v1.1 ion tiler.

I'll try to provide more details while I keep testing, and have a more indepth inspection ad the code if I find something

gkjohnson:
> This bug can be replicated with any sufficiently complex obj converted into tileset using v1.1 ion tiler.

Can you please provide a specific tile set that exhibits the issue? Also please follow the [bug report](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/new?assignees=&labels=bug&projects=&template=bug_report.md&title=). Trying to accurately recreate your issue from just a description and no tile set or images is a lot of effort and guess work.

cc @AnthonyGlt

phoenixbf:
Indeed. We are preparing a dataset that exhibits the issue

phoenixbf:
Ok, so here is a sample dataset that exhibits the issue (the url can be referenced directly by other viewers)
https://explore100.cnr.it/collections/100/models/v1.1/tileset.json

here is also a reference of issue in 3DTRenderer
![photo_2024-11-19_22-55-58](https://github.com/user-attachments/assets/c4256b88-ac23-4dec-9808-b7a5421191b9)

and the same tileset in Cesium ion viewer:
![photo_2024-11-19_22-56-06](https://github.com/user-attachments/assets/dc51c1ed-6141-496a-a4ff-64d76aa31def)

[...]

PR #893, opened 2024-12-23
DebugTilesPlugin: add the displayParentBounds option (#730)

### LINKED ISSUE #730: DebugTilesPlugin: Add support for displaying parent bounds

sguimmara:
This PR fixes #730.

Adds the `.displayParentBounds` property to display the bounds of tiles between the root tile and the first visible tile.

The helpers for the intermediate bounds use ~a thicker line and~ semi-transparent materials to help distinguishing them from the "leaf" tiles.

image

~Note: not all GPUs support thick lines (see the [comment in the three.js doc](https://threejs.org/docs/index.html?q=box3help#api/en/materials/LineBasicMaterial.linewidth)), so on some platforms the line will remain 1-pixel thick. The alternative would be to use different materials, but that would increase complexity and maintenance.~

This uses a reference counting mechanism as suggested in #730.

### DIFF
--- a/src/plugins/three/DebugTilesPlugin.js
+++ b/src/plugins/three/DebugTilesPlugin.js
@@ -1,12 +1,13 @@
 import { Box3Helper, Group, MeshStandardMaterial, PointsMaterial, Sphere, Color } from 'three';
 import { SphereHelper } from './objects/SphereHelper.js';
 import { EllipsoidRegionLineHelper } from './objects/EllipsoidRegionHelper.js';
-import { traverseSet } from '../../base/traverseFunctions.js';
+import { traverseAncestors, traverseSet } from '../../base/traverseFunctions.js';
 
 const ORIGINAL_MATERIAL = Symbol( 'ORIGINAL_MATERIAL' );
 const HAS_RANDOM_COLOR = Symbol( 'HAS_RANDOM_COLOR' );
 const HAS_RANDOM_NODE_COLOR = Symbol( 'HAS_RANDOM_NODE_COLOR' );
 const LOAD_TIME = Symbol( 'LOAD_TIME' );
+const PARENT_BOUND_REF_COUNT = Symbol( 'PARENT_BOUND_REF_COUNT' );
 
 const _sphere = /* @__PURE__ */ new Sphere();
 const emptyRaycast = () => {};
@@ -46,6 +47,7 @@ export class DebugTilesPlugin {
 	constructor( options ) {
 
 		options = {
+			displayParentBounds: false,
 			displayBoxBounds: false,
 			displaySphereBounds: false,
 			displayRegionBounds: false,
@@ -69,6 +71,7 @@ export class DebugTilesPlugin {
 		this.regionGroup = null;
 
 		// options
+		this._displayParentBounds = options.displayParentBounds;
 		this.displayBoxBounds = options.displayBoxBounds;
 		this.displaySphereBounds = options.displaySphereBounds;
 		this.displayRegionBounds = options.displayRegionBounds;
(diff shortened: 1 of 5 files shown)


sguimmara:
> @gkjohnson the default example is not ideal to test it since the child bounds tightly fit in the parent bounds. Do you have an example where there is more loose space between parent/child bounds ?

Nevermind I just realized that we could pass arbitrary URLs in the example

src/plugins/three/DebugTilesPlugin.js
```diff
@@ -648,6 +669,113 @@ export class DebugTilesPlugin {
...
+
+		if ( tile.__visible || ! this.displayParentBounds ) {
+
+			material.linewidth = 1;
+			material.opacity = 1;
+
+		} else {
+
+			material.linewidth = 4;
+			material.opacity = 0.2;
```
gkjohnson, line 684: I'd prefer to avoid using line width so we're consistent between platforms - it's not working on my machine, for eample. As you say it looks like we'll need a better treatment for these lines, though, so they're more distinguishable. The random colors for each depth tier don't really help, either. But I think we can evolve this as it's actually used for debugging stuff - the logic for toggling parents is great to have added, though.

src/plugins/three/objects/EllipsoidRegionHelper.js
```diff
@@ -81,8 +81,12 @@ function getRegionGeometry( ellipsoidRegion ) {
...
 	}
 
-	// compute the vertex normals so we can get the edge normals
-	geometry.computeVertexNormals();
+	if ( computeNormals ) {
+
+		// compute the vertex normals so we can get the edge normals
+		geometry.computeVertexNormals();
+
+	}
```
gkjohnson, line 89: Can you explain this change?

gkjohnson (commented):
Thanks - a couple comments. I'm hoping this can be used to help understand what's happening in #844.

As expected it can look a bit messy but this is great progress

image

src/plugins/three/objects/EllipsoidRegionHelper.js
sguimmara: When profiling a tileset with region bounds, I noticed that the biggest performance hit was computing normals for ellipsoid helpers. However, the line-based variant of the helper do not need normals AFAIK.

src/plugins/three/DebugTilesPlugin.js
sguimmara: I removed the line width change.

sguimmara: On the top of my heads, here are a few ideas to help visualize the relationship between parent and child bounds:
- optionally connecting lines between the center of the parent bounds to the centers of child bounds, the result would look like some kind of "nervous system", that would help figuring the structure of the hierarchy
- using [materials that support dashes](https://threejs.org/docs/?q=dash#api/en/materials/LineDashedMaterial) and use dashed lines for intermediate bounds
- optionally, when hovering a tile (using raycasting), display only the bounds of the sub-tree that this tile belongs to, up to the root tile

gkjohnson: > optionally, when hovering a tile (using raycasting), display only the bounds of the sub-tree that this tile belongs to, up to the root tile

This one I think would probably be the most useful. I think the other visual treatments for the parent or child lines will be really noisy and a bit difficult to understand. I'll merge this for now and we can consider this in the future

gkjohnson:
Very nice work! Thanks for the addition

Issue #906, opened 2025-01-06
Granular progress tracking ?


sguimmara:
**Is your feature request related to a problem? Please describe.**

In Giro3D, we have an API to query the ["percentage of progress"](https://giro3d.org/latest/examples/tracking_progress.html) of a given object. The exact semantics depend on the object itself. For example, in the case of a tiled image layer, the progress is the number of downloaded tiles over the total amount of visible tiles.

To implement such an API for 3D Tiles-based entities, we need to know the "amount of total work" versus the "amount of work already done".

**Describe the solution you'd like**

For the `TilesRenderer`, this could be an API (typically an event) to know when the loading of a tile starts and an event to know when the loading has finished (this event already exists, `load-model`).

The simplest implementation would simply add the `load-model-start` event that would be raised at the very beginning of `parseTile()`.

**Describe alternatives you've considered**

I tried dabbling with the `.stats` property, but it does not provide the information I need.

gkjohnson:
> The simplest implementation would simply add the load-model-start event that would be raised at the very beginning of parseTile().

For a few reasons I don't think this is going to get you the load detail you want. For one it wouldn't include the download time or time that an item is queued. The time from the beginning to end of `parseTile` also may not even last more than 1 frame so it will likely flicker between 0 and 100 pretty quickly. And including things like time queued for parsing and downloading will mean accounting for tile cancellation, etc.

I think using the lru cache, download, and parse queue fields can be used to calculate the load percent, though:

```js
const loading =
    downloadQueue.itemList.length + downloadQueue.currJobs +
    parseQueue.itemList.length + parseQueue.currJobs;

// or

const loading = stats.downloading + stats.parsing;

const percent = 1.0 - ( loading / lruCache.itemList.length );

console.log( `Percent loaded: ${ 100 * percent }%` );
```

`ParseQueue.itemList.length` lists the total number of jobs that are currently waiting to run. `ParseQueue.currJobs` lists the current jobs that are currently running (and have therefore been removed from the item list queue). And `LRUCache.itemList` lists the total number of tiles that have either been loaded or have been queued to load. If this looks good we can add something like a `loadPercent` member field to `TilesRendererBase` to expose this more directly. Alternatively we could list the number of tiles actively loading and the number of tiles in the lru cache so it can be calculated.

[...]

PR #907, opened 2025-01-06
feat(TilesRenderer): add the load-model-start event (#906)

sguimmara:
This PR adds the load-model-start event that is triggered at the very beginning of parseTile(), and is the symetrical equivalent of load-model.


gkjohnson:
Closing in favor of #920

PR #908, opened 2025-01-07
fix(GLTFExtensionsPlugin.d.ts): make all parameters optional

sguimmara:
This PR fixes the `.d.ts` files for various plugins so that the signature for optional parameters matches what the Javascript version accepts.


gkjohnson:
Thanks!

PR #920, opened 2025-01-11
Track the number of tiles loading since loading had finished previously

### LINKED ISSUE #906: Granular progress tracking ?

gkjohnson:
Fix #906

**TODO**
- ~Rename `inCacheBeforeReset` - `inCacheSinceLoad`?~
- Consider exposing convenience members / function - `loadProgress` / `loadProgressSinceCompletion`, `getLoadProgress( sinceLoad = true )`

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,14 +392,6 @@ If true then the `raycast` functions of the loaded tile objects are overriden to
 
 If you would like to manage raycasting against tiles yourself this behavior can be disabled if needed by setting `optizeRaycast` to false.
 
-### .preprocessURL
-
-```js
-preprocessURL = null : ( uri : string | URL ) => string | URL;
-```
-
-Function to preprocess the url for each individual tile geometry or child tile set to be loaded. If null then the url is used directly.
-
 ### .lruCache
 
 ```js
@@ -445,6 +437,14 @@ manager : LoadingManager
 ```
 
 The manager used when loading tile geometry.
+
+### .loadProgress
+
+```js
+readOnly loadProgress : Number
+```
+
+Returns the total load progress between `[0, 1]`. Progress is measured since the last set of loading tiles completed.
 
 ### .constructor
 
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -108,7 +108,7 @@ export class TilesRendererBase {
 		this.fetchOptions = {};
 		this.plugins = [];
 		this.queuedTiles = [];
-		this.cachedSinceReset = new Set();
+		this.cachedSinceLoadComplete = new Set();
 
 		const lruCache = new LRUCache();
 		lruCache.unloadPriorityCallback = lruPriorityCallback;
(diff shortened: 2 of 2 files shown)


gkjohnson:
cc @sguimmara This should be what you're looking for regarding #906. Test it out and let me know what you think. Calculating the load percent since the last load completed should work like so:

```js
const active = tiles.stats.downloading + tiles.stats.parsing;
const total = tiles.stats.inCacheSinceReset;
console.log( ( 100 * ( 1.0 - active / total ) ).toFixed( 2 ) );
```

Thought `inCacheSinceReset` will reset to 0 once a load has completed so you'll have to check for that case to avoid dividing by 0.

sguimmara:
That seems to do the job ! Thanks. Just one slight change: if total is zero, the value is `NaN`. In my case, I consider that if `total` is zero, then the progress is 100%.

sguimmara:
I also updated `TilesRendererBase.d.ts` with the following additions:

```diff
+ export type Statistics = {
+	inCacheSinceLoad: number,
+	inCache: number,
+	parsing: number,
+	downloading: number,
+	failed: number,
+	inFrustum: number,
+	used: number,
+	active: number,
+	visible: number,
+ };

export class TilesRendererBase {

	readonly rootTileSet : object | null;
	readonly root : object | null;

+	stats: Statistics;
```

But that might not be a good idea to expose the stats if we use a method to query the progress instead

gkjohnson:
I'd like to expose this as a member but I'm struggling with naming. Perhaps both calculations don't need to be exposed and "loadPercent" is enough.

sguimmara:
> I'd like to expose this as a member but I'm struggling with naming. Perhaps both percentages don't need to be exposed and "loadPercent" is enough.

I wouldn't expose a percentage (as in the [0, 100] range), but a normalized value in the [0, 1] range though.

What about a `.getLoadingProgress()` method ?

gkjohnson:
> I wouldn't expose a percentage (as in the [0, 100] range), but a normalized value in the [0, 1] range though.

I plan to expose it as [0, 1]

> What about a .getLoadingProgress() method ?

Is this preferred to a `tiles.loadPercent` or `loadProgress` getter?

sguimmara:
> Is this preferred to a `tiles.loadPercent` or `loadProgress` getter?

Will `loadPercent` avoid `NaN` values ? If so, then let's use it directly.

PR #925, opened 2025-01-14
refactor(controls): add missing properties in declaration files

sguimmara:
@gkjohnson It's not very clear in the .js files which fields are internal and other are meant to be exposed as public APIs, so I'm not 100% sure. I looked into the various examples for fields that were used as a starting point.


gkjohnson:
Thanks!

PR #955, opened 2025-02-03
Fix DebugTilesPlugin updates with matrixAutoUpdate = false

jsulli:
Specifically, if the global default `Object3D.DEFAULT_MATRIX_AUTO_UPDATE` is disabled, debug objects from the `DebugTilesPlugin` do not update correctly. All that was missing was a `matrixWorld` update on the main container, and matrix updates on the helpers themselves (where applicable). 

There are several other places in this library that fail when matrixAutoUpdate is disabled, but they're relatively easy to fix application-side with the event callbacks already in place, but that wasn't feasible here. I may make some PR's for that in the future just to provide some bugfixes for applications that have matrix auto updates disabled for performance reasons.

I do not have tilesets that allow testing of boundingSphere and boundingRegion, so I just emulated the work I did for boundingBoxes - validation on those would be helpful.

### DIFF
--- a/src/plugins/three/DebugTilesPlugin.js
+++ b/src/plugins/three/DebugTilesPlugin.js
@@ -762,7 +762,9 @@ export class DebugTilesPlugin {
 			if ( boxHelperGroup ) {
 
 				boxGroup.add( boxHelperGroup );
+				boxGroup.updateMatrixWorld( true );
 				boxHelperGroup.updateMatrixWorld( true );
+				boxHelperGroup.children[ 0 ].updateMatrix();
 
 				this._updateHelperMaterial( tile, boxHelperGroup.children[ 0 ].material );
 
@@ -771,6 +773,7 @@ export class DebugTilesPlugin {
 			if ( sphereHelper ) {
 
 				sphereGroup.add( sphereHelper );
+				sphereGroup.updateMatrixWorld( true );
 				sphereHelper.updateMatrixWorld( true );
 
 				this._updateHelperMaterial( tile, sphereHelper.material );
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thanks for the contribution! 

> Specifically, if the global default Object3D.DEFAULT_MATRIX_AUTO_UPDATE is disabled

I'm generally not a fan at all of this flag in three.js because it changes otherwise reliable behavior out from under other libraries and dependencies, causing problems as we're seeing here. And supporting both code paths just adds more code to test and maintain. The matrix updates for tiles renderer are also already highly optimized such that no matrix operations are performed unless transforms changed so there's no need to adjust the auto update fields on the tiles renderer meshes.

The solution you've proposed also strictly adds more matrix operations for the common use case (not modifying these flags). My recommendation generally is to only modify the auto update flags for the three.js objects your app is specifically controlling and not change the settings that libraries depend on.

> they're relatively easy to fix application-side with the event callbacks already in place, but that wasn't feasible here.

If you're going to continue using the globally modified defaults you can detect addition / removal of elements from the `DebugTilesRenderer.boxGroup` (using the three.js `childadded` event) or checking for and updating the box transforms in the `tile-visibility-change` callback since they are created if necessary in that callback.

Let me know if that works for you!

jsulli:
Very reasonable points. I like to have them disabled because it can make a noticeable performance difference in heavy scenes, but it can be quite cumbersome and it's not well suited for a library such as this so I'm not surprised. 

I hadn't thought to just listen to the group for child events, sounds like a very feasible workaround, I'll give it a try. Thanks!

PR #963, opened 2025-02-09
Add ellipsoid support for TMS / XYZ tiles

gkjohnson:
**TODO**
- Remove limited map range support?
- Add remaining vertices for the caps


gkjohnson:
TMS geodetic:



XYZ mercator:

PR #981, opened 2025-02-17
Fix incorrect isMatrixType matching for metadata schema

Desplandis:
This PR fixes the issue where requesting properties of matrix type on structural metadata tables returned scalar values. This was due to an incorrect regexp for matrix type matching.


gkjohnson:
Great, thank you!

PR #990, opened 2025-02-19
Add MVT support

### LINKED ISSUE #982: Plugins: Add support for Mapbox Vector Tiles (MVT) format
### Resources
- [OpenStreetMap Vector Tiles wiki](https://wiki.openstreetmap.org/wiki/Vector_tiles)
- [MVT Spec](https://github.com/mapbox/vector-tile-spec/tree/master/2.1)
- [Another Vector Tiles Spec](https://openmaptiles.org/schema/)?

### Implementation

The TMS and XYZ Format Plugins implicitly support loading and generating of geometry (ellipsoid and planar) for each tile. We can add support for vector tiles via a separate plugin so it can be optionally bundled - particularly since it will require protobuf. 

We can start with the addition of an `MVTLoader` (a la the other the other tile format loaders) that loads the MVT file in a planar format. Then next steps can involve integrating it into the existing TMS / XYZ loaders:

```js
fetch( './path/to/file.mvt' )
  .then( res => res.arrayBuffer() )
  .then( buffer => {

    // returns a three.js mesh based on vector paths
    return new MVTLoader().parse( buffer );

  } );
```

The MVTLoader can take advantage of three.js' [Shape](https://threejs.org/docs/?q=shape#api/en/extras/core/Shape), [ShapePath](https://threejs.org/docs/?q=shape#api/en/extras/core/ShapePath), and [ShapeGeometry](https://threejs.org/docs/?q=shape#api/en/geometries/ShapeGeometry) classes. Three.js' [SVGLoader](https://threejs.org/docs/?q=shape#api/en/geometries/ShapeGeometry) may also be helpful to reference.

### Open Questions

The Shape classes will be able to generate shape paths for planar geometry easily but in order to support an ellipsoidal projection we need vertices _within_ these path shapes so the ellipsoid shape can be properly represented. We can see how things look before investigating how to solve the issue but I expect we will see artifacts when generating the globe geometry. One option is the further tessellate large triangle or edges after construction to some threshold before performing the ellipsoid projections.

_edit: This [constrained delaunay triangulation library](https://mikolalysenko.github.io/cdt2d/) could be used to triangulate the shapes with internal points. Alternatively the geometry could be "gridded" after triangulation to allow for transformation._

crubier:
Fixes #982

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,10 +9,12 @@
       "version": "0.4.5",
       "license": "Apache-2.0",
       "dependencies": {
+        "@mapbox/vector-tile": "^2.0.3",
         "@react-three/postprocessing": "^2.16.3",
         "@takram/three-atmosphere": "^0.6.0",
         "@takram/three-geospatial": "^0.0.1-alpha.4",
         "@takram/three-geospatial-effects": "^0.0.1-alpha.4",
+        "pbf": "^4.0.1",
         "postprocessing": "^6.36.4"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


gkjohnson:
Closing in favor of #1575

PR #1005, opened 2025-02-25
Initial setup of RegionTilesLoadingPlugin

### LINKED ISSUE #378: Provide more general spatial query functionality to get tiles localized to a region
This is a common problem for every Tile Renderer I tried.
I'm building a very basic no-code game/experience editor and wanted to implement some position based games where I can input latitude and longitude and I get the highest resolution tiles from Google Photorealistic tiles service.
On Unity and Unreal Google released Geospatial Creator but I'm using Three.js so I would like this 3D tiles renderer.

Unfortunately there are a lot of automations going on, requesting tiles seems to be related to camera pan/zoom, raycasting and so on.

It would be great to have some sort of API to request and position, I don't know, 4 siblings tiles close to location x,y,zoom without having to continuously requesting tiles as the camera moves.

I've also considered traversing the tileset but I don't know if **I'm missing something from the spec**, It's not working as expected.

This example uses real data from Google root node but if you check the console the first check returns true and the other four return false (at least one should be true).

I tried sphere.containsPoint and Box3.containsPoint

https://codesandbox.io/s/frosty-glitter-ghkcl2?file=/src/index.ts:86-768

### LINKED ISSUE #406: Allow for only loading tiles within a region
Ie don't load tiles outside of a specific volume. Similar to frustum culling but with an arbitrary shape and disallowing tiles outside the volume. Could also just allow for a custom "inView" or "inFrustum" callback.

Related to #378.

### LINKED ISSUE #1000: Tiles renderer: add support for raycasting to the highest LoD

The current raycaster only works on tiles within the current viewport.

Suppose I have a photo along with its position and orientation data. When I click on a point in the photo and attempt to find its corresponding location in the 3D world, the tiles in that area may be loaded at a low level of detail. As a result, the intersection is performed on a low-detail tile, leading to an inaccurate point calculation.

The function will be asynchronous, loading the most detailed tiles along the ray path and then performace the intersection test to ensure the most accurate intersection test.

A function that given a ray, and then the tiles on that ray be loaded would also works.

**Additional context**
Cesium have this function here:
https://github.com/CesiumGS/cesium/blob/1.126/packages/engine/Source/Scene/Scene.js#L4670

WilliamLiu-1997:
Fix #1000 
Fix #406 
Fix #378

Hi @gkjohnson , this is the in progress PR, so you can have a look at the issue described here
[https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/406#issuecomment-2680448952](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/406#issuecomment-2680448952)

This plugin can add multiple shape (OBB,Sphere,EllipsoidRegion,Frustum,Ray) by call `addLoadRegion` function (input format should be `{shape:OBB|Sphere|EllipsoidRegion|Frustum|Ray, errorTarget:number}`)

It has two mode
1. load tiles within the region according to the errorTarget of the region, other tiles are loaded as normal
2. only load tiles within the region according to the errorTarget of the region
This can be control by function `setOnlyLoadTilesInRegions`

In the mode 2, it is fine, but if using mode 1 and `tilesRenderer.targetError` is small, the plugin does not work.

**In addition, even if I set the targetError to 0, the level of detail of tileset won't increase to the highest.** Are there some settings I missed or is there some issue on the targetError?

The example/index.js is modified to temporarily test the plugin.
Set `rayCaster` to `ALL_HITS` or `FIRST_HIT_ONLY` will enable the plugin, `showOnlyTilesInRegion` controls the mode of the plugin
Thanks


gkjohnson:
I've adjusted the plugin to behave in line with the rest of the plugin architecture and adjusted the error logic to make sure it behaves as expected. Let me know if this feels right to you and I can give more in depth feedback about the code structure.

example/index.js
```diff
@@ -95,7 +92,19 @@ function reinstantiateTiles() {
...
+	tiles.registerPlugin(
+		new CesiumIonAuthPlugin( {
+			apiToken: params.ionAccessToken,
+			assetId: params.ionAssetId,
+		} )
+	);
+	tiles.maxByteSize = 536870912 * 2;
+	tiles.maxSize = 10000;
+	tiles.autoDisableRendererCulling = false;
+	tiles.loadSiblings = true;
```
gkjohnson, line 105: Some notes:
- `loadSiblings` is no longer an option on TilesRenderer
- `maxSize` is also not an option on TilesRenderer - perhaps you intended to set this on the lru cache
- `maxByteSize` is not an option on TilesRenderer, either. You're likely trying to set maxByte***s***Size on lru cache.
- It shouldn't be necessary to change `autoDisableRendererCulling`

The reason tiles aren't loading when the error is set lower is because the lru memory cache is filling up quickly (this tile set is using 1024x1024 textures) and not allowing any more tiles to load.

src/plugins/three/RegionTilesLoadingPlugin.js
```diff
@@ -0,0 +1,236 @@
...
+		return false;
+
+	}
+
+	calculateError( tile ) {
+
+		if ( tile.__inRegion ) {
+
+			const { tiles } = this;
+			return tile.geometricError - tile.__regionErrorTarget + tiles.errorTarget + 100;
```
WilliamLiu-1997, line 219: Hi @gkjohnson , I am not sure why we need `+100` here?

WilliamLiu-1997 (commented):
Hi @gkjohnson , everything feels right to me, execpt the `calculateError` function, I am confused about `+100` in the function, thanks very much.

src/plugins/three/RegionTilesLoadingPlugin.js
```diff
@@ -207,24 +207,20 @@ export class RegionTilesLoadingPlugin {
 
 		}
 
-		if ( ! tile.__inRegion && ( ! this._onlyLoadTilesInRegions || this.__regionsArraySorted.length === 0 ) ) {
-
-			return true;
-
-		}
-
```
WilliamLiu-1997: In addition, this change will make `onlyLoadTilesInRegions` mode unuseable, do we still want this mode?

src/plugins/three/RegionTilesLoadingPlugin.js
gkjohnson: Sorry I accidentally committed this while testing something. The `+ 100` can be removed.

src/plugins/three/RegionTilesLoadingPlugin.js
gkjohnson: This option shouldn't be necessary - if a user wants to toggle between displaying only tiles in a certain region then they should remove all regions and cameras other than the ones they want to be considered.

gkjohnson (commented):
Thanks! I've just add some more detailed comments. I think we should also make the `addRegion` function behave a bit more like I've laid out in https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/406#issuecomment-2677698105 so it's a more generic. Specifically the passed object should take custom functions for performing the shape intersection and calculate error test so users can provide their own custom shapes if needed. Something like so:

```js
addRegion( {
  intersectsTile: ( boundingVolume, tile ) => Boolean
  calculateError: ( boundingVolume, tile ) => Number,
} );
```

Then we can provide some pre-made classes for intersecting primitives like spheres, boxes, or rays (I think we can skip "EllipsoidRegion" and "Frustum" for now):

```js
class RayRegion {
  errorTarget = 10;
  ray = null;
  constructor( ray ) { /* ... */ }
  
  intersectsTile( boundingVolume ) {
    // perform any ray transformations needed
    return boundingVolume.intersectsRay( this.ray );
   }
  
  calculateError() {
    return this.errorTarget;
  }
}
```

---

And the last couple things are - we need to decide if the provided "regions" should be in world space or local tile set space. And then if we could add some basic docs to the `/plugins/README.md` file that would be great.

Very nice work on this! I appreciate it.

example/index.js
```diff
@@ -697,4 +697,4 @@ function render() {
 
 	}
 
-}
+}
```
gkjohnson: Looks like we lost a newline here.

gkjohnson (commented):
Just made a few comments and then I may make a few other changes later.

> Do I need to add an example to show this plugin?

A simple example showing only the plugin usage would be appreciated. Maybe something like rendering just the data in specific region (toggleable sphere, ray, etc). The Mars or earth data would be good for the example.

WilliamLiu-1997:
Hi @gkjohnson , I have made some adjustments according to your comments, is there anything else needed for the plugin, if it looks good for you I will start to write the example

gkjohnson:
Thanks - just made the following changes:
- Move all region-load-related code into the same file.
- Resume and move some intersection logic from the existing shapes and TileBoundingVolume classes.
- Change the regions to operate in the tile local coordinate frame.
- Avoid modifying the tile with new fields.
- Fix up some types.

I think the plugin should be good to go, now, if you could take another look. Though I haven't been able to test it properly, yet.

WilliamLiu-1997:
Thanks! I have created an example that includes Ray, Sphere, and Box (OBB) regions. I also adjusted the calculateError function in the plugin—it should now find the maxError instead of the minError. Additionally, I fixed some small issues, such as type definitions and intersection tests.
I think everything is ready now?

gkjohnson:
Thanks - it looks like I was referencing three.js' OBB implementation rather than the 3d tiles one when modifying the shape intersect functions. I'll make some changes to the example to use the Mars data set at a later time, as well.

Thanks!

PR #1008, opened 2025-02-26
r3f components typing

kayden1940:


kayden1940:
@gkjohnson mind taking a look? I personally find this useful.

gkjohnson:
Hi and thanks for the contribution. Unfortunately I don't really understand how all of this templating and typing is working. If this is going to be merged I'll have to ask for some commitment to maintaining and answering questions about these files because I won't be able to do so myself. 

From what I can tell, as well, these don't fully describe what types can be passed into the the elements. From the [docs](https://github.com/NASA-AMMOS/3DTilesRendererJS/tree/master/src/r3f#tilesrenderer): subproperties can be specified via dashes, similar to other react three fiber component fields (`lruCache-minSize={ 100 }` is the same as `tiles.lruCache.minSize=100`, for example).

If this is something you can support then I think we can do the following to get this merged:
- Limit the specified types to only those components that are exposed publicly.
- Add support for subproperties specified via dashes.
- Add test files to verify that some of the inferred type generation is working as expected.

kayden1940:
> Hi and thanks for the contribution. Unfortunately I don't really understand how all of this templating and typing is working. If this is going to be merged I'll have to ask for some commitment to maintaining and answering questions about these files because I won't be able to do so myself.
> 
> From what I can tell, as well, these don't fully describe what types can be passed into the the elements. From the [docs](https://github.com/NASA-AMMOS/3DTilesRendererJS/tree/master/src/r3f#tilesrenderer): subproperties can be specified via dashes, similar to other react three fiber component fields (`lruCache-minSize={ 100 }` is the same as `tiles.lruCache.minSize=100`, for example).
> 
> If this is something you can support then I think we can do the following to get this merged:
> 
> * Limit the specified types to only those components that are exposed publicly.
> * Add support for subproperties specified via dashes.
> * Add test files to verify that some of the inferred type generation is working as expected.

Happy to take this challenge, see what I can do :))

kayden1940:
Expected to finish it this week hopefully

kayden1940:
@gkjohnson 
The new commit consist of updates below:
Limiting the types to publicly components only;
Adding support for subproperties specified via dashes.
Test cases to verify the typing is working(test/types/r3f-types.test.tsx).

I have tested the typing manually in a minimal project as well.
Thanks for reviewing and this lovely library. I imigne there would be some catch up to do untill the library reaches v1.0 and I am here for it. Welcome to reach me for follow up. I am active in github or twitter(kaydent_cc).

src/r3f/components/TilesRenderer.d.ts
```diff
@@ -0,0 +1,79 @@
...
+	setCamera?: (camera: Camera) => boolean;
+	deleteCamera?: (camera: Camera) => boolean;
+	setResolutionFromRenderer?: (
+		camera: Camera,
+		renderer: WebGLRenderer,
+	) => boolean;
+	dispose?: () => void;
+	onLoadTileSet?: () => void;
+	onLoadContent?: () => void;
+	onTileVisibilityChange?: (tile: any, visible: boolean) => void;
```
gkjohnson, line 73: Not sure if there's an easy way to do this, as well, but all events from TilesRenderer can be added as props by converting dash casing to camel casing and prepending "on" - like so:

```
load-model -> onLoadModel
load-tile-set -> onLoadTileSet
```

test/types/r3f-types.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+	});
+
+	test("TilesPlugin correctly infers plugin constructor parameters", () => {
+		render(<TilesPlugin plugin={DeepZoomImagePlugin} args={{ foo: "bar" }} />);
+	});
+
+	test("DashedProperties allows kebab-case props", () => {
+		render(
+			<TilesRenderer
+				url="https://raw.githubusercontent.com/NASA-AMMOS/3DTilesSampleData/master/msl-dingo-gap/0528_0260184_to_s64o256_colorize/0528_0260184_to_s64o256_colorize/0528_0260184_to_s64o256_colorize_tileset.json"
```
gkjohnson, line 39: For readability can we move this url to a variable and reuse it across the tests

test/types/r3f-types.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+});
+TilesRenderer;
+
+describe("TilesRenderer Typing", () => {
+	test("TilesRenderer accepts correct props", () => {
+		render(
+			<TilesRenderer
+				url="https://raw.githubusercontent.com/NASA-AMMOS/3DTilesSampleData/master/msl-dingo-gap/0528_0260184_to_s64o256_colorize/0528_0260184_to_s64o256_colorize/0528_0260184_to_s64o256_colorize_tileset.json"
+				enabled
+				cameras={[new Camera()]}
```
gkjohnson, line 27: Setting "cameras" this way isn't valid. The TilesRenderer automatically grabs the r3f active render camera and uses it automatically.

kayden1940:
> Thanks! Added some comments - can we also remove the .DS_STORE and pnpm files from the commits.

Addressing them in the next commit, thanks for reviewing.

src/r3f/components/TilesRenderer.d.ts
kayden1940: I believe it's within typescript capacity but I gonna look into it to confirm

src/r3f/components/TilesRenderer.d.ts
```diff
@@ -0,0 +1,79 @@
...
+	enabled?: boolean;
+	autoDisableRendererCulling?: boolean;
+	cameras?: Camera[];
+	setCamera?: (camera: Camera) => boolean;
+	deleteCamera?: (camera: Camera) => boolean;
+	setResolutionFromRenderer?: (
+		camera: Camera,
+		renderer: WebGLRenderer,
+	) => boolean;
+	dispose?: () => void;
```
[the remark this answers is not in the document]
kayden1940: Correct, I am removing all the overlapping properties, thanks for pointing this out.

kayden1940:
New commit attempt to address the feedback, all events from TilesRenderer is now added dynamically mapped as props for the r3f component.
image

src/r3f/components/TilesRenderer.d.ts
```diff
@@ -0,0 +1,72 @@
...
+type TilesRendererEventMapForR3f = {
+	[K in keyof TilesRendererEventMap as `on${Capitalize<CamelCase<K>>}`]?: EventHandler<K>;
+};
+
+interface TilesRendererProps
+	extends Partial<TilesRendererImpl>,
+		TilesRendererEventMapForR3f {
+	url?: string;
+	enabled?: boolean;
+	cameras?: Camera[];
```
gkjohnson, line 65: "cameras" is not a valid property for the TilesRenderer component.

package.json
```diff
@@ -78,4 +81,4 @@
     "react": "^18.3.1",
     "react-dom": "^18.3.1"
   }
-}
+}
```
gkjohnson, line 84: super small nit: lets avoid changes to the new line at the end of the file.

package.json
```diff
@@ -50,7 +50,10 @@
     "@takram/three-atmosphere": "^0.6.0",
     "@takram/three-geospatial": "^0.0.1-alpha.4",
     "@takram/three-geospatial-effects": "^0.0.1-alpha.4",
+    "@testing-library/dom": "^10.4.0",
```
gkjohnson, line 53: Why do we need `@testing-library/dom`? It doesn't look like it's used anywhere in the project.

gkjohnson (commented):
Last few change requests. Also it looks like the "pnpm-lock.yaml" file is still present in the PR. Lets remove that, as well, and we can get this merged.

Thanks!

package.json
kayden1940: it's require to use @testing-library/react
https://testing-library.com/docs/react-testing-library/intro/

package.json
kayden1940: Quite odd, in the repo the new line at the end is not showing
https://github.com/kayden1940/3DTilesRendererJS/blob/master/package.json

kayden1940: Sorry but I can't find a way to git add this change, git seems to assume it's unchanged despite all the git hacks I find on the internet, the file in my repo is not showing the ending line either.

kayden1940:
New commit:
Remove pnpm-lock.yaml,
remove invalid property for the TilesRenderer component.

gkjohnson:
I've just fixed the new line and restored the package-lock.json file. The last thing to do is to fix some of the lint errors. Running eslint with --fix may fix a number of these. Otherwise you can test the code style by using `npm run lint` in the repo or by installing an eslint plugin for your code editor.

kayden1940:
> I've just fixed the new line and restored the package-lock.json file. The last thing to do is to fix some of the lint errors. Running eslint with --fix may fix a number of these. Otherwise you can test the code style by using `npm run lint` in the repo or by installing an eslint plugin for your code editor.

Seems .eslintrc.json in the project is a deprecated format. See if I can find a way around this for now.
https://eslint.org/docs/latest/use/configure/migration-guide

gkjohnson:
> Seems .eslintrc.json in the project is a deprecated format. See if I can find a way around this for now.

The implicit version used in the package.json is 8.57.1. I've made the change in the PR - I've also made the following changes:
- Fix line ending issues in the files. It looks like your editor is probably configured to use different line / file endings. I recommend looking into an "editorconfig" plugin to make sure the editor using the line endings specified in the `.editorconfig` file.
- Add "tsx" to the list of files in the test folder to run through eslint.
- Fix the lint errors in the types test files.

It looks like there are still a number of errors in the types tsx test file when looking at the issues in the vscode console:

image

And Jest is reporting that no assertions are being run in the tests. Is this set up properly to evaluate the types?

image

I don't believe that file is being run at all through tsconfig to test the types when running `npm run test`. I don't know enough about setting up a typescript project to make sure it's being tested, though. How is the tsconfig supposed to be configured here?

kayden1940:
> > Seems .eslintrc.json in the project is a deprecated format. See if I can find a way around this for now.
> 
> The implicit version used in the package.json is 8.57.1. I've made the change in the PR - I've also made the following changes:
> - Fix line ending issues in the files. It looks like your editor is probably configured to use different line / file endings. I recommend looking into an "editorconfig" plugin to make sure the editor using the line endings specified in the `.editorconfig` file.
> - Add "tsx" to the list of files in the test folder to run through eslint.
> - Fix the lint errors in the types test files.
> 
> It looks like there are still a number of errors in the types tsx test file when looking at the issues in the vscode console:
> 
> image
> 
> And Jest is reporting that no assertions are being run in the tests. Is this set up properly to evaluate the types?
> 
> image
> 
> I don't believe that file is being run at all through tsconfig to test the types when running `npm run test`. I don't know enough about setting up a typescript project to make sure it's being tested, though. How is the tsconfig supposed to be configured here?

The test I wrote is a standalone testing library, and I can and should plug the testing to jest in this case, will get back to you asap(this week), thanks for walking through with me.

kayden1940:
> > Seems .eslintrc.json in the project is a deprecated format. See if I can find a way around this for now.
> 
> The implicit version used in the package.json is 8.57.1. I've made the change in the PR - I've also made the following changes:
> 
> * Fix line ending issues in the files. It looks like your editor is probably configured to use different line / file endings. I recommend looking into an "editorconfig" plugin to make sure the editor using the line endings specified in the `.editorconfig` file.
> * Add "tsx" to the list of files in the test folder to run through eslint.
> * Fix the lint errors in the types test files.
> 
> It looks like there are still a number of errors in the types tsx test file when looking at the issues in the vscode console:
> 
> image
> And Jest is reporting that no assertions are being run in the tests. Is this set up properly to evaluate the types?
> 
> image
> I don't believe that file is being run at all through tsconfig to test the types when running `npm run test`. I don't know enough about setting up a typescript project to make sure it's being tested, though. How is the tsconfig supposed to be configured here?

I need to revisit the test and my ide setup, thanks for walking this through with me.

kayden1940:
> * Add "tsx" to the list of files in the test folder to run through eslint.

Seems jest already recognizes and testing the tsx in our case.

kayden1940:
> I don't believe that file is being run at all through tsconfig to test the types when running `npm run test`.

I am positive jest already recognizes and testing the tsx in our case. The warning log are correctly marking the lines in the test.

test/types/r3f-types.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+	} );
+
+	test( 'GlobeControls rejects invalid props', () => {
+
+		render( <GlobeControls foo={'bar'} /> );
+
+	} );
+
+} );
+TilesRenderer;
```
gkjohnson, line 27: Is this line necessary?

gkjohnson:
> I am positive jest already recognizes and testing the tsx in our case. 

The tests aren't passing, though. What I'm trying to understand is _how_ the test hooks are working? How is jest being used here? The `describe and `test` functions don't seem to be doing anything because there are no `assert` / `expect` calls. What we want is for the tests to "pass" if the types are correctly failing when an unknown field / type is used. I'm not seeing how that's being done if it's possible at all.

test/types/r3f-types.test.tsx
kayden1940: omg i goofed sorry about that.

kayden1940:
> > I am positive jest already recognizes and testing the tsx in our case.
> 
> The tests aren't passing, though. What I'm trying to understand is _how_ the test hooks are working? How is jest being used here? The `describe and `test`functions don't seem to be doing anything because there are no`assert`/`expect` calls. What we want is for the tests to "pass" if the types are correctly failing when an unknown field / type is used. I'm not seeing how that's being done if it's possible at all.

The expect assertion has not been implemented yet which I will be focusing on next commit, current test.tsx only looks for if properties of a component are typed correctly (manual check).

kayden1940:
@gkjohnson want to know more about the intention of this line in jest.config.json
`
"testPathIgnorePatterns": ["types"]
`
This blocked `npm run test` to scan the tsx.

kayden1940:
The only option for from the extent of my knowledge is jest-runner-tsd, on my researching.

kayden1940:
From the extent of my knowledge, I don't see a good enough option for that for now.

sguimmara:
Indeed, maybe we should put those test in a different folder so that it is not caught by jest, and then run a simple `tsc --noEmit` on the folder just for the typechecker to run.

gkjohnson:
It's unfortunate that there's no method for testing and validating type compatibility in typescript. In that case I agree - lets remove the jest functions from the type checking files and we can just perform visual confirmation for now. Unfortunately if we use `tsc` then the tests will fail because we've included types that we know are incorrect to ensure the generic typing is rejecting them correctly.

kayden1940:
> Indeed, maybe we should put those test in a different folder so that it is not caught by jest, and then run a simple `tsc --noEmit` on the folder just for the typechecker to run.

> Agreed on this approach using tsd, should we roll?

gkjohnson:
> Agreed on this approach using tsd, should we roll?

What's `tsd`? Assuming you mean `tsc` how will we individually test the cases were we expect the types to fail (ie the cases where we're deliberately assigning an invalid type to make sure it's caught)?

gkjohnson:
Oh is this `tsd`: https://www.npmjs.com/package/tsd? If this lets us test types then that would be great.

kayden1940:
> Oh is this `tsd`: https://www.npmjs.com/package/tsd? If this lets us test types then that would be great.

Correct, that's for my next commit then :)

kayden1940:
> Oh is this `tsd`: https://www.npmjs.com/package/tsd? If this lets us test types then that would be great.

End up using tsc directives instead so less dependence involve.

src/r3f/components/CameraTransition.jsx
```diff
@@ -1,6 +1,6 @@
 import { forwardRef, useEffect, useMemo } from 'react';
 import { useFrame, useThree } from '@react-three/fiber';
-import { CameraTransitionManager } from '3d-tiles-renderer';
+import { CameraTransitionManager } from '../../three/controls/CameraTransitionManager.js';
```
gkjohnson, line 3: Please be careful not to commit unnecessary changes like this.

src/r3f/components/TilesRenderer.d.ts
```diff
@@ -0,0 +1,69 @@
+import type {
+	TilesRenderer as TilesRendererImpl,
+} from '../../index';
+import type { TilesRendererEventMap } from '../../three/TilesRenderer';
```
gkjohnson, line 4: These should import from the same location, right?

tsconfig.test.json
```diff
@@ -2,6 +2,9 @@
 	"extends": "./tsconfig.json",
 	"compilerOptions": {
 		"noEmit": true,
+		"skipLibCheck": true,
```
gkjohnson, line 5: Why are we enabling "skipLibCheck"? The docs seem to indicate that this makes the type checking less strict.

gkjohnson (commented):
Great! Just a few more comments - I fixed some lint errors and restored the TilesRenderer types test file.

src/r3f/components/CameraTransition.jsx
kayden1940: It's necessary from what I can tell, ts yell 3d-tiles-renderer have not been typed at run time, unwanted error.

src/r3f/components/TilesRenderer.d.ts
kayden1940: Would be better yes.

src/r3f/components/CameraTransition.jsx
gkjohnson: This is not a typescript file, though.

tsconfig.test.json
kayden1940: skipLibCheck skipping only the packages the test is using, we are not testing those packages in this case, so this is for skipping unwanted error that have nothing to do with 3DTilesRendererJS. You can run test without it to see the result.

src/r3f/components/CameraTransition.jsx
gkjohnson: Oh I actually see now that this is more consistent with other files that don't import from the root index file. It shouldn't be related to tyepscript, though. We can keep this change.

kayden1940: which one?

kayden1940: Oh that's weird, you are right, my ide throw error on that earlier. Fixing this next commit.

tsconfig.test.json
gkjohnson: I've just fixed two issues in the d.ts files cause problems when this is disable but this one remains. Is the implication here that react-three/fiber has incorrect types? Or uses an incompatible version of something?

image

kayden1940: For example if 3DTilesRendererJS is using lodash, any typing error on the lodash side will show up.

kayden1940: This is the only error it pops up to me before your commits, can I take a look to your fixed error?

kayden1940: > I would prefer to not disable stricter type checking when the whole point is to check types.

But is this included to fix all the external packages tho

src/r3f/components/CameraTransition.jsx
kayden1940: **_
image
_**

tsconfig.test.json
kayden1940: > Is the implication here that react-three/fiber has incorrect types? Or uses an incompatible version of something?

Either way it's boiling down to react-three/fiber has incorrect types in this codebase and it's not in 3DTilesRendererJS.

kayden1940: > This is the only error it pops up to me before your commits, can I take a look to your fixed error?

@gkjohnson This one is important removing this shouldn't be resulting into this or else something is really off.

gkjohnson:
> Why removing DS_STORE from ignore?

DS_STORE is a Mac OS file that shouldn't be getting committed or added via git, anyway, so it shouldn't be necessary. The only place I've seen this file show up is copying a folder from Mac to a Windows computer to look at it.

tsconfig.test.json
gkjohnson: > This is the only error it pops up to me before your commits, can I take a look to your fixed error?
> ...
> This one is important removing this shouldn't result into internal error popping up or else something is really off.

These are the errors that appear in your last commit 70e062e8deca6bc9d91a2462b679b27dbbe686d3 when "skipLibCheck" is set to "false".

image

gkjohnson: > Either way it's boiling down to react-three/fiber has incorrect types in this codebase

_Does_ it have incorrect types? Or are we using incompatible versions of type versions or something like that? I find it hard to believe that r3f is publishing bad types.

kayden1940:
> _edit_ What platform are you working on? How did DS_STORE files show up in your commits?
macOS, I don't exactly know the reason but I can apply local .gitignore on my machine.

tsconfig.test.json
kayden1940: > Or are we using incompatible versions of type versions or something like that?

I am finding this out.

kayden1940: > I find it hard to believe that r3f is publishing bad types.

[Consider how big r3f it's not impossible.](https://github.com/pmndrs/react-three-fiber/issues/925). I am still in search.

kayden1940: You are right skipLibCheck skip more than external libraries and I shouldn't use it in this case.

kayden1940: > > Or are we using incompatible versions of type versions or something like that?

Found out the root cause:
    "@types/react": "^18.3.12",
    "@types/react-dom": "^18.3.1",
Downgrading them from 19 to 18 resolve the error

kayden1940:
managed to fix the r3f type issue by installing type/react, this one I can't figure a way to.
image

gkjohnson:
Perhaps @RodrigoHamuy has some ideas on how to deal with some of these jest / typescript issues in https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/1008#issuecomment-2746507954? We're seeing some issues when trying to get a test running to validate types.

Otherwise I'll have to take a look later this week.

RodrigoHamuy:
@gkjohnson sorry, did try a few things, but so far no luck. I personally use `vitest` and `three-stdlib` instead of `jest` and `three/examples` because in the past I run a lot into issues like this with bundlers. Sorry!

kayden1940:
@gkjohnson anything I can help following up? I am down for overhaul if that's needed.

gkjohnson:
Unfortunately I won't have time to dig into this for awhile. Additionally dealing with these types of typescript and build issues is exactly what I'm trying to avoid having to spend time one.

> @gkjohnson anything I can help following up? I am down for overhaul if that's needed.

For now lets disable the explicit typing tests via jest (or whatever is breaking) so we can do visual checking so we can get this merged. Then if you could make a new issue and outline what you might recommend in an overhaul for this that would be great and we can fix the overall testing for types in another PR.

kayden1940:
Latest commit rollback testing types with jest, op to switch to vitest in the next pr, honestly jest is not helping in this case.

gkjohnson:
Thanks - we can give this a go! Per https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/1008#issuecomment-2765816903, do you mind making an issue to discuss how we can overhaul testing for types (or in general if jest is considered more defunct)

kayden1940:
Here are some tasks I see worth considering, dropping here for brainstorming.
Updating ts to latest version gracefully: regardless of how the next step would be this seems essential, at that point it's not only about r3f components anymore, so best be careful;
My problem with jest is it involve many extensions to simple things and the ambiguous warnings and configs have been dragging from doing the tasks. I have good experience with vitest and three-stdlib, so I am looking forward to seeing if they are a good fit for the whole puzzle (again, big surgery).
Although that's a lot of work, but if done right, I can see it would pay off long terms for not having to thinking about & deal with them.
Would love to have hands to borrow, and splitting them into each pr might be a good idea. fyi I am currently making a game with those r3f components (first reason is to have better understanding of the design).

kayden1940:
also I need to look into why my prettier/eslint not respecting the codebase setting.

gkjohnson:
Thanks @kayden1940 - to reiterate: could you please make a [new, separate _Github Issue_](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues) in this project with your thoughts laid out as separate bullet points / sections so it's clear and can be tracked. This is a now closed PR so any discussion here will be easily lost. Creating an issue and clearly separating the tasks will also help increase visibility to others who may be able to help, as well, and simplify further discussion.

kayden1940:
Happy to, was dropping the chain of thought somewhere so I don't forget

PR #1009, opened 2025-02-28
Add TypeScript definition for setLatLonToYUp

felipunky:
Note that at: https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/4c51400763c3689f26cb4efaf4e301e964265471/src/three/TilesRenderer.js#L957 it is mentioned that this function will be deprecated in the future.


gkjohnson:
Thanks @felipunky - to be honest I had forgotten that this function was not deprecated, yet. I would prefer to deprecate the function and not add typescript definitions for it.

If you'd like to reorient the globe so that it displays a certain point at the origin you can use the [ReorientationPlugin](https://github.com/NASA-AMMOS/3DTilesRendererJS/tree/master/src/plugins#reorientationplugin).

felipunky:
Thanks Garrett, got it to work with the ReorientationPlugin.

PR #1014, opened 2025-03-03
Fix LoadRegionPlugin typings

xawill:
Thank you for this new plugin! However there were a few typos in the typings probably due to copy/paste which failed to compile in Typescript projects. Hopefully I caught them all, but feel free to double check.

### DIFF
--- a/src/plugins/three/LoadRegionPlugin.d.ts
+++ b/src/plugins/three/LoadRegionPlugin.d.ts
@@ -1,10 +1,10 @@
-import { TileBoundingVolume } from '../../../three/math/TileBoundingVolume';
-import { Tile } from '../../../base/Tile';
-import { TilesRenderer } from '../../../three/TilesRenderer';
+import { TileBoundingVolume } from '../../three/math/TileBoundingVolume';
+import { Tile } from '../../base/Tile';
+import { TilesRenderer } from '../../three/TilesRenderer';
 import { OBB } from '../../three/math/OBB';
 import { Sphere, Ray } from 'three';
 
-interface Region {
+class Region {
 
 	errorTarget: number;
 	intersectsTile( boundingVolume: TileBoundingVolume, tile: Tile, tilesRenderer: TilesRenderer ): boolean
@@ -12,7 +12,7 @@ interface Region {
 
 }
 
-export class ReorientationPlugin {
+export class LoadRegionPlugin {
 
 	addRegion( region: Region ): void;
 	removeRegion( region: Region ): void;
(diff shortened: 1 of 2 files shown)


gkjohnson:
Thanks! Any ideas how these things can be caught in a CI check? Perhaps we need to add the `src/plugins/index.d.ts` files to the tsconfig.js file so it's checked with the `tsc` call?

xawill:
Yes, exactly. I checked and you already have everything setup correctly except the missing "src/plugins/index.d.ts" file in tsconfig. However I just tried and there are currently 12 errors in 2 files (regarding `userData` and Array default type in GLTFExtensionMetadata) which are not trivial to fix. My recommendation would be to just remove these faulty typings and keep typings simple in general. Let me know if you need further support with this.

src/plugins/three/LoadRegionPlugin.d.ts
```diff
@@ -1,18 +1,18 @@
...
-import { Tile } from '../../../base/Tile';
-import { TilesRenderer } from '../../../three/TilesRenderer';
+import { TileBoundingVolume } from '../../three/math/TileBoundingVolume';
+import { Tile } from '../../base/Tile';
+import { TilesRenderer } from '../../three/TilesRenderer';
 import { OBB } from '../../three/math/OBB';
 import { Sphere, Ray } from 'three';
 
-interface Region {
+declare class Region {
```
gkjohnson, line 7: Can you explain what "declare" is doing here?

gkjohnson:
> My recommendation would be to just remove these faulty typings and keep typings simple in general. Let me know if you need further support with this.

Thanks for looking into this. I'm not a Typescript expert so any advice is appreciated. I'll take you recommendations on the fixes for these.

And yes looking at the fixes it looks like the there were quite a few hasty errors 😅 I figured these were being tested so it's good to have it added to the CI.

src/plugins/three/LoadRegionPlugin.d.ts
xawill: It only declares the class for the TypeScript compiler, not for the JavaScript runtime. It is required by TS to either have `declare` or `export` in top-level declarations. Based on your original code, I imagined you didn't want to export the class.

xawill:
I am not a TS expert neither; let's say power user.
Here is my proposal: defining types so that users can use them, but leave it up to them to redefine anything according to their specific use cases.
Note however that I didn't check the JS code the make sure the typings make sense. I just inferred based on current typings understanding, so please double check.

gkjohnson:
Thanks - I think we can merge this as-is and if someone who knows more about the typing system wants to make some more robust type contributions later on we can update them.

Issue #1023, opened 2025-03-10
How to enable tilt and rotation for googleMapsExample?


jeantimex:
Is there a way to enable tilt and rotation like Google Earth?

gkjohnson:
As the issue template for questions states, please include sufficient detail to understand your question. I'm not sure what "tilt and rotation like Google Earth" means.

gkjohnson:
Answered in more detail [here](https://x.com/jeantimex/status/1898986959391146246).

jeantimex:
Thanks for your reply! Let me explain my questions in details.

- 3d-tiles-renderer version: 0.4.7
- three.js version: 0.174.0
- Link to the project: https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/googleMapsExample.html

I would like to control the tilt and rotation using shift key + dragging, that is shift key + dragging up and down the map will change the tilt, shift key + dragging the cursor left and right will change the rotation of the map.

Right now I can use two fingers and right cursor to achieve, but I can’t use shift key + cursor dragging.

According to the current code of the Environment Control, it looks like shift key + cursor dragging should be supported, but it doesn’t work for me.
![Image](https://github.com/user-attachments/assets/1a446b52-5e2d-4b63-bd5b-86c91dc42bf3)

I am using MacBook Pro with macOS 15 Sequoia, Chrome browser 134.0.6998.35/36

gkjohnson:
Thanks - look like this broke at some point. I think the change should be to listen to key events on the page root node rather than the passed canvas element itself for key events specifically, similar to three.js' OrbitControls. Would you like to make a PR?

jeantimex:
Yep I can try that! Thanks!

jeantimex:
I just submitted https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/1029

PR #1029, opened 2025-03-12
Fixed a bug that caused shift+click not to work

jeantimex:
Fixed https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1023

The reason why the current solution doesn't work is because the `pointerdown` event handler calls `preventDefault()` which makes the canvas element cannot be focused, however the keydown event relies on the canvas element to be focused.

Instead of listening to the `keydown` and `keyup` events on the canvas element in order to handle the shift + click gesture, we can simply detect whether the shift key is pressed using [MouseEvent.shiftKey](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/shiftKey) property.

This fix should allow end users to use shift key to control the tilt and rotation of the globe.

### DIFF
--- a/src/three/controls/EnvironmentControls.js
+++ b/src/three/controls/EnvironmentControls.js
@@ -214,31 +214,9 @@ export class EnvironmentControls extends EventDispatcher {
 		this.pointerTracker.domElement = domElement;
 		domElement.style.touchAction = 'none';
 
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thank you!

Issue #1024, opened 2025-03-10
R3F Atmosphere example is broken


gabschettino:
Takram's Atmosphere example is broken [here](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/r3f/atmosphere.html)

It reads:
`GET https://takram-design-engineering.github.io/three-geospatial/atmosphere/scattering_float.bin 404 (Not Found)`

When trying to get  the binary files for the textures.

gkjohnson:
Thanks for the report! Looks like I forgot to fix the external link in the demo. Would you like to make a PR that fixes the issue? The quick fix would be to download and push the bin files from the associated atmosphere version to a data folder in `r3f` or use fixed 3d git commits at references to the github content directly.

It would also be great to update the atmosphere package to the latest version and add some of the newer effects like clouds if you're interested.

gabschettino:
Would be glad to!

> It would also be great to update the atmosphere package to the latest version and add some of the newer effects like clouds if you're interested.

Interested as well, but that would fit into a different branch, as I think it would need to update some peer deps as well, right?

gkjohnson:
> Interested as well, but that would fit into a different branch, as I think it would need to update some peer deps as well, right?

Lets see if we can update the demo by just updating dev dependencies. Since no core code should have to change to make update the demo we shouldn't have to change any dependencies (peer, optional, etc)

gabschettino:
Regarding the Atmosphere example fix:
I tried to move all the binaries to an assets folder, however I've got an error message similar to [this one](https://github.com/takram-design-engineering/three-geospatial/issues/31) reported in Takram's Atmosphere repo last month. Which could be resolved running git LFS.

However, the latest version of Atmosphere as of now (0.8.0) [simplified ](https://github.com/takram-design-engineering/three-geospatial/commit/9903d43fafb60fbc9c8e12e83319d4c2bc0285c7) this setup to get the binaries directly from GitHub if left undefined, so I prefered to use the same approach.

gkjohnson:
Thanks - I didn't realize the files were that big. Referencing the github files directly sounds best.

PR #1030, opened 2025-03-13
updated atmosphere binaries reference

gabschettino:
Related to #1024


gkjohnson:
Thanks! Can we change the the remaining references to `https://takram-design-engineering.github.io/` and add a comment stating what the referenced commit is? Ideally we'd use a version tag rather than a commit hash for readability.

gkjohnson:
> How about leaving it undefined so it always refers to the latest compatible ref that I maintain?

Sounds even better! We should update to the latest version to take advantage of this, right?

shotamatsuda:
I just noticed this version is before the change allowing textures to be undefined. Yes, we need to update it to v0.8.0 or later for it, but that might introduce other issues. I’ll check if the example works with the latest compatible version for R3F v8, which is atmosphere v0.10.2.

shotamatsuda:
Yes, the example works fine with v0.10.2. Please update the packages to the below:

```
"@takram/three-atmosphere": "^0.10.1",
"@takram/three-geospatial": "^0.1.0",
"@takram/three-geospatial-effects": "^0.1.0",
```

then, remove the `textures` prop of `Atmosphere` and `data` prop of `Stars`.

package.json
```diff
@@ -47,9 +47,9 @@
     "@babel/preset-env": "^7.21.5",
     "@react-three/drei": "^9.114.3",
     "@react-three/postprocessing": "^2.16.3",
-    "@takram/three-atmosphere": "^0.6.0",
-    "@takram/three-geospatial": "^0.0.1-alpha.4",
-    "@takram/three-geospatial-effects": "^0.0.1-alpha.4",
+		"@takram/three-atmosphere": "^0.10.3",
+		"@takram/three-geospatial": "^0.1.0",
+		"@takram/three-geospatial-effects": "^0.1.0",
```
gkjohnson, line 52: Last thing - lets make sure our indentation is consistent in the package.json file.

gabschettino: Oops, sorry about that, didn't catch it in my editor

gkjohnson:
Thank you!

PR #1040, opened 2025-03-20
Create an empty group if a loaded glb contains no scene

### LINKED ISSUE #1039: Console Error When Tileset has GLB file that has no scene
**Describe the bug**

For some of the tilesets, if it has `glb`/`gltf` files that do not have `scene`/`vertex`/`triangle`, the tilesRenderer will continuously post error. However, Cesium won't

**To Reproduce**

Sorry I cannot provide the problematic tileset

***Code***

In line 669 it tries to update the matrix of the scene but the scene is undefined

***Live example***

Issue GLB file

Normal GLB File

You can see the `glb` file that causes the error has 0 vertex and triangle

**Expected behavior**

TilesRenderer should simply ignore the `glb` files that do not have `scene`, instead of continuously post error for this

WilliamLiu-1997:
Fixes #1039

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -663,7 +663,6 @@ export class TilesRenderer extends TilesRendererBase {
 					// rotation fix which is why "multiply" happens here.
 					result.scene = result.scene || new Group();
 					const { scene } = result;
-
 					scene.updateMatrix();
 					scene.matrix
 						.multiply( upRotationMatrix )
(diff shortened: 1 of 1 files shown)


src/three/TilesRenderer.js
```diff
@@ -661,10 +662,15 @@ export class TilesRenderer extends TilesRendererBase {
 					// (such as applying RTC_CENTER) meaning they should happen _after_ the z-up
 					// rotation fix which is why "multiply" happens here.
 					const { scene } = result;
-					scene.updateMatrix();
-					scene.matrix
-						.multiply( upRotationMatrix )
-						.decompose( scene.position, scene.quaternion, scene.scale );
+
+					if ( scene ) {
```
gkjohnson, line 666: Thanks - I think it will be easier to fix this if we just assign "scene" here if it's not present:

```js
result.scene = result.scene || new Group();
```

No other changes should be needed I don't think.

gkjohnson:
Thanks!

PR #1042, opened 2025-03-20
Fix for EXT_mesh_features/EXT_structure_metadata

wsw0108:


gkjohnson:
Please provide an explanation of what your PR is fixing and a demonstration of the bug.

wsw0108:
@gkjohnson these null checks are for gltf with below structure:
```json5
{
  "nodes": [
    {
      "name": "Root",
      "children": [1],
    },
    {
      "name": "Mesh",
      "mesh": 0
    }
}
```

gkjohnson:
Thank you for the explanation - the current suggested fix will not work even in your example case, though, because `! 0` will evaluate to `true` and the function will return early even though `0` is a valid mesh and primitive index.

Instead lets use the optional chaining syntax to access the mesh objects (note the added question mark):

```js
const primitive = parser.json.meshes[ meshes ]?.primitives[ primitives ];
```

wsw0108:
@gkjohnson changes had make as your suggestion, and I had test locally.

gkjohnson:
Thank you!

PR #1044, opened 2025-03-26
Add progressive subtree processing

### LINKED ISSUE #1034: Streamlining tile subtree loading performance
With tile sets like the Google Photorealistic Tiles tile set moving the camera can cause a ton of new internal tile sets to load resulting in severe framerate hiccups when a ton of tiles are preprocessed and initialized. While tiles are already preprocessed as-needed when they are first traversed shifting a zoomed-in camera to a new location can result in a new tile subtree loading and the whole (or nearly whole) subtree being initialized on the next traversal after load.

Instead of initializing the tiles immediate during traversal we can queue tiles to process during traversal and only process a certain amount per frame - similar to the [ImageFormatPlugin](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/5ed555896bebea09b9e4bee6537cff782b36652d/src/plugins/three/images/ImageFormatPlugin.js#L136-L143) approach to generating children progressively with a `PriorityQeueu`.

- When encountering an unprocessed tile it can be queued it instead of immediately processing them.
- The queue can be sorted using the same priority as the download and parse queues.
- Traversal should end if an unprocessed tile is encountered.
- Because the traversal relies on children being traversable and loadable we'll also need to end traversal if all children are not processed. Perhaps the parent tile can be considered incomplete if immediate children are incomplete.
- Tiles must be triggered to be re-traveresed and tiles renderer updated when a new set of tiles is processed.

gkjohnson:
Fix #1034 

**Later**
- Add option to run priority queue for a certain limit of time
- Find new way to ensure we can't double-process a child tile between callback deletion, evaluation (perhaps "has" is not the best way to handle this but works because processing is immediate)
- Handle implicit tiling processing in a way that doesn't require special changes to the plugin
- How do we remove subchildren from processing if the parent subtree has been removed? Do we need to? It means keeping some memory around unnecessarily, as well. Though tiles will process relatively quickly so this may not be an issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ tilesRenderer2.setResolutionFromRenderer( camera, renderer );
 tilesRenderer2.lruCache = tilesRenderer.lruCache;
 tilesRenderer2.downloadQueue = tilesRenderer.downloadQueue;
 tilesRenderer2.parseQueue = tilesRenderer.parseQueue;
+tilesRenderer2.processNodeQueue = tilesRenderer.processNodeQueue;
 
 // add them to the scene
 scene.add( tilesRenderer.group );
@@ -397,7 +398,7 @@ downloadQueue = new PriorityQueue : PriorityQueue
 
 _NOTE: This cannot be set once [update](#update) is called for the first time._
 
-Max jobs defaults to `10`.
+Queue for downloading tile content. Max jobs defaults to `10`.
 
 ### .parseQueue
 
@@ -407,7 +408,17 @@ parseQueue = new PriorityQueue : PriorityQueue
 
 _NOTE: This cannot be modified once [update](#update) is called for the first time._
 
-Max jobs defaults to `1`.
+Queue for parsing downloaded tile content. Max jobs defaults to `1`.
+
+### .processNodeQueue
+
+```js
+processNodeQueue = new PriorityQueue : PriorityQueue
+```
+
+_NOTE: This cannot be set once [update](#update) is called for the first time._
+
+Queue for expanding and initializing tiles for traversal. Max jobs defaults to `25`.
 
 ### .group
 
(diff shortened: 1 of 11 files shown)


pmi-gbs:
Thanks for addressing this.

I just tried this out, but I'm getting an error when using raycasts:

```
raycastTraverse.js:130 Uncaught TypeError: Cannot destructure property 'boundingVolume' of 'tile.cached' as it is undefined.
    at raycastTraverse (raycastTraverse.js:130:10)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at raycastTraverse (raycastTraverse.js:156:1)
    at TilesRenderer.raycast (TilesRenderer.js:248:19)
    at TilesGroup.raycast (TilesGroup.js:24:1)
    at intersect (three.core.js:45336:1)
    at intersect (three.core.js:45348:1)
    at intersect (three.core.js:45348:1)
    at intersect (three.core.js:45348:1)
    at Raycaster.intersectObject (three.core.js:45300:1)
```

Maybe the relevant tile is not yet processed and doesn't have a `boundingVolume` yet?

gkjohnson:
Thanks - looks like the "first hit" code path is resilient to the change but both raycast traverse functions should probably be updated. There shouldn't be any reason to traverse children if they haven't already been initialized by the renderer.

PR #1047, opened 2025-03-31
gltf-plugin: null check for subclasses of PropertySetAccessor

wsw0108:
Not all of the properties of a class in schema will appear in it‘s instances.

### DIFF
--- a/src/plugins/three/gltf/metadata/classes/PropertyAttributeAccessor.js
+++ b/src/plugins/three/gltf/metadata/classes/PropertyAttributeAccessor.js
@@ -8,7 +8,7 @@ class PropertyAttributeClassProperty extends ClassProperty {
 
 		super( enums, classProperty, attributeProperty );
 
-		this.attribute = attributeProperty.attribute;
+		this.attribute = attributeProperty?.attribute;
 
 	}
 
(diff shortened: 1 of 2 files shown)


gkjohnson:
Can you provide example files including structural metadata that breaks? It looks like this will break other cases beyond just initialization.

wsw0108:
[web-ifc-example-1.0.zip](https://github.com/user-attachments/files/19534415/web-ifc-example-1.0.zip)

gkjohnson:
> field predefinedType missing.
> web-ifc-example-1.0.zip

Thank you - it looks like this zip only includes a file with a TableAccessors. Is there an example with PropertyAttributes?

But either way it looks like this case is checked for when accessing the data [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/e6afdb708b00acd760d2b5afdce2ce6234e1733a/src/plugins/three/gltf/metadata/classes/PropertyAttributeAccessor.js#L59-L63) and [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/e6afdb708b00acd760d2b5afdce2ce6234e1733a/src/plugins/three/gltf/metadata/classes/PropertyTableAccessor.js#L167-L171) - the constructor just wasn't checking for the property to be present. If the property is not defined then lets make sure "values" is set to "null" rather than "undefined":

```js
this.values = tableProperty?.values ?? null;
```

wsw0108:
Updated. @gkjohnson 
Thanks for your review.

gkjohnson:
Thanks @wsw0108 - we can merge this but do you mind answer the question above:

> it looks like this zip only includes a file with a TableAccessors. Is there an example with PropertyAttributes?

PR #1062, opened 2025-04-08
Allow plugins to set inView to false in calculateTileViewError

jsulli:
Based on [this](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/TilesRenderer.js#L973C5-L978C6):
```
if ( viewErrorTarget.inView ) {

	inView = true;
	inViewError = Math.max( inViewError, viewErrorTarget.error );

}
```

It's not possible for any plugin to set `inView` to false if it was previously set to `true` by the camera frustum check in the base `calculateTileViewError` function in `TilesRenderer`. 

It's not clear to me how the example for the `LoadRegionPlugin` actually works because of this flaw, is it running on older code? https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/example/loadRegion.js

When running locally, `LoadRegionPlugin` does not work, nor does my own plugin that serves a similar function, _until_ I make the change from this PR, then they both work.

### DIFF
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -970,9 +970,9 @@ export class TilesRenderer extends TilesRendererBase {
 			if ( plugin !== this && plugin.calculateTileViewError ) {
 
 				plugin.calculateTileViewError( tile, viewErrorTarget );
-				if ( viewErrorTarget.inView ) {
+				if ( viewErrorTarget.inView || viewErrorTarget.inView === false) {
 
-					inView = true;
+					inView = viewErrorTarget.inView;
 					inViewError = Math.max( inViewError, viewErrorTarget.error );
 
 				}
--- a/src/three/TilesRenderer.js
+++ b/src/three/TilesRenderer.js
@@ -970,7 +970,8 @@ export class TilesRenderer extends TilesRendererBase {
 			if ( plugin !== this && plugin.calculateTileViewError ) {
 
 				plugin.calculateTileViewError( tile, viewErrorTarget );
-				if ( viewErrorTarget.inView || viewErrorTarget.inView === false) {
+				// allow false to pass
+				if ( viewErrorTarget.inView || viewErrorTarget.inView === false ) {
 
 					inView = viewErrorTarget.inView;
 					inViewError = Math.max( inViewError, viewErrorTarget.error );


gkjohnson:
The code is working as intended but perhaps you're expecting it to do something other than designed. Can you explain your use case?

> It's not possible for any plugin to set inView to false if it was previously set to true by the camera frustum check in the base calculateTileViewError function in TilesRenderer.

This is by design. Just like camera frustums the tiles triggered to be loaded by regions are additive. Are you expecting to be able to specify a region where _not_ to load tiles so they're cut out?

> is it running on older code?

All of the examples are automatically built and run the latest code in the repository.

jsulli:
> The code is working as intended but perhaps you're expecting it to do something other than designed. Can you explain your use case?

My use case is basically that of the load region plugin, but with more granular control (complex concave shapes). Given your response I'm not sure why the `LoadRegionPlugin` doesn't work in my application, but I'll investigate more deeply to figure out what's going wrong.

> Just like camera frustums the tiles triggered to be loaded by regions are additive. Are you expecting to be able to specify a region where not to load tiles so they're cut out?

This isn't my specific goal right now, but I have future goals where this is a capability I'd like to have. (e.g cutouts in the tileset where an architectural model will be rendered). Wouldn't it be the most flexible to allow plugins to set these values however they want?

gkjohnson:
> (complex concave shapes)

To do perform concave shape intersections you can write a function that performs the bounding volume intersections. See the premade region classes [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/e48092de3bd8cc746322d165fb01e958d68beb15/src/plugins/three/LoadRegionPlugin.js#L85-L154):

```js
plugin.registerRegion( {
  intersectsTile: ( boundingVolume ) => /* perform concave shape intersection test */,
  calculateError: ( tile, tilesRenderer ) => tile.geometricError - errorTarget + tilesRenderer.errorTarget,
} );
```

> Wouldn't it be the most flexible to allow plugins to set these values however they want?

The plugin was designed specifically to support regions to load as they are working right now. If there are other use cases I'm happy to discuss the use cases and figure out the right way to support it but I don't just want to make quick changes to make it happen because performing the kind of "boolean" operations on tile visibility being suggested is not just a matter of letting the plugin return false I don't believe. I'm also not convinced it's really needed.

If you're having trouble getting regions to work I can take a look at an example of what's not working. And we can talk through other more complex use cases when they arise.

jsulli:
After spending some more time digging into this I finally realized the fundamental misconception I had:

I expected `LoadRegionPlugin` to control which tiles load (instead of camera frustum), while *also* retaining the role the camera has in refinement. The problem I was having all along was that I was still setting a camera on the tiles renderer, so the frustum check always passed for tiles that were within the camera view.

This felt like such a natural assumption that both me and a team member came to it independently - it's probably worth specifying in the documentation that the plugin doesn't just take over _visibility_ but also _refinement_. I'm happy to make a PR for that clarification if you agree.

Now that I understand the actual function of that plugin I realize I should just make my own - since my goal is to only load tiles that are within a defined region, while also retaining the same camera-based refinement that the renderer has without any plugins.

Your response makes much more sense now, but now that I actually understand what's going wrong I still think it would give the most flexibility to allow plugins to override the camera's decisions about visibility [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/TilesRenderer.js#L954): 
```
if ( boundingVolume.intersectsFrustum( frustum ) ) {

        inView = true;
        inViewError = Math.max( inViewError, error );
        inViewDistance = Math.min( inViewDistance, distance );

}
```
Since that will allow fine control over what renders without having to re-implement any camera logic on the plugin side.

gkjohnson:
> it's probably worth specifying in the documentation that the plugin doesn't just take over visibility but also refinement. I'm happy to make a PR for that clarification if you agree.

Any help with improved documentation would be appreciated 😁 Fundamentally the goal of adding "regions" is to allow for loading tiles up to a certain error target in a specific region. For more context the plugin was added in #1005. Adding a second camera looking at a specific spot could achieve the same thing. Users asked for this feature to do things like load tiles with enough detail to perform more precise analysis in certain areas (raycasting or physics, for example) and extracting specific subsets of a tile set for export (no camera in this case).

> since my goal is to only load tiles that are within a defined region, while also retaining the same camera-based refinement (and frustum exclusion) that the renderer has without any plugins.

Can you explain what you're planning to do with this and why? Possibly in a new issue with screenshots, etc? I don't want to shut down the idea of adding something to the project to support this but I want to make sure I'm understanding what the use case is more completely. Just loading the tiles that overlap a certain region and camera frustum will result in an extremely ugly, jagged tile set.

gkjohnson:
> Initially I just built cropping with clip planes, every tile is given a set of clip planes forming a convex hull. This works great and gives nice edges to the crop over just tile exclusion, but it wastes a ton of bandwidth loading invisible data

Great - understood that you're interested in picking out and visualizing specific portions of a tileset (just the Eiffel tower and the area around it, for example) and that you've checked that the number of unnecessarily loaded tiles is meaningful.

> there are other complications with raycasting and occlusion from invisible geometry that have to be dealt with.

Can you explain this? To be clear simply scoping what tiles are considered "in frustum" will not fix raycasting - you will still have to perform something like shader-based clipping to clean up the shape up and avoid the jagged square edges from popping in and out as LoDs are swapped between. The renderer will also continue to load sibling tile data (though not to the same level of detail) since at the moment the renderer is designed to ensure there are no data gaps when shifting the camera. That means that the low LoD "active" sibling tiles as well as visible-tile edges will still be raycast against and you'll have to perform filtering on the raycasts either way. I'm not sure what you mean by "occlusion" in this case, though.

Some alternatives that would fix both issues would be to do something like run CSG operations to actually clip the geometry (and restore it when adjust the clip zones) though this has performance implications relating to tile loads.

> I can make an issue for this with more specific requirements, but that feels like asking you to solve my problems for me 😂

That's okay! If I can understand how people are using the library in more detail it makes it a lot easier for me to shape the direction it's going. I also generally enjoy hearing how it's being used in the real world 😅 If we can come up with a good scope here maybe we can modify the plugin fulfill both use cases.

jsulli:
> Can you explain this? To be clear simply scoping what tiles are considered "in frustum" will not fix raycasting - you will still have to perform something like shader-based clipping to clean up the shape up and avoid the jagged square edges from popping in and out as LoDs are swapped between.

Yep, we do shader clipping to make sure the edges are always nice. Here's an example of a clipped plan during loading:

https://github.com/user-attachments/assets/1ca9af24-58d4-4033-ae11-84e4b7605615

This is not done with a plugin because this is behavior that we had implemented in an old homeade loader, so it's just changing some materials with the tile load events, nothing fancy.

> The renderer will also continue to load sibling tile data (though not to the same level of detail) since at the moment the renderer is designed to ensure there are no data gaps when shifting the camera. That means that the low LoD "active" sibling tiles as well as visible-tile edges will still be raycast against and you'll have to perform filtering on the raycasts either way. I'm not sure what you mean by "occlusion" in this case, though.

Yeah, I see the sibling loading as well - it's small enough that I'm unconcerned though. Here's an example of clipping A) without excluding any loading:

https://github.com/user-attachments/assets/1f9e3f17-cbe1-4d31-82e1-4299a32c8eba

B) With the changes from this PR: 

https://github.com/user-attachments/assets/ea4cfd83-89c3-49cd-b174-e3f0b2cddb5d

You can see that there are some tiles that are wholly outside of the clip region that still get loaded, but overall the reduction in loading demand is huge and still a very worthwhile improvement.

Just for completeness sake, here's the full unclipped tileset.
Screenshot 2025-04-11 at 12 27 30 PM


By occlusion, I just mean basically doing a bunch of raycasting against tiles that can't be seen. It's expensive and performance improvements would be one of the main goals when clipping the tileset in this way. There also needs to be filtering of intersections to exclude any that are outside of the convex hull. This isn't difficult but does increase complexity, and we do some particular operations that do lots of horizontal raycasts through tilesets, with a convex hull with enough edges that starts to add a performance cost.

Actually cutting the mesh at the point of clip is an interesting idea, but I can't imagine the performance of that being anything above terrible.

> That's okay! If I can understand how people are using the library in more detail it makes it a lot easier for me to shape the direction it's going. I also generally enjoy hearing how it's being used in the real world 😅 If we can come up with a good scope here maybe we can modify the plugin fulfill both use cases.
Thanks for taking the time.

Thanks for hearing out what I'm looking for! I've been working on this type of problem for years at this point and you obviously have been too. It's a very fun and challenging problem space with lots of tradeoffs, but I love talking about it and figuring out ways to improve it 🙂

gkjohnson:
Great! Thanks for the videos and explanation - I think this feels like a nice usecase to support more cleanly.

Here's what I'm thinking:

- Adjust the arguments for "LoadRegionPlugin" shapes to take an argument indicating that only the shapes in the given volume are loaded (your use case)

```js
plugin.addRegion( {
    intersectsTile: () => /* ... */,
    calculateError: () => /* ... */,

    // defaults to false, need help naming this field
    intersection: true,
} );
```

- Adjust the "calculateTileViewError" function to return "false" if region is not intersected and marked as "intersection" (or whatever we should call this field). Otherwise return "null" if no regions are intersecting. Lets format this so multiple non-overlapping shapes can be marked as "intersection" so you can have multiple regions displayed / loaded.
- Adjust the calling `TilesRenderer.calculateTileViewError` function to ignore the result if it's null (to achieve the current behavior).

Hopefully that makes sense - then you can set the region to use a high error target which means the camera error target will always be used while it respects the "intersection" state:

```js
plugin.addRegion( new SphereRegion( /* errorTarget = */ Infinity, /* intersection = */ true ) );
```

Note that this _won't_ "just work" if you want to invert the shape - eg cut out a box shaped hole from the larger tile set since in that case you'd need to check for complete containment of each tile rather than just intersection.

> raycasting...  It's expensive and performance improvements would be one of the main goals when clipping the tileset in this way. 

Using something like [three-mesh-bvh](https://github.com/gkjohnson/three-mesh-bvh) on the leaf nodes may be another way to improve raycasting performance if you need to do a ton of raycasts. I think others have done this but I haven't tried it out myself.

> Actually cutting the mesh at the point of clip is an interesting idea, but I can't imagine the performance of that being anything above terrible.

I've wanted to try adding a [three-bvh-csg](https://github.com/gkjohnson/three-bvh-csg) plugin at some point which would let you perform operations on non-manifold geometry (as is often the case with tiles) with the `HOLLOW_INTERSECTION` op. It has its quirks but may be good enough, is pretty fast and these operations only need to be performed on the tiles at the edges of the shape. There's also [manifold-3d](https://www.npmjs.com/package/manifold-3d) but I don't think it works on non-manifold input.

jsulli:
Those sound like great improvements!

For the new field you describe, maybe `exclusive` would explain it
```
plugin.addRegion( {
    intersectsTile: () => /* ... */,
    calculateError: () => /* ... */,

    // Do not load tiles outside of this region, multiple regions can be exclusive
    exclusive: true,
} );
```

gkjohnson:
I was thinking about it more and "mask" might make sense, as well. We can decide during the PR.

PR #1077, opened 2025-04-15
Add missing headers to subtree fetch

jsulli:
### Changes
Just grabs the fetchOptions from the tiles renderer and gives it to the subtree loader so it can be used for fetching subtree files.


gkjohnson:
Great, thank you!

PR #1091, opened 2025-04-23
Fix implicit any in types

Methuselah96:
Some of the types in the declaration files are implicitly `any`. This PR fixes those and adds `"strict": "true"` to the `tsconfing.json` to prevent further regressions.


gkjohnson:
Thanks! Looks like not setting "strict" lets a lot of errors through.

Issue #1057, opened 2025-04-06
Typescript related overhaul


kayden1940:
Here is an early draft on what typescript related things I see worth consider updating.

Misc:
- ts version update, blocker.

Typing:
- plugins might be better to have their standalone interface, few of them have compulsory fields and respected errors should be thrown.

Testing:
After wrestling with Jest's ambiguous warnings (specifically with installed libraries and how no one is able to figure out), configs, a long list of dependence required to make it work with tsx. It seems necessary to me to consider alternative to meet the goal. Early assessment suggesting vitest and three-stdlib are in the list to explore, options are open.

Remark:
Not directly related, I am in progress on building a medium scale project in tsx to test out the library. Naturally I will find more things to discuss along the way, hence this list is not final. Welcome for anyone to join in.

kayden1940:
@gkjohnson I am thinking should the typing and testing be in this early stage. API changes are great, just that for typing and testing to catch up as the APIs refining would be a bit rough.

gkjohnson:
> I am thinking should the typing and testing be constantly updating in this early stage.

For what it's worth I don't consider the project early stage. It's being used in a number of production applications and - aside from the switch to use plugins - the basic API for creating and rendering the tile set has not changed really since the project was created 6 years ago.

That aside I don't think there's a lot of value in creating tests for testing things like "is `group` correctly defined on the TilesRenderer class". I think testing has more value in checking whether some of the complex type templating is defined correctly. Eg "is event listener typing working correctly?" "can arbitrary event types still be listened to?" "Is r3f component typing accepting and limiting the correct properties?" These are all things that have required fairly complex typing constructs and it's not obvious from looking at the definitions that all corner cases are coverd effectively and that's what I think is valuable to add tests for.

PR #1092, opened 2025-04-23
Add extensions to imports in declaration files

Methuselah96:
Including extensions in imports ensures maximal compatibility for users with different TSConfig settings. This is enforced by updating the `module` setting in the `tsconfig.json` to `node16`.


gkjohnson:
Thanks! Are there any resources you can link me to on this topic? I feel like I've typically seen typescript files use no extensions for imports. Is this incorrect?

Methuselah96:
There is some ambiguity and wiggle room, but, in general, if your JS files are using extensions, it's easiest to have your TS files use extensions as well. [Here](https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library) is some guidance on choosing compiler options for libraries, where using extensions is recommended for maximal compatibility with different types of consumers.

In my case, I have a project that is using `"module": "node16"`, so it was producing errors for this library due to lack of file extensions, hence the PR.

gkjohnson:
Thanks for the info - as you my already know, I'm not a Typescript expert. I know you've been taking the time to keep three.js' types in sync, as well, so I appreciate your contributions here. 

If you have any interest or availability the project could probably use your insight into some other issues related to typing and type validation. See #1008 and #1057.

Thank you again!

Issue #1081, opened 2025-04-17
R3F: Globe demo does not load correctly initially


gkjohnson:
Likely due to the progressive child initialization + auth plugin delays + demand r3f rendering + update on change path.

gkjohnson:
This is because the new deferred node processing system does not dispatch an event notifying any system that it needs to be rerendered.

PR #1100, opened 2025-04-24
Refactors mapTiles example to render-on-change

### LINKED ISSUE #1082: Provide example of render on Change with GlobeControls
### Question

In the Readme as well as in #112 and #118 it is stated how to use the TilesRenderer for render-on-change approach. However, I could not find any example how to use the GlobeControls without render loop. In other projects using for example OrbitControls, I am just listening to the `change` event to trigger rendering the scene. Here with GlobeControls I had some struggles.

For minimum working example, I cloned the repo and made changes to the `mapTiles.js` example.

```js
controls.addEventListener( 'change', () => {

    render();

} );

function render() {

    // requestAnimationFrame( render );

    controls.update();
    camera.updateMatrixWorld();

    tiles.errorTarget = params.errorTarget;
    tiles.setCamera( camera );
    tiles.setResolutionFromRenderer( camera, renderer );
    tiles.update();

    renderer.render( scene, camera );

}
```

However, to make it finally work, I also needed to remove `this.dispatchEvent( _changeEvent );` within the `update()` method of the `EnvironmentControls` (and add it to `wheelCallback` to react on changes when zooming).

I assume the GlobeControls/EnvironmentControls are not yet ready or have not yet been tested for working without render loop. Or how are they supposed to work for this use case (without changing their code)?

### Supplemental Data

_No response_

### Library Version

v0.4.8

### Three.js Version

r170

Brakebein:
Changes the mapTiles example to use an event-driven, render-on-change approach. This improves performance by reducing unnecessary renders. Listens for control changes, tile loading events, and window resize events to trigger rendering.

Adds rxjs dependency as devDependency for event handling and throttling.

Closes #1082

### DIFF
--- a/example/mapTiles.js
+++ b/example/mapTiles.js
@@ -10,15 +10,15 @@ import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
 let controls, scene, renderer;
 let tiles, camera;
 
-// throttled render callback to maximum of 50 FPS
-const render$ = throttle( render, 20 );
-
 const params = {
 
 	errorTarget: window.devicePixelRatio,
 	planar: false,
 
 };
+
+// debounced render function
+const debounceRender = debounce( render );
 
 init();
 render();
@@ -126,14 +126,12 @@ function initTiles() {
 	}
 
 	// listen to events to call render() on change
-	controls.addEventListener( 'change', render$ );
-	controls.addEventListener( 'end', render$ );
-	tiles.addEventListener( 'load-model', render$ );
-	tiles.addEventListener( 'force-rerender', render$ );
+	controls.addEventListener( 'change', debounceRender );
+	controls.addEventListener( 'end', debounceRender );
+	tiles.addEventListener( 'load-model', debounceRender );
+	tiles.addEventListener( 'force-rerender', debounceRender );
+	tiles.addEventListener( 'load-tile-set', debounceRender );
 
-	// workaround for initial loading
-	tiles.addEventListener( 'load-tile-set', render );
-	
 	render();
 
 }
@@ -146,7 +144,7 @@ function onWindowResize() {
 
 	renderer.setSize( window.innerWidth, window.innerHeight );
 
-	render$();
+	debounceRender();
 
 }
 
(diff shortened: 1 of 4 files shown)


gkjohnson:
Thanks for making a PR - do you mind adjusting it so it doesn't require any dependencies like rxjs? It should be easy enough to make a custom debounced render function:

```js
function debounce( callback ) {

  let scheduled = false;
  return () => {

    if ( ! scheduled ) {

      queueMicrotask( () => {
        
        called = false;
        callback();

      } );

    }

  };
}
```

Brakebein:
It needs to be a throttled render function instead of debounced one, otherwise we won't properly see any updates. I changed it now to vanilla js. It constantly renders as long there are controls or tiles updates, and stops rendering when there are no update events triggered anymore.

example/mapTiles.js
```diff
@@ -122,6 +125,17 @@ function initTiles() {
...
 	}
 
+	// listen to events to call render() on change
+	controls.addEventListener( 'change', render$ );
+	controls.addEventListener( 'end', render$ );
+	tiles.addEventListener( 'load-model', render$ );
+	tiles.addEventListener( 'force-rerender', render$ );
+
+	// workaround for initial loading
+	tiles.addEventListener( 'load-tile-set', render );
```
gkjohnson, line 135: Why do you feel this is a workaround? There's no reason to not debounce this, as well.

gkjohnson:
> It needs to be a throttled render function instead of debounced one, otherwise we won't properly see any updates.

This is indicative of bug - specifically #1081. Debouncing should work just fine so I've adjusted the demo to use it and I'll review this PR once I've fixed the #1081.

package-lock.json
```diff
@@ -5597,6 +5597,19 @@
...
+      "integrity": "sha512-hTdwr+7yYNIT5n4AMYp85KA6yw2Va0FLa3Rguvbpa4W3I5xynaBZo41cM3XM+4Q6fRMj3sBYIR1VAmZMXYJvRQ==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "tslib": "^1.9.0"
+      },
+      "engines": {
+        "npm": ">=2.0.0"
+      }
+    },
```
gkjohnson, line 5612: Lets revert the package-lock.json changes.

Brakebein:
> > It needs to be a throttled render function instead of debounced one, otherwise we won't properly see any updates.
> 
> This is indicative of bug - specifically #1081. Debouncing should work just fine so I've adjusted the demo to use it and I'll review this PR once I've fixed the #1081 issue.

Okay, this is a more elegant approach using requestAnimationFrame(). This ensures that the maximum FPS is used. The workaround I had in the earlier approach somehow solved the initial loading problem, which is most certainly due to the issue you want to solve in #1081.

Regarding _debounce_ vs _throttle_ (which is probably only a naming issue): from my understanding, your solution also aims to throttle the render calls, i.e., to reduce the number of calls to the maximum number of FPS.
- When _throttle_ is being called it fires your function periodically (eg. every 20ms).
- _Debounce_ only fires at the end (eg. 20ms after the last event).

[Difference between throttling and debouncing a function](https://stackoverflow.com/questions/25991367/difference-between-throttling-and-debouncing-a-function)

gkjohnson:
I just fixed #1081 and adjusted this PR with the new event needed so it should be good now. The new event will be available in the new release.

And you're right, I'll also change the function name to "throttle" since it's not resetting a timer every call.

PR #1121, opened 2025-05-09
Fix call to updateWorldMatrix not updating appropriate matrices

### LINKED ISSUE #1119: TilesRenderer: Calling updateWorldMatrix( updateParents, updateChildren ) results in out of sync matrices
Since `updateWorldMatrix` will not call `updateMatrixWorld` where the smart updates are implemented. Could likely add an override function like so:

```js
updateWorldMatrix( updateParents, updateChildren ) {

  super.updateWorldMatrix( updateParents, false );

  if ( updateChildren ) {

    this.updateMatrixWorld( true );

  }

}
```

See https://github.com/repalash/threepipe/issues/13#issuecomment-2860873541

gkjohnson:
Fix #1119

### DIFF
--- a/src/three/TilesGroup.js
+++ b/src/three/TilesGroup.js
@@ -91,4 +91,16 @@ export class TilesGroup extends Group {
 
 	}
 
+	updateWorldMatrix( updateParents, updateChildren ) {
+
+		if ( this.parent && updateParents ) {
+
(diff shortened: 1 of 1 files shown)


raiderss:
Very clean code with good test coverage. This is a great example of best practices.

PR #1123, opened 2025-05-11
TS - Extend from base Loader classes in loaders, three.js loaders

repalash:
Hi, 
The loaders in typescript don't extend from the LoaderBase, but they do in javascript, as a result some properties like workingPath were not accessible from the three.js loader like `B3DMLoader` in typescript. 

In this, updated the classes to extend from the correct base class, and added generic types for return types of `load` and `parse` functions which seem to be the only properties that changed.

Since it was a very quick change, didn't create an issue first, let me know if its okay.


gkjohnson:
Great, thank you!

PR #1128, opened 2025-05-13
Add additional properties to EnvironmentControls types.

repalash:
Added some properties that were missing and from the types. 

`up` and `fallbackPlane` are required to be set from outside like in the `deepZoom` example. Added some others that seemed relevant. 

`camera`, `scene` and `tilesRenderer` have setters, so can be made readonly in types or removed if you don't want to expose it in the api.

### DIFF
--- a/src/three/controls/EnvironmentControls.d.ts
+++ b/src/three/controls/EnvironmentControls.d.ts
@@ -1,4 +1,4 @@
-import { Camera, EventDispatcher, Object3D, Vector3 } from 'three';
+import { Camera, Clock, EventDispatcher, Object3D, Plane, Raycaster, Vector3 } from 'three';
 import { TilesRenderer } from '../TilesRenderer.js';
 
 export interface EnvironmentControlsEventMap {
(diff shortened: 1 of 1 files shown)


src/three/controls/EnvironmentControls.d.ts
```diff
@@ -11,6 +11,12 @@ export class EnvironmentControls extends EventDispatcher<EnvironmentControlsEven
 
 	readonly isEnvironmentControls: true;
 
+	domElement: HTMLElement;
+	camera: Camera;
+	scene: Object3D;
+	tilesRenderer: TilesRenderer;
```
gkjohnson, line 17: Lets mark these as "readonly" for now.

repalash:
Done the changes. 
Since reorientOnDrag and scaleZoomOrientationAtEdges are used in GlobeControls, they might be required if someone wants to implement in TS later. Removed from now, i guess can be added later if someone asks

gkjohnson:
Thank you!

PR #1130, opened 2025-05-13
Fix and improve typings for LoadRegionPlugin

xawill:
Experimenting with the new `LoadRegionPlugin` in Typescript, I noticed a few issues and missing things. Let me know if any rework is required from me, otherwise it can be merged as far as I am concerned. Thanks!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,6 +10,7 @@ export * from './three/math/EllipsoidRegion.js';
 export * as GeoUtils from './three/math/GeoUtils.js';
 export * from './three/math/GeoConstants.js';
 export * from './three/math/OBB.js';
+export * from './three/math/TileBoundingVolume.js';
 
 // three.js controls
 export { GlobeControls } from './three/controls/GlobeControls.js';
--- a/src/index.js
+++ b/src/index.js
@@ -10,7 +10,6 @@ export * from './three/math/EllipsoidRegion.js';
 export * as GeoUtils from './three/math/GeoUtils.js';
 export * from './three/math/GeoConstants.js';
 export * from './three/math/OBB.js';
-export * from './three/math/TileBoundingVolume.js';
 
 // three.js controls
 export { GlobeControls } from './three/controls/GlobeControls.js';
--- a/src/plugins/three/LoadRegionPlugin.d.ts
+++ b/src/plugins/three/LoadRegionPlugin.d.ts
@@ -19,7 +19,6 @@ export class LoadRegionPlugin {
 	removeRegion( region: BaseRegion ): void;
 	hasRegion( region: BaseRegion ): void;
 	clearRegions(): void;
-	dispose(): void;
 
 }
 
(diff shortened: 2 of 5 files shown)


src/index.js
```diff
@@ -10,6 +10,7 @@ export * from './three/math/EllipsoidRegion.js';
 export * as GeoUtils from './three/math/GeoUtils.js';
 export * from './three/math/GeoConstants.js';
 export * from './three/math/OBB.js';
+export * from './three/math/TileBoundingVolume.js';
```
gkjohnson, line 13: What's the motivation for exporting this class? I think we can just export the types and not the class itself. The "Tile" object for each node is still fairly wild west - at some point it's going to make sense to go back and make that a bit more well-defined now that the project has solidified more.

src/plugins/three/LoadRegionPlugin.d.ts
```diff
@@ -4,8 +4,9 @@ import { TilesRenderer } from '../../three/TilesRenderer.js';
 import { OBB } from '../../three/math/OBB.js';
 import { Sphere, Ray } from 'three';
 
-declare class Region {
+declare class BaseRegion {
```
gkjohnson, line 7: This should probably be declared as an "interface" rather than a class. It's okay to pass an object into the region functions:

```js
plugin.addRegion( {
  errorTarget: 10,
  intersectsTile: /* ... */
} )
```

src/plugins/three/LoadRegionPlugin.d.ts
```diff
@@ -14,28 +15,29 @@ declare class Region {
...
 export class LoadRegionPlugin {
 
-	addRegion( region: Region ): void;
-	removeRegion( region: Region ): void;
-	hasRegion( region: Region ): void;
+	addRegion( region: BaseRegion ): void;
+	removeRegion( region: BaseRegion ): void;
+	hasRegion( region: BaseRegion ): void;
 	clearRegions(): void;
+	dispose(): void;
```
gkjohnson, line 22: Dispose is not a public function.

src/plugins/three/LoadRegionPlugin.js
```diff
@@ -83,7 +83,7 @@ export class LoadRegionPlugin {
 }
 
 // Definitions of predefined regions
-class BaseRegion {
+export class BaseRegion {
```
gkjohnson, line 86: What's the motivation for exporting this?

src/index.js
xawill: You are right

src/plugins/three/LoadRegionPlugin.d.ts
xawill: Interfaces in TypeScript cannot have constructors. I could create both: `export class BaseRegion implements IRegion` and have `addRegion(region: IRegion): void;`?

xawill:
I am not sure why Node.js CI / build 18.x is failing. Probably not linked to this PR, is it?

src/plugins/three/LoadRegionPlugin.d.ts
gkjohnson: A constructor shouldn't be important for the object being passed to the class and better captures the original design intent of the function, I think. But these kinds typing decisions are not really where I enjoy spending my time, unfortunately, so I'll have to lean on those who are more familiar with using them 😅

gkjohnson:
Thanks! Let's merge this -

> I am not sure why Node.js CI / build 18.x is failing. Probably not linked to this PR, is it?

No it's not - there's a flaky test related to region bounding volumes that I need to fix at some point.

PR #1131, opened 2025-05-13
Fix typings for QuantizedMeshPlugin

xawill:
Thank you for this new `QuantizedMeshPlugin` plugin! Experimenting with it in Typescript, I noticed the exports were missing. Let me know if any rework is required from me, otherwise it can be merged as far as I am concerned. Thanks!


gkjohnson:
Thanks!

PR #1132, opened 2025-05-13
Fix various typings

xawill:
I am not sure exactly what is public API, but I think at least the typings for Ellipsoid were just missing.

For the `cameras` of `TilesRenderer`, I am not sure if this is supposed to be exposed. My use case was to compute the camera elevation in the `intersectsTile` function of a custom `BaseRegion` from `LoadRegionPlugin`. Let me know if you see a better way to do it.

### DIFF
--- a/src/three/TilesRenderer.d.ts
+++ b/src/three/TilesRenderer.d.ts
@@ -22,6 +22,7 @@ export interface TilesRendererEventMap {
 export class TilesRenderer<TEventMap extends TilesRendererEventMap = TilesRendererEventMap> extends TilesRendererBase implements EventDispatcher<TEventMap> {
 
 	ellipsoid: Ellipsoid;
+	cameras: Camera[];
 	autoDisableRendererCulling : boolean;
 	optimizeRaycast : boolean;
 
--- a/src/three/math/Ellipsoid.d.ts
+++ b/src/three/math/Ellipsoid.d.ts
@@ -1,4 +1,4 @@
-import { Vector3, Matrix4 } from 'three';
+import { Vector3, Matrix4, Ray } from 'three';
 
 export enum Frames {}
 export const ENU_FRAME: Frames;
@@ -16,6 +16,7 @@ export class Ellipsoid {
 	getCartographicToNormal( lat: number, lon: number, target: Vector3 ): Vector3;
 	getPositionToNormal( pos: Vector3, target: Vector3 ): Vector3;
 	getPositionToSurfacePoint( pos: Vector3, target: Vector3 ): Vector3;
+	getPositionElevation( pos: Vector3 ): number;
 
 	getAzElRollFromRotationMatrix(
 		lat: number, lon: number, rotationMatrix: Matrix4,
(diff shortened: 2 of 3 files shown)


src/three/math/Ellipsoid.d.ts
```diff
@@ -34,4 +35,12 @@ export class Ellipsoid {
 	getEastNorthUpFrame( lat: number, lon: number, target: Matrix4 ): Matrix4;
 	getEastNorthUpAxes( lat: number, lon: number, vecEast: Vector3, vecNorth: Vector3, vecUp: Vector3, point?: Vector3 ): void;
 
+	intersectRay( ray: Ray, target: Vector3 ): Vector3 | null;
+
+	calculateHorizonDistance( latitude: number, elevation: number ): number;
+	calculateEffectiveRadius( latitude: number ): number;
+
```
gkjohnson, line 42: These functions were added for GlobeControls and I don't want to consider them public at the moment. If they're going to be made public I'd like to consider some more descriptive names for the functions.

gkjohnson:
Closing for now

src/three/math/Ellipsoid.d.ts
xawill: I am using `calculateEffectiveRadius` in trigonometry calculations (linear distance between two GPS coordinates) for an animation of globe rotation to know if the place the camera is currently looking at is within a certain tolerance radius of the animation destination. If it is in the radius, the camera doesn't move; if it is outside, the camera animates.
See example [here](https://github.com/xawill/earth-3d-viewer/blob/cda47a99914d3f8f01baa4680bfdd7ee898a0905/src/app/viewer/viewer.component.ts#L337)

src/three/TilesRenderer.d.ts
```diff
@@ -22,6 +22,7 @@ export interface TilesRendererEventMap {
 export class TilesRenderer<TEventMap extends TilesRendererEventMap = TilesRendererEventMap> extends TilesRendererBase implements EventDispatcher<TEventMap> {
 
 	ellipsoid: Ellipsoid;
+	cameras: Camera[];
```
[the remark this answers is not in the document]
xawill: Not exactly. I am using the standard `LoadRegionPlugin` but with a custom region. In `intersectsTile`, I would like to have the camera elevation to only activate the region if the camera is close enough to the ground.
To be concrete, the custom region is the country of Switzerland for which is available a more detailed tileset than Google's and I would like Google's entire globe for visual consistency at high level view and once we zoom closer to Switzerland have another dataset for which I use `LoadRegionPlugin` to load.
See example code [here](https://github.com/xawill/earth-3d-viewer/blob/cda47a99914d3f8f01baa4680bfdd7ee898a0905/src/app/utils/OutsideSwitzerlandRegion.ts#L45).

xawill:
Thank you @gkjohnson for taking a look and sorry for not getting back to you earlier.

I saw some of the changes were already fixed in #1174, but there is an error there I am now fixing.

 I also answered your comments. I am now rebasing this PR; would you mind reopening it? Thx!

xawill:
@gkjohnson In case you missed my comment above, any chance reopening the PR please?

gkjohnson:
Sorry I missed this previously - it looks like I'm not able to reopen the PR (see the grayed out button below). Can you make a new one?

image

Issue #1102, opened 2025-04-26
Test of Quantized mesh on some data


MatthiasZdravkovic:
Hi, thanks again for the help : https://x.com/garrettkjohnson/status/1915806000244052059

Can you test this quantized data ?

Tell me if it works @gkjohnson 

[output.zip](https://github.com/user-attachments/files/19918123/output.zip)

gkjohnson:
Thanks! Do you mind posting a picture of what you expect the data to look like? I'll take a look some time next week.

MatthiasZdravkovic:
Sure, no problem ! 
Below are pictures from Blender. This is Philip's Island, in Victoria's state in Australia

I converted my DEM file using this [tool](https://github.com/Gaia3D/mago-3d-terrainer)

Let me know if something is wrong with the data.
Thanks again

![Image](https://github.com/user-attachments/assets/d67ead79-6445-4a7b-a92a-a30a290e90d2)

![Image](https://github.com/user-attachments/assets/8abd2a7b-10cf-4820-9749-88449c2b9bf6)

xawill:
You can also test officiel terrain of Switzerland with [this tileset](https://3d.geo.admin.ch/ch.swisstopo.terrain.3d/v1/layer.json). More info about the data set [here](https://api3.geo.admin.ch/services/sdiservices.html#terrain-service).

It is supposed to look like something like this:

<img width="1480" alt="Image" src="https://github.com/user-attachments/assets/71474deb-8d64-4cf6-a807-7038a12bdaec" />

xawill:
I tested locally and I can confirm the officiel terrain of Switzerland loads perfectly fine with the  minor changes proposed in #1134.

I even managed to texture the terrain with the Swiss official orthophotos, using the `TextureOverlayPlugin`, even though I currently have a few issues to load everything efficiently. This is also not the highest available quality of texture, since it takes the highest zoom level of the quantized mesh terrain and not the one of the orthophoto currently. WIP.

<img width="1510" alt="Capture d’écran 2025-05-15 à 00 14 34" src="https://github.com/user-attachments/assets/5e57a938-f925-4fb4-888c-5af7c6b3074c" />

gkjohnson:
Thanks for the data, guys - testing the Swiss data set I'm seeing the data look quite nice:

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/856db44c-4857-4dd0-94c3-351384ac0b4e" />

[...]

PR #1134, opened 2025-05-14
Make QuantizedMeshPlugin more tolerant

xawill:
As mentioned [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1102#issuecomment-2854753190), I wanted to test the new `QuantizedMeshPlugin` with the officiel terrain of Switzerland ([layer.json](https://3d.geo.admin.ch/ch.swisstopo.terrain.3d/v1/layer.json)).

These changes allow the tileset to render, as the layer.json file properties are not exhaustive and errors where raised otherwise.

I even managed to texture the terrain with the Swiss official orthophotos, using the `TextureOverlayPlugin`, even though I currently have a few issues to load everything efficiently. This is also not the highest available quality of texture, since it takes the highest zoom level of the quantized mesh terrain and not the one of the orthophoto currently. WIP.

Capture d’écran 2025-05-15 à 00 14 34

### DIFF
--- a/src/plugins/three/QuantizedMeshPlugin.js
+++ b/src/plugins/three/QuantizedMeshPlugin.js
@@ -20,7 +20,7 @@ function isAvailable( layer, level, x, y ) {
 		available,
 	} = layer;
 
-	if ( level >= minzoom && level <= maxzoom && level < available.length ) {
+	if ( level >= ( minzoom ?? 0 ) && level <= ( maxzoom ?? Infinity ) && level < available.length ) {
 
 		// TODO: consider a binary search
 		const availableSet = available[ level ];
@@ -150,7 +150,7 @@ export class QuantizedMeshPlugin {
 
 				this.layer = json;
 
-				if ( json.extensions.length > 0 ) {
+				if ( json.extensions?.length > 0 ) {
 
 					tiles.fetchOptions.header[ 'Accept' ] += `;extensions=${ json.extensions.join( '-' ) }`;
 
--- a/src/plugins/three/QuantizedMeshPlugin.js
+++ b/src/plugins/three/QuantizedMeshPlugin.js
@@ -15,12 +15,12 @@ const _vec = /* @__PURE__ */ new Vector3();
 function isAvailable( layer, level, x, y ) {
 
 	const {
-		minzoom,
-		maxzoom,
+		minzoom = 0,
+		maxzoom = Infinity,
 		available,
 	} = layer;
 
-	if ( level >= ( minzoom ?? 0 ) && level <= ( maxzoom ?? Infinity ) && level < available.length ) {
+	if ( level >= minzoom && level <= maxzoom && level < available.length ) {
 
 		// TODO: consider a binary search
 		const availableSet = available[ level ];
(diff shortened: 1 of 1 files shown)


src/plugins/three/QuantizedMeshPlugin.js
```diff
@@ -20,7 +20,7 @@ function isAvailable( layer, level, x, y ) {
 		available,
 	} = layer;
 
-	if ( level >= minzoom && level <= maxzoom && level < available.length ) {
+	if ( level >= ( minzoom ?? 0 ) && level <= ( maxzoom ?? Infinity ) && level < available.length ) {
```
gkjohnson, line 23: I didn't realize these fields could be undefined. Lets default them above instead of inlining these values:

```js
	const {
		minzoom = 0,
		maxzoom = Infinity,
		available,
	} = layer;
```

src/plugins/three/QuantizedMeshPlugin.js
```diff
@@ -150,7 +150,7 @@ export class QuantizedMeshPlugin {
 
 				this.layer = json;
 
-				if ( json.extensions.length > 0 ) {
+				if ( json.extensions?.length > 0 ) {
```
gkjohnson, line 153: Similar to above, lets extract all the fields of the json into local variables and default them. I think it makes it easier to follow what's happening:

```js
const {
  bounds,
  projection = 'EPSG:4326',
  extensions = []
} = json;
```

I don't suppose you know of a specification for QMesh's layer.json anywhere? It seems fairly adhoc.

gkjohnson (commented):
Amazing! The screenshot looks great.

> I even managed to texture the terrain with the Swiss official orthophotos, using the TextureOverlayPlugin, even though I currently have a few issues to load everything efficiently.

It's still in the plans to load and overlay something like TMS tiles on top of the QMesh (or any other) data sets but that's in progress. If you have any comments on what kinds of performance issues you're running into that would be helpful.

xawill:
Yeah, I saw it was WIP, but thought I would try it out with `TextureOverlayPlugin` I had just discovered. 

Let me open a dedicated issue to improve `QuantizedMeshPlugin` when I have a shareable codebase and live demo (in a few hours/days).

Note that in my case, the qm-terrain layer is TMS and the orthophotos are WMTS. I just had to flip the y index:

```
const [, z, x, y] = tileUrl.match(/\/(\d+)\/(\d+)\/(\d+)\.terrain/)!;
const wmtsFlippedY = Math.pow(2, parseInt(z)) - 1 - parseInt(y);
```

src/plugins/three/QuantizedMeshPlugin.js
xawill: Actually, I think `maxzoom` should not be present in the layer.json. As you did in your `maxLevel` getter, it should be `layer.available.length`. But let's keep it as you proposed just above. `maxLevel` and `layer.available.length` are not semantically exactly the same thing.

gkjohnson: `maxZoom` is present in the in the layer.json file but it doesn't necessarily seem to match `available.length`. There's no spec afaik for this file beyond the CesiumTerrainProvider file you linked below. The `maxZoom` field [is used in that file](https://github.com/CesiumGS/cesium/blob/46ed67c0bf601b52f3b1acdb7bb13d5decc5a181/Source/Core/CesiumTerrainProvider.js#L220-L221) but it's not clear how it differs from `available.length`, unfortunately. We may have to ask someone from Cesium at some point.

gkjohnson:
Thanks for the changes!

> Note that in my case, the qm-terrain layer is TMS and the orthophotos are WMTS. I just had to flip the y index:

Is there much more to TMS vs WMTS? Would it be easy to make a `WMTSTilesPlugin` based on something like the [TMSTilesPlugin](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/plugins/three/images/EPSGTilesPlugin.js#L55)?

Issue #1133, opened 2025-05-14
Move optionalDependencies to peerDependencies


mqrty:
### Description

Currently react, react-dom and @react-three/fiber are listed under optionalDependencies in the package.json. When a host project also has these as dependencies, this can cause problems. An example would be two React instances being loaded at runtime, leading to failures when using any 3DTilesRendererJS R3F components.

### Solution

Convert the three mentioned packages to peerDependencies

### Alternatives

Using bundler aliases to force imports of the three packages from the host apps node_modules but that's more of a workaround.

gkjohnson:
After digging into it a bit more it sounds like `optionalDependencies` don't really behave as the name implies. Moving the packages to peerDependencies and adding some options to `peerDependenciesMeta` would be best (see [here](https://stackoverflow.com/questions/62047806/how-do-i-set-a-peer-dependency-optional)):

```js
"peerDependenciesMeta": {
  "react": {
    "optional": true
  }
}
```

Would you like to make a PR?

PR #1135, opened 2025-05-15
Fix optional dependencies

mqrty:
Declared react, react-dom and @react-three/fiber as optional peerDependencies as discussed in #1133


gkjohnson:
Great, thank you!

Issue #1137, opened 2025-05-19
Event typing inconsistencies


mqrty:
### Description

I noticed issues with some of the event types:

#### In [TilesRenderer.d.ts](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/TilesRenderer.d.ts#L7-L20) and [TilesRenderer.test.ts](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/test/types/TilesRenderer.test.ts):
- missing 'needs-update'
- missing 'load-error'

#### In the [README](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/README.md?plain=1#L308-L339):
- missing 'add-camera', 'delete-camera' and 'camera-resolution-change'
- 'tile-visibility-change' is missing `visible: boolean`
- 'load-content' falsely has `tileSet` and `url`, these are never dispatched, in the test.ts it is also using emptyEvent (see [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/test/types/TilesRenderer.test.ts#L27))
- 'needs-update' also has `tileSet` and `url` but i don't see these being passed when dispatching
- 'load-error' has `string | URL`  as type for `url` but `url` is not passed when dispatching this for the root tileset (see [here](https://github.com/mqrty/3DTilesRendererJS/blob/66a098aadcf25485157a0597ef3d44e4b0dde891/src/base/TilesRendererBase.js#L309-L313))

### Solution

Resolve the issues listed above. Should i create a PR?

### Alternatives

Accept missing/wrong types or use `any` type :D

gkjohnson:
Hello! Yes a PR would be great - thanks for the catch.

PR #1139, opened 2025-05-21
Fix event typing inconsistencies

mqrty:
Hi, heres what i came up with to fix the event typing inconsistencies mentioned in #1137.


gkjohnson:
Great, thank you!

PR #1140, opened 2025-05-21
Fix load-tile-set event content to match type and docs

dbuck:
Similar to #1139 - we were using the url in the old onLoadTileset callback, and while it's in the type/docs for the new event, it's not passed in the code.

```
// Fired when a new root or child tile set is loaded.
{ type: 'load-tile-set', tileSet: Object, url: String }
```

Is this correct? at a glance in TilesRendererBase, it doesn't look like it fires when child tile set's are loaded, and if it DOES, then the url I added here is incorrect? 👋 catching up on a few years of changes - and our existing tileset's don't get past the root yet.

### DIFF
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -297,6 +297,7 @@ export class TilesRendererBase {
 					this.dispatchEvent( {
 						type: 'load-tile-set',
 						tileSet: root,
+						url: this.rootURL,
 					} );
 
 				} )


gkjohnson:
Hey Dave! Thanks for the fix and good to see you back 😁 

> at a glance in TilesRendererBase, it doesn't look like it fires when child tile set's are loaded

it looks like I changed this in https://github.com/NASA-AMMOS/3DTilesRendererJS/commit/f9875270be6c4d3d1df93720c19d5640e5f8bb31 and neglected to keep the event firing for children tiles. We can add the `load-tile-set` event [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/af9d7390b56e88a621467ad7ece025d90e87477b/src/base/TilesRendererBase.js#L886-L898) if `isExternalTileSet` has been set to true if you'd like to make that change. Otherwise I can take a look in another PR.

> then the url I added here is incorrect? 

It's unfortunately _not_ exactly correct. The plugins architecture has allowed for a lot of flexibility but also means we don't always know a URL that's being loaded. In fact there may not be a single proper tile set URL that's really loaded. As two examples:
- The Cesium Ion Plugin constructs a URL on the fly and makes _two_ requests for the tile set.
- The Quantized Mesh Plugin and all tiled image plugins construct the root tile set structure implicitly.

Though I think it's fair to consider these corner cases.

Additonally, before loading any data [all the URLs are preprocessed](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/af9d7390b56e88a621467ad7ece025d90e87477b/src/base/TilesRendererBase.js#L664-L665) to add query parameters, etc as needed. This URL isn't exposed in the "load root tile set" phase due to the above cases but if we want to add a url here assuming the typical, non-null root url cases we should preprocess it again (assuming all these functions are deterministic, which they should be):

```js
let processedUrl = this.rootURL;
if ( processedUrl !== null ) {
  
  this.invokeAllPlugins( plugin => processedUrl = plugin.preprocessURL ? plugin.preprocessURL( processedUrl, null ) : processedUrl );

}

// ...

this.dispatchEvent( {  type: 'load-tile-set', url: processedUrl /* ... */ } );
```

> our existing tileset's don't get past the root yet.

Are you able to share any data sets? How are you using these functions?

---

Apologies for letting the docs fall out of data and missing the function. It's been a lot to keep up on with the time I have available. I think the project is due for some proper testing infrastructure. Someday 🤞

dbuck:
I sanity checked with a dataset just using the examples and it's working fine. 

> How are you using these functions?

The problem is in one of our monkeypatches. we had a few of them to insert url signing and modify error based on app context. At a glance it looks like most (everything?) of what we're doing can be shifted to Plugins.

I'll try and write something up when I get back to a real keyboard.

> Are you able to share any data sets?

I need to check, I think we should be able to after I find or make a scan I own. it would be nice to get something static up that doesn't require our auth.

dbuck:
> In fact there may not be a single proper tile set URL that's really loaded. As two examples:
> The Cesium Ion Plugin constructs a URL on the fly and makes two requests for the tile set.
> The Quantized Mesh Plugin and all tiled image plugins construct the root tile set structure implicitly.

Ok that's interesting! We were just using the url for informational/debugging purposes so I can just as easily swap the change to be a PR to remove the url from the docs/tests/types to align with current behaviors?

gkjohnson:
I think it's useful to have the URL for debugging, as you mention, so why don't we add it back. In some cases it may not be reliable but we can deal with that if it becomes a problem:

- Preprocess the root url again after the root tile set has been loaded to include it in the event (as demonstrated above), which will work in the common case, I think.
- Add a "load-tile-set" event when the child tile sets are loaded [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/af9d7390b56e88a621467ad7ece025d90e87477b/src/base/TilesRendererBase.js#L886-L898).

src/base/TilesRendererBase.d.ts
```diff
@@ -19,6 +19,7 @@ export class TilesRendererBase {
 	lruCache : LRUCache;
 	parseQueue : PriorityQueue;
 	downloadQueue : PriorityQueue;
+	processNodeQueue: PriorityQueue;
```
dbuck, line 22: snuck another type fix in - documented in readme, but wasn't here

gkjohnson:
Great, thanks!

Regarding your new issues - once you have some kind of repro let me know. It's not exactly clear to me what's happening in your example snippets but ideally you shouldn't have to force traversal.

PR #1142, opened 2025-05-21
Add default type for RTC_CENTER

### LINKED ISSUE #1141: RTC_CENTER missing default component type
### Description

Hello again, i noticed the following error occurs when trying to load tilesets that use RTC_CENTER in the tile content Feature Table.

```
Error: FeatureTable : Feature type not provided for "RTC_CENTER".
    at parseBinArray (FeatureTable.js:25:10)
    at FeatureTable.getData (FeatureTable.js:137:17)
    at I3DMLoader.js:87:39
    at GLTFLoader.js:2686:5
```

Explicitly providing the feature/component type should not be necessary according to the 3D Tiles specs (both [1.0](https://docs.ogc.org/cs/18-053r2/18-053r2.html#103) and [1.1](https://docs.ogc.org/cs/22-025r4/22-025r4.html#toc26)).

> The semantic defines the allowed data type, e.g., when "POSITION" in Instanced 3D Model refers to the binary body, the component type is FLOAT and the number of components is 3.

I will submit a PR soon with proposed changes.

### Reproduction steps

Try to load a tileset that has tile content using RTC_CENTER and does not explicitly specify the component type.

### Code

No specific settings required

### Live example

No live example, sorry

### Screenshots & Tile Set

_No response_

### Library Version

0.4.9

### Three.js Version

0.170.0

mqrty:
Here are the proposed changes to resolve #1141

### DIFF
--- a/src/three/loaders/B3DMLoader.js
+++ b/src/three/loaders/B3DMLoader.js
@@ -55,7 +55,7 @@ export class B3DMLoader extends B3DMLoaderBase {
 				const { batchTable, featureTable } = b3dm;
 				const { scene } = model;
 
-				const rtcCenter = featureTable.getData( 'RTC_CENTER' );
+				const rtcCenter = featureTable.getData( 'RTC_CENTER', 1, 'FLOAT', 'VEC3' );
 				if ( rtcCenter ) {
 
 					scene.position.x += rtcCenter[ 0 ];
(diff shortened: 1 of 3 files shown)


gkjohnson:
Great, thank you! I don't have a data set to test with so I trust this is working with whichever tile data sets you have available.

mqrty:
If you still want to test it you can try with this simple i3dm tileset: [3dtilesrendererjs-i3dm-rtc-center.zip](https://github.com/user-attachments/files/20382940/3dtilesrendererjs-i3dm-rtc-center.zip)

In the kitchen sink it looks like this after my changes:

image

PR #1159, opened 2025-05-30
examples: add UNLIT option to customMaterial, and add url param options to specify initial up/material

dbuck:
customMaterial:
* add an UNLIT option that uses the MeshBasicMaterial
* enable tilesets that use draco/ktx
* url param options to specify initial up and initial material ex: `http://localhost:5173/customMaterial.html?up=Z&material=GRADIENT` would start with +Z up and with the GRADIENT material active

index:
* url param options to specify initial up ex: `http://localhost:5173/?up=Z`

Basically just enough that I can do `http://localhost:5173/customMaterial.html?up=Z&material=UNLIT#https://raw.githubusercontent.com/matterport/3DTilesSampleData/refs/heads/main/uskTEa4gEb5/mesh_tiles/4.json` without having to swap options in the menu every refresh while tinkering and trying to make a repro case! 😅 

Alternately, if you'd prefer to keep the changeMaterial simple and not have the GLTF plugin stuff in the example, I could add an unlit material type to the base `DebugTilesPlugin` and then it would just be use index and keep the `?up=Z` change?

### DIFF
--- a/src/plugins/three/DebugTilesPlugin.js
+++ b/src/plugins/three/DebugTilesPlugin.js
@@ -1,4 +1,4 @@
-import { Box3Helper, Group, MeshStandardMaterial, PointsMaterial, Sphere, Color } from 'three';
+import { Box3Helper, Group, MeshStandardMaterial, PointsMaterial, Sphere, Color, MeshBasicMaterial } from 'three';
 import { SphereHelper } from './objects/SphereHelper.js';
 import { EllipsoidRegionLineHelper } from './objects/EllipsoidRegionHelper.js';
 import { traverseAncestors, traverseSet } from '../../base/traverseFunctions.js';
@@ -449,7 +449,7 @@ export class DebugTilesPlugin {
 
 						} else if ( colorMode === UNLIT ) {
 
-							c.material = new MeshStandardMaterial( {
+							c.material = new MeshBasicMaterial( {
 								color: c[ ORIGINAL_MATERIAL ].color,
 								flatShading: c[ ORIGINAL_MATERIAL ].flatShading,
 								map: c[ ORIGINAL_MATERIAL ].map
--- a/src/plugins/three/DebugTilesPlugin.js
+++ b/src/plugins/three/DebugTilesPlugin.js
@@ -41,7 +41,6 @@ const RANDOM_COLOR = 7;
 const RANDOM_NODE_COLOR = 8;
 const CUSTOM_COLOR = 9;
 const LOAD_ORDER = 10;
-const UNLIT = 11;
 
 const ColorModes = Object.freeze( {
 	NONE,
@@ -55,7 +54,6 @@ const ColorModes = Object.freeze( {
 	RANDOM_NODE_COLOR,
 	CUSTOM_COLOR,
 	LOAD_ORDER,
-	UNLIT,
 } );
 
 export class DebugTilesPlugin {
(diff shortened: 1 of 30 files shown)


gkjohnson:
> Alternately, if you'd prefer to keep the changeMaterial simple and not have the GLTF plugin stuff in the example, I could add an unlit material type to the base DebugTilesPlugin and then it would just be use index and keep the ?up=Z change?

Overall I think these are find additions but I think it would be best to add an "unlit" option to DebugTilesPlugin and keep these changes relegated to the index.html file so the other examples are kept simple. There are a number times where the standard material causes issues since tiles are often tested without lighting, anyway.

dbuck:
@gkjohnson updated, just re-using the originalMaterial map/color/flatShading seemed best?

url param takes in the string part of the enum so that it connects correctly with the debug ui ex: `&colorMode=RANDOM_COLOR` or `&colorMode=UNLIT` etc

src/plugins/three/DebugTilesPlugin.js
```diff
@@ -445,6 +447,14 @@ export class DebugTilesPlugin {
...
 							c.material = pointsMaterial;
 
+						} else if ( colorMode === UNLIT ) {
+
+							c.material = new MeshStandardMaterial( {
+								color: c[ ORIGINAL_MATERIAL ].color,
+								flatShading: c[ ORIGINAL_MATERIAL ].flatShading,
+								map: c[ ORIGINAL_MATERIAL ].map
+							} );
+
```
gkjohnson: What's the goal of the "UNLIT" flag? Using a MeshStandardMaterial here still result in a dark model when there are no lights. And none of the other debug options will show up with no lights, either, since they use a mesh standard material.

It used to be the case that DebugTilesRenderer would always render unlit but was changed in https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/195, which I think was good but there are still often cases where using unlit materials is preferred since I'm finding it to be the case that there are a lot of tiles use cases that don't require lighting since photogrammetry tiles do not need to be lit (eg using basic material) or normals. I'm wondering if an "unlit" flag on the DebugTilesRenderer to force all debug materials (including debug mode and original textures) to be forced to use unlit materials:

```js
tiles.registerPlugin( new DebugTilesPlugin( {
  mode: DebugTilesPlugin.ColorModes.RANDOM_COLOR,
  unlit: true,
} );
```

dbuck: Ohno, it was supposed to be a MeshBasicMaterial, apparently let my auto-complete do me dirty.

dbuck: A general unlit option would probably have been nice for my use in the mp app, since we don't have lights in the scene until a user adds some furniture heh. So at the time I made our own debug colorizations that are a little more mp specific ex: https://my.matterport.com/show?m=uskTEa4gEb5&debugTiles=1&tiledmesh=1&dmenu=1&tileStatsOverlay=true&sm=2&sr=-1.12%2C1.49%2C1.11&sp=15.97%2C2.29%2C-8.22&colorize%253AuskTEa4gEb5=byLod

gkjohnson: I've made a PR here into this branch that adds support for the "unlit" flag: https://github.com/matterport/3DTilesRendererJS/pull/6.

Let me know what you think and can both merged if it looks good.

gkjohnson:
Thanks!

Issue #1166, opened 2025-06-05
Google Maps Example wrong height with orthographic camera


mqrty:
### Description

Hello, in the google maps example if you switch to the orthographic camera the height in the hash gets extreme. For example if you go [here](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/googleMapsExample.html#lat=40.7787&lon=-73.9645&height=3461.14&az=-17.85&el=-83.87&roll=18.26) and enable the orthographic camera, the height will jump from about 3461 to 1630483.

### Reproduction steps

1. Go to [google 3d tiles example](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/googleMapsExample.html#lat=40.7787&lon=-73.9645&height=3461.14&az=-17.85&el=-83.87&roll=18.26)
2. Toggle orthographic camera
3. Observe height

### Code

See [googleMapsExample.js](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/v0.4.10/example/googleMapsExample.js)

I guess you cannot just take `camara.position` directly when the camera is orthographic [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/v0.4.10/example/googleMapsExample.js#L214):
```
const localCameraPos = camera.position.clone().applyMatrix4( tilesMatInv );
```

I would be very interested in the correct solution, in my app i work around the same problem by calling `TransitionManager.syncCameras()` and then reading `TransitionManager.perspectiveCamera.position` but i dont know whether thats the right approach.

### Live example

https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/googleMapsExample.html#lat=40.7787&lon=-73.9645&height=3461.14&az=-17.85&el=-83.87&roll=18.26

### Library Version

v0.4.10

### Three.js Version

r170

gkjohnson:
Yes the lat / lon in the demo is designed to be applied to the perspective camera and was never updated to apply to the orthographic camera. The orthographic camera fundamentally works differently from the perspective camera in order to avoid rendering artifacts like near plane clipping - specifically it is pulled back significantly leading to a high height and an apparently inaccurate calculated latitude and longitude.

Calculating the perspective camera transformation to update the hash and seems like a fine solution to me. Would you like to make a PR with the change for the demo?

gkjohnson:
Fixed in #1170

PR #1170, opened 2025-06-06
Google Maps Example fix hash in orthographic mode

mqrty:
Hi, heres the PR for #1166.

Currently this just adjusts `updateHash` to always use the perspective camera.

I'm not sure how to properly adjust `initFromHash`. It would be nice to do it similarly but `syncCameras` would need to sync from perspective to orthographic and I'm unsure how to accomplish that without briefly changing the `transition.mode` or meddling with internals in order to force `fromCamera` to the perspective camera. Maybe we could add an optional parameter to `syncCameras` to explicitly specify the sync direction?

### DIFF
--- a/example/googleMapsExample.js
+++ b/example/googleMapsExample.js
@@ -272,7 +272,7 @@ function initFromHash() {
 	tiles.group.updateMatrixWorld();
 
 	// get the position fields
-	const camera = transition.camera;
+	const camera = transition.perspectiveCamera;
 	const lat = parseFloat( urlParams.get( 'lat' ) );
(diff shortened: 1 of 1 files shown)


gkjohnson:
> I'm not sure how to properly adjust initFromHash.

`CameraTransitionManager.mode` can be assigned directly to immediately switch the camera mode. So something like this should work (untested):

```js
const currentMode = transition.mode;
transition.mode = 'perspective';
// ... adjust the perspective camera parameters

transition.syncCameras();
transition.mode = currentMode;
```

gkjohnson:
Great, thanks!

PR #1171, opened 2025-06-09
ReorientationPlugin: Add missing transformLatlonHeightToOrigin method to type definition

### LINKED ISSUE #1168: ReorientationPlugin.transformLatLonHeightToOrigin method missing from TypeScript definitions
Hi team,

ReorientationPlugin.[transformLatLonHeightToOrigin ](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/5cb9305501509b4b8a4982dd839068cec1d1084d/src/plugins/three/ReorientationPlugin.js#L108)method missing from TypeScript definitions.

The README documentation for ReorientationPlugin also shows a public method transformLatLonHeightToOrigin:
https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/5cb9305501509b4b8a4982dd839068cec1d1084d/src/plugins/README.md?plain=1#L550

However, when inspecting the TypeScript definitions for `ReorientationPlugin` (as generated or included in the package's `.d.ts` files), this method is completely missing.
Here's the [current](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/5cb9305501509b4b8a4982dd839068cec1d1084d/src/plugins/three/ReorientationPlugin.d.ts#L1) TypeScript definition I'm seeing:

```typescript
export class ReorientationPlugin {
	constructor( options?: {
		up?: '+x' | '-x' | '+y' | '-y' | '+z' | '-z',
		recenter?: boolean,
		lat?: number | null,
		lon?: number | null,
		height?: number,
	} );
}
```

**Steps to Reproduce:**

1. Import ReorientationPlugin in a TypeScript project using 3d-tiles-renderer.
2. Create an instance of ReorientationPlugin.
3. Attempt to call myPluginInstance.transformLatLonHeightToOrigin(lat, lon, height).
4. Observe a TypeScript error: "Property 'transformLatLonHeightToOrigin' does not exist on type 'ReorientationPlugin'."

This method is crucial for dynamically updating the tile set's origin without disposing and re-instantiating the plugin (which would be a more heavy-handed approach).

Could you please update the TypeScript definitions to reflect the presence of this method?

Thank you for your time and work on this excellent library!

vkerttula:
Adds the missing `transformLatlonHeightToOrigin` method to the `ReorientationPlugin` TypeScript definitions.

This method is available at runtime but was not declared in the types, which caused a TypeScript error when using it as documented.

Fixes #1168

### DIFF
--- a/src/plugins/three/ReorientationPlugin.d.ts
+++ b/src/plugins/three/ReorientationPlugin.d.ts
@@ -9,4 +9,6 @@ export class ReorientationPlugin {
 		height?: number,
 	} );
 
+	transformLatLonHeightToOrigin( lat: number, lon: number, height?: number ): void;
+
 }


gkjohnson:
Great, thank you!

PR #1174, opened 2025-06-10
Make `enabled` in `DebugTilesPlugin` robust.

pmi-gbs:
The `enabled` property was slightly buggy, if it was disabled it before registering the plugin, it still got enabled.
This should make this more robust, and add more typing definitions as well for the constructor (which now also allows `enabled` as property).

### DIFF
--- a/src/plugins/three/DebugTilesPlugin.js
+++ b/src/plugins/three/DebugTilesPlugin.js
@@ -13,7 +13,7 @@ const _sphere = /* @__PURE__ */ new Sphere();
 const emptyRaycast = () => {};
 const colors = {};
 
-// Return a consistant random color for an index
+// Return a consistent random color for an index
 export function getIndexedRandomColor( index ) {
 
 	if ( ! colors[ index ] ) {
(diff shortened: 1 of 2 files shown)


src/plugins/three/DebugTilesPlugin.js
```diff
@@ -158,19 +159,17 @@ export class DebugTilesPlugin {
...
+			}
 
-				}
+			this._enabled = v;
 
-			} else {
+			if ( this._enabled && this.tiles ) {
 
-				this.dispose();
+				this.init( this.tiles );
```
gkjohnson: Great, thanks! This is a good change.

For the sake of clarity can we call both `init` and `dispose` only when `tiles` does not equal null? Something like so:


```js
if ( v !== this._enabled && this.tiles !== null ) {

	if ( v ) {

		this.init( this.tiles );

	} else {

		this.dispose();

	}

}

this._enabled = v;
```

This should also mean we don't need to check for the presence of "tiles" in the `dispose` function.

pmi-gbs: Yes that also reads nicer.

pmi-gbs:
Whoops I think I missed the last sentence, I removed the check for tiles.

gkjohnson:
Looks great, thanks!

PR #1177, opened 2025-06-11
Set `plugin[ PLUGIN_REGISTERED ] = false` when unregistering a plugin

pmi-gbs:
Not sure if this behavior is intentional, but I think when it's possible to unregister a plugin, it should also be able to be registered again.

For context: I'm using a custom material that doesn't support the fade plugin (or it's at least buggy/throws exceptions, when changing the material on the fly), so I'm temporarily disabling this plugin for this material, until the original material is used again.

### DIFF
--- a/src/base/TilesRendererBase.js
+++ b/src/base/TilesRendererBase.js
@@ -219,6 +219,8 @@ export class TilesRendererBase {
 
 		if ( plugins.includes( plugin ) ) {
 
+			plugin[ PLUGIN_REGISTERED ] = false;
+
 			const index = plugins.indexOf( plugin );
 			plugins.splice( index, 1 );
 			if ( plugin.dispose ) {


gkjohnson:
The current behavior is intentional and changing it potentially has some follow on implications that I don't think I'm ready to fully think through or address at the moment. To lay it out a bit more, plugins are currently designed to be added to a _single_ tile set and not used again once removed. When a plugin is unregistered the "dispose" function is called to clean up any references to the previously-attached TilesRenderer as well as any other objects that need to be released (eg render targets, etc). To that end there's not much value in re-registering an already used plugin, anyway, since it will have to recreate any per-tile or tileset objects, anyway. You might as well use a new one.

If a plugin _could_ be reregistered (assuming the above) the dispose function would have to clean up the plugin to a point where it's basically just been created so "init" can be safely called again, but I'm not sure what the value would be there. 

> I'm using a custom material that doesn't support the fade plugin (or it's at least buggy/throws exceptions, when changing the material on the fly), so I'm temporarily disabling this plugin for this material, until the original material is used again.

If this is the specific use case I think it would be best to adjust fade plugin behavior to be more resilient to encountering materials that have not been "wrapped" or wrap material that it finds that have not been initialized (though now we have to deal with how to correctly dispose the material when that temporary material is added and then removed). Supporting arbitrary materials with the project is powerful but in general supporting small modifications (like fade, etc) to those shaders is unfortunately difficult. Node materials in three.js may make this less complicated in the future.

pmi-gbs:
Thanks for the extensive answer, I expected something like that.

> If this is the specific use case I think it would be best to adjust fade plugin behavior to be more resilient to encountering materials that have not been "wrapped" or wrap material that it finds that have not been initialized

Yep that was the other idea I had, but haven't gone with that yet, as it, as you already say needs a careful design to be cleanup materials correctly etc.

I think for now I'm just creating a new instance of the fade plugin when registering it again (with the original materials), which I think should be fairly safe.

gkjohnson:
It occured to me that if you add and remove the tiles fade plugin multiple tiles the [material "wrap" function](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/plugins/three/fade/wrapFadeMaterial.js#L3) (the function that injects the fade support) will be fire multiple times, which it's not all that resilient to. If you're going to add and remove the plugins this is an area that I think is probably worth looking in to. Happy to answer any questions about the code, of course.

Issue #1161, opened 2025-06-02
Add React 19 Support


druzinov:
I'm currently working on a project that uses React 19.1.0, and I'm seeing an npm error when updating from v0.4.9 to 0.4.10 due to the new peerOptional dependency on react@"^18.3.1".

```
npm ERR! Could not resolve dependency:
npm ERR! peerOptional react@"^18.3.1" from 3d-tiles-renderer@0.4.10
```

I'm assuming the conflict will be the same for react-dom, three, and @react-three/fiber. Would it be possible to update those peer dependency version ranges to support React 19? Thanks!

Environment:
- Node: 22.16.0
- npm: 11.4.1
- react: 19.1.0
- 3d-tiles-renderer: 0.4.10
- three: 0.176.0
- @react-three/fiber: 9.1.2

gkjohnson:
Would you like to make a PR that adjusts the peer dependency to support React 19? I'd prefer not to drop support for React 18, though, so lets set the field such that both versions can be supported.

druzinov:
@gkjohnson It took some time to sort this out with my company, but yes now I am able to make a PR to adjust that. I will make sure support is maintained for both React 18 and React 19. I will put up a PR from a different account sometime next week.

dana-ruzinov:
@gkjohnson Hello! This is that other account, any reason why I wouldn't be able to push up my branch? I can see the repo is public, but I don't think I have push access.

gkjohnson:
You have to create a [fork of the repository](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/about-forks) to push your own changes. You cannot directly push changes to or modify someone else's repository without added permissions.

gkjohnson:
Fixed in #1190

PR #1190, opened 2025-06-27
Support React 19

dana-ruzinov:
Changes:
- Broadened peer dependency versions to include react/react-dom 19 support, where react three fiber needs to be at version 9 for react 19 to work

Relates to ticket #1161

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,98 +1,98 @@
 {
-	"name": "3d-tiles-renderer",
-	"version": "0.4.10",
-	"description": "https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification",
-	"keywords": [
-		"3d-tiles",
-		"3dtiles",
-		"graphics",
-		"b3dm",
-		"gltf",
-		"threejs",
-		"rendering",
-		"geometry",
-		"terrain",
-		"cesium"
-	],
-	"type": "module",
-	"sideEffects": false,
-	"main": "src/index.js",
-	"module": "src/index.js",
-	"files": [
-		"src/*"
-	],
-	"exports": {
-		".": "./src/index.js",
-		"./r3f": "./src/r3f/index.jsx",
-		"./plugins": "./src/plugins/index.js",
-		"./src/*": "./src/*"
-	},
-	"scripts": {
-		"start": "vite --config ./vite.config.js",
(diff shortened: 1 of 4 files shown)


dana-ruzinov:
@gkjohnson Quick question about testing the package build/unit tests against different peer dependency versions. To do that I could test it in a CI/CD pipeline with job matrices... but I'm really unfamiliar with GitHub pipelines. Any thoughts on how to go about this?

package.json
```diff
@@ -74,15 +74,14 @@
     "jest-cli": "^27.1.1",
     "leva": "^0.9.35",
     "postprocessing": "^6.36.4",
-    "three": "^0.170.0",
```
gkjohnson: why was three.js removed from dev dependencies? This means it will not install on clean installation, no?

gkjohnson: I suppose the other react dependences aren't included here, either 🤔

dana-ruzinov: You're including it as a peer dependency already, I figure if the other peer dependencies don't need a matching dev dependency three shouldn't either.

gkjohnson: From [this stackoverflow answer](https://stackoverflow.com/a/35207983) it looks like peer dependencies are now automatically installed - but the dev dependency version was a newer version and I'd prefer not to change the versioning used for dev in this PR.

If you're interested we could also add support for testing multiple three.js versions to ensure the project works on all relevant releases. This can be done like so (see the same approach in [three-mesh-bvh](https://github.com/gkjohnson/three-mesh-bvh/blob/master/.github/workflows/node.js.yml#L18-L20)):

```yml
matrix:
    three-version: [0.159.0, 0.168.0, latest]
# ...
- run: npm ci
- run: npm install three@${{ matrix.three-version }}
- run: npm test
```

This forces the versions in the matrix to be installed before testing. I leave it up to you whether you'd like to add this - but lets keep the three.js dev dependency for now.

gkjohnson:
I've made some of the remaining requested changes - it seems that npm is running into some odd infinite loops when trying to install the older and latest versions of three.js possibly related to some of the dev dependency conflicts so I've skipped the version matrix for now.

Thanks!

dana-ruzinov:
Thanks!

PR #1202, opened 2025-07-01
TilesAttributionOverlay: Enable the use outside of HTTPS contexts

shotamatsuda:
This PR enables the use of `TilesAttributionOverlay` outside of HTTPS contexts.

**Issue to solve:**

The class ID for `TilesAttributionOverlay` is generated using `crypto.randomUUID`, which is only available in HTTPS contexts. This prevents it from being used in local development over HTTP except on loopback addresses. Since cryptographically secure randomness of class IDs is not essential, it's reasonable to provide a fallback to allow its use outside of HTTPS contexts.

### DIFF
--- a/src/r3f/components/TilesAttributionOverlay.jsx
+++ b/src/r3f/components/TilesAttributionOverlay.jsx
@@ -2,6 +2,22 @@ import { useContext, useState, useEffect, useMemo } from 'react';
 import { TilesRendererContext } from './TilesRenderer.jsx';
 import { CanvasDOMOverlay } from './CanvasDOMOverlay.jsx';
 
+function randomUUID() {
+
+	if ( window.crypto.randomUUID ) {
+
+		return window.crypto.randomUUID();
+
+	}
+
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thanks! I didn't know that. Interesting that `getRandomValues` is available in all contexts, though. Do you think there's a reason to no just always use the `getRandomValues` code path?

shotamatsuda:
I didn't know that either. I think `crypto.randomUUID` is just fine, and there is no reason to avoid it if available, except for its rather long footprint. Although, if a UUID format is not required, I'm inclined to do:

```
crypto.getRandomValues(new Uint32Array(1))[0].toString(16)
```

and its entropy is sufficient for the purpose. Please let me know any edits I should make.

gkjohnson:
> ```js
> crypto.getRandomValues(new Uint32Array(1))[0].toString(16)
> ```

I think this is definitely good enough.

gkjohnson:
Great, thank you!

PR #1213, opened 2025-07-07
Correctly remove event listener on UpdateOnChangePlugin dispose

WilliamLiu-1997:
From line 38-40 of UpdateOnChangePlugin.js, three event listener `needs-update`, `add-camera`, `delete-camera` are added.
However, in the `dispose` function, they are not correctly removed
Instead, `load-content`, `camera-add`, `camera-delete` are removed but they are never added


gkjohnson:
Great, thank you!

PR #1248, opened 2025-07-29
Remove unused __childrenWereVisible property from TileInternal and TilesRendererBase

shawn0326:
The __childrenWereVisible property was no longer being used and was cluttering the tile initialization code. Removing deprecated properties helps keep the codebase clean and reduces confusion for developers working with the tile system.


gkjohnson:
Great! Thank you -

Issue #1187, opened 2025-06-26
Suggestion: Separate 3D Tiles Core Logic into a Standalone Base npm Package and Manage with Monorepo


shawn0326:
Hi, I’d like to share a suggestion and would love to hear your thoughts!

### Why consider this change?
- **Decoupling:** By separating the 3D Tiles scheduling core from the three.js implementation, the base logic could be reused by other rendering engines, not just three.js.
- **Modularity & Maintainability:** A more modular structure would make the codebase easier to maintain and understand, with core logic and renderer-specific logic clearly separated.
- **Broader Adoption:** Publishing the base as a standalone package may encourage more developers to use and contribute, including those building on top of custom renderers.

### Current challenges
- The current implementation mixes core scheduling logic with three.js-specific code, which makes it harder to reuse or maintain.
- The plugins directory contains both base and three-related logic, which can be confusing.
- The core scheduling logic directly relies on three.js math classes, making decoupling more difficult.

### Possible approach
1. Split the codebase into `base`, `three`, and `r3f` (if needed) directories. For example, plugins currently include both base and three subfolders—separating them could make the structure clearer.
2. Manage the project as a monorepo, so that each package (`base`, `three`, `r3f`) is independently importable. This way, developers could use only the base package for their custom renderer integration if needed.
3. If feasible, make the base package rely on its own math classes, fully removing the dependency on three.js.

I believe this separation could bring long-term benefits to the project and its community~

gkjohnson:
Hey! Thanks for the suggestion and the explanation. I'm not sure if I see a lot of value in migrating to multiple packages or a monorepo but I do agree that some reorganization is in order. I think most of the issues you raised can be addressed separately. Though generally the big bottleneck is the amount of overhead that some of changes may bring to the project including maintenance, documentation, etc.

> Decoupling: By separating the 3D Tiles scheduling core from the three.js implementation, the base logic could be reused by other rendering engines, not just three.js.

A lot of the core 3d tiles logic is already separated but I agree that more could be. Most of the reason it's not separated further is due to dependencies on three.js math functions for calculations, as you've mentioned.

> Broader Adoption: Publishing the base as a standalone package may encourage more developers to use and contribute, including those building on top of custom renderers.

[...]

PR #1249, opened 2025-07-29
Move decodeOctNormal method to threejs directory

shawn0326:
As discussed in https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1187#issuecomment-3112794427, I have moved the `decodeOctNormal` method from the `core` to the `three` folder. I am currently unsure whether relocating it to the `loaders` folder would be more appropriate, as it is only utilized by `PNTSLoader` at the moment.


gkjohnson:
Thanks! Before merging this can you comment on https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1187#issuecomment-3115284140? The more I've been working on things recently and thinking about it the more I think it's best to transition the core to use just the three.js math functionality. There are a number of places where I feel like I'm bending over backwards to try to avoid using three functions or moving functionality to the three.js folder because some basic functionality is needed and generally I think it will be best for the stability and consistency of everyone using the library if as much code can be shared as possible while appreciating that subsequent users will not want to be importing the whole of three.js rendering code.

gkjohnson:
Thanks! Lets merge this for now - though it may make it's way back into core with the recent discussions about the math functions.

Issue #1116, opened 2025-05-06
Best way to prevent tiles from loading in Region


xawill:
### Question

I am trying to use the new `LoadRegionPlugin` to fulfill my use case of loading the Google 3D Tiles everywhere except in Switzerland (described in #813).

I created a custom `Region` for Switzerland and it works great to show the tiles inside the region at a desired precision (like in the plugin example). However, I cannot make it work to <ins>**hide**</ins> the tiles inside the region (while keeping usual error computation outside the region).

I thought about "inverting" the region to be "Not Switzerland", but in that case the whole World is loading at the specified errorTarget which is not tied to screen space error. I tried to understand how you compute it in the TilesRendererBase to implement my way but found it non trivial and therefore wanted to ask if this is the correct way to do it before doing it. 

Could you confirm this is currently not supported by the plugin or did I missed something?

My use case is similar to #457, except that at the time `LoadRegionPlugin` didn't exist. The goal is to avoid having to use clipping planes.

It might also be linked to #1062. By the way, what is the status of the proposed PR to include a `intersection` or `exclusive` parameter to the plugin?

Thank you for your time and the great library :)

### Library Version

0.4.8

### Three.js Version

r175

gkjohnson:
> Could you confirm this is currently not supported by the plugin or did I missed something?

Right the `LoadRegionPlugin` is designed to trigger tile loads that intersect the given shape up to the provided `errorTarget`. There's no designed support for keeping the camera screen space error in the given region or "clipping" the region out of the tiles to load. As you mention, the changes described in #1062 are likely most of the way to the kind of functionality that you're looking for. Though as noted in that thread "inverting" the shape will not work so simply and we'd need to implement custom intersection logic for that behavior:

> Note that this won't "just work" if you want to invert the shape - eg cut out a box shaped hole from the larger tile set since in that case you'd need to check for complete containment of each tile rather than just intersection.

---

> what is the status of the proposed PR to include a intersection or exclusive parameter to the plugin?

I'm not sure what @jsulli's plans are here but they're the person to ask.

jsulli:
I don't have immediate plans to make a new plugin, I have a local override in place that plugs in the changes from https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/1062 at runtime which fits my use-case for now. I do still plan to take a stab at this in the future because I'd like my approach to be available to everyone using the library, but it's not at the top of my list.

[...]

PR #1261, opened 2025-08-09
Allow LoadRegion to be masked

xawill:
Following discussion in #1116, I am finally opening a PR with my proposal.

It is tested and working for my own use case (explained in the issue); let me know if it fits in the direction you want for the project :)

### DIFF
--- a/src/three/plugins/LoadRegionPlugin.js
+++ b/src/three/plugins/LoadRegionPlugin.js
@@ -6,6 +6,7 @@ export class LoadRegionPlugin {
 	constructor() {
 
 		this.name = 'LOAD_REGION_PLUGIN';
+		this.priority = - 1000;
 		this.regions = [];
 		this.tiles = null;
 
@@ -55,9 +56,16 @@ export class LoadRegionPlugin {
 		const boundingVolume = tile.cached.boundingVolume;
 		const { regions, tiles } = this;
 
+		let mask = false;
 		let visible = false;
 		let maxError = - Infinity;
 		for ( const region of regions ) {
+
+			if ( region.mask ) {
+
+				mask = true;
+
+			}
 
 			const intersects = region.intersectsTile( boundingVolume, tile, tiles );
 			if ( intersects ) {
@@ -72,6 +80,14 @@ export class LoadRegionPlugin {
 		target.inView = visible;
 		target.error = maxError;
 
+		return ( ( ! visible && mask ) ? false : null ); // Returns false if should force mask the tile.
+
+	}
+
+	raycastTile( tile, scene, raycaster, intersects ) {
+
+		return ! tile.__inFrustum;
+
 	}
 
 	dispose() {
@@ -85,9 +101,10 @@ export class LoadRegionPlugin {
 // Definitions of predefined regions
 export class BaseRegion {
 
-	constructor( errorTarget = 10 ) {
+	constructor( errorTarget = 10, mask = false ) {
 
 		this.errorTarget = errorTarget;
+		this.mask = mask;
 
 	}
 
@@ -103,9 +120,9 @@ export class BaseRegion {
 
 export class SphereRegion extends BaseRegion {
 
-	constructor( errorTarget = 10, sphere = new Sphere() ) {
+	constructor( errorTarget = 10, mask = false, sphere = new Sphere() ) {
 
-		super( errorTarget );
+		super( errorTarget, mask );
 		this.sphere = sphere.clone();
 
 	}
@@ -120,9 +137,9 @@ export class SphereRegion extends BaseRegion {
 
 export class RayRegion extends BaseRegion {
 
-	constructor( errorTarget = 10, ray = new Ray() ) {
+	constructor( errorTarget = 10, mask = false, ray = new Ray() ) {
 
-		super( errorTarget );
+		super( errorTarget, mask );
 		this.ray = ray.clone();
 
 	}
@@ -137,9 +154,9 @@ export class RayRegion extends BaseRegion {
 
 export class OBBRegion extends BaseRegion {
 
-	constructor( errorTarget = 10, obb = new OBB() ) {
+	constructor( errorTarget = 10, mask = false, obb = new OBB() ) {
 
-		super( errorTarget );
+		super( errorTarget, mask );
 		this.obb = obb.clone();
 		this.obb.update();
 
--- a/src/core/renderer/tiles/traverseFunctions.js
+++ b/src/core/renderer/tiles/traverseFunctions.js
@@ -47,8 +47,8 @@ function resetFrameState( tile, renderer ) {
 		tile.__allChildrenLoaded = false;
 
 		// update tile frustum and error state
-		renderer.calculateTileViewError( tile, viewErrorTarget );
-		tile.__inFrustum = viewErrorTarget.inView;
+		const shouldForceShowTile = renderer.calculateTileViewError( tile, viewErrorTarget );
+		tile.__inFrustum = shouldForceShowTile === false ? false : viewErrorTarget.inView;
 		tile.__error = viewErrorTarget.error;
 		tile.__distanceFromCamera = viewErrorTarget.distanceFromCamera;
 
--- a/src/three/plugins/LoadRegionPlugin.d.ts
+++ b/src/three/plugins/LoadRegionPlugin.d.ts
@@ -6,8 +6,9 @@ import { Sphere, Ray } from 'three';
 
 declare class BaseRegion {
 
-	constructor( errorTarget?: number );
+	constructor( errorTarget?: number, mask?: boolean );
 	errorTarget: number;
+	mask: boolean;
 	intersectsTile( boundingVolume: TileBoundingVolume, tile: Tile, tilesRenderer: TilesRenderer ): boolean
 	calculateError( tile: Tile, tilesRenderer: TilesRenderer ): number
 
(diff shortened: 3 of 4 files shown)


src/three/plugins/LoadRegionPlugin.js
```diff
@@ -55,10 +56,17 @@ export class LoadRegionPlugin {
...
 		let visible = false;
 		let maxError = - Infinity;
 		for ( const region of regions ) {
 
+			if ( region.mask ) {
+
+				mask = true;
+
+			}
+
```
gkjohnson: I'm not sure about this logic. It looks like if any region is specified as a "mask" then any region is used as such, which doesn't seem right.

src/three/plugins/LoadRegionPlugin.js
```diff
@@ -72,6 +80,14 @@ export class LoadRegionPlugin {
...
 		target.error = maxError;
 
+		return ( ( ! visible && mask ) ? false : null ); // Returns false if should force mask the tile.
+
+	}
+
+	raycastTile( tile, scene, raycaster, intersects ) {
+
+		return ! tile.__inFrustum;
+
```
gkjohnson: Why is this needed? This function is intended for overriding raycasting which shouldn't be needed in this plugin.

src/three/plugins/LoadRegionPlugin.js
```diff
@@ -72,6 +80,14 @@ export class LoadRegionPlugin {
 		target.inView = visible;
 		target.error = maxError;
 
+		return ( ( ! visible && mask ) ? false : null ); // Returns false if should force mask the tile.
```
gkjohnson: Instead of returning a boolean we should assign "inView" on the target object to align with the existing API:

- "true" means the tile is in view and should be traversed.
- "false" means the tile is _not_ in and should not be traversed (even if it's in the camera view).
- "null" means not in view of the regions but do not override the camera traversal.

src/three/plugins/LoadRegionPlugin.js
```diff
@@ -6,6 +6,7 @@ export class LoadRegionPlugin {
 	constructor() {
 
 		this.name = 'LOAD_REGION_PLUGIN';
+		this.priority = - 1000;
```
gkjohnson: Why did the priority change?

gkjohnson (commented):
Thanks! Left a few comments - I think this is a good use case to support. cc @jsulli for feedback, as well.

src/three/plugins/LoadRegionPlugin.js
xawill: I noticed that without overriding, the tiles, even masked, would still be hit by raycasting (during GlobeControls events for instance). The pivot point should not be on hidden tiles, IMO. My particular use case is that two tilesets are layered at the same location (Google 3d tiles and Switzerland-specific dataset). If I want to hide google tiles over Switzerland, I don't want google 3d tiles to be hit by raycasting (it looks like they are higher altitude, probably due to geoid/ellipsoid mismatch).

src/three/plugins/LoadRegionPlugin.js
xawill: See comment above regarding raycasting. BatchedTilesPlugin also overrides `raycastTile`, so I needed LoadRegionPlugin's `raycastTile` to take precedence. Actual value was arbitrary and based on other plugins. It just needs to be lower than any other plugin overriding `raycastTile`.

src/three/plugins/LoadRegionPlugin.js
xawill: Just to be sure we have the same understanding for `mask`, `mask=true` for a specific region means that only the tiles intersecting that region will be displayed.
I see your concerns now that I am thinking about it. Basically the question is: in case two regions overlap and they have conflicting `mask` flag, which region should take precedence? Let me think a bit about it and propose a solution.

edit: previous comment content was completely wrong after further thinking, now I understand what you mean.

src/three/plugins/LoadRegionPlugin.js
xawill: Ok, sorry. I was not sure about the API here and whether you wanted it to be updated, based on your previous comments encouraging us to create a PR about it. Getting my head around it and fixing it ASAP.

xawill: Could you please clarify if the three cases you described above are current or future API? Because for me current API doesn't support `inView=null` anywhere (so I guess this is future API). But, based on my experiments before these changes, a tile with `inView=false` can still be rendered. So it would mean introducing a breaking change where `inView=false` no indicates that the tile must not be rendered? Or did I misunderstood your explanation? 
In any case, I am currently experiencing difficulties implementing it "your way". A comparison of the API before/after these changes would help. Thanks a lot!

src/three/plugins/LoadRegionPlugin.js
gkjohnson: The TilesRenderer has a concept of "active tiles" which are tiles that are loaded and represent data that is technically present but not rendered. This allows for raycasting to surfaces offscreen which GlobeControls and EnvironmentControls use, for example, to prevent the the camera from sinking through the tile set surface. Limiting raycasting to only tiles in the frustum will break this behavior.

I understand what you're interested in but I think for now we'll have to leave this up to end users to manage (eg filter hits that are within these mask volumes) and wait to see more about the uses cases these feature is used for. I'll try to think through it a bit more, as well.

src/three/plugins/LoadRegionPlugin.js
gkjohnson: Here's my expectation for this feature:

- By default a "region" shape will allow for loading tiles that intersect it up to a target screenspace error - useful for loading tiles in a target region to a certain level of detail or visualization, raycasting, sampling, export, etc.
- If a shape is marked as "mask" then only tiles that will be loaded due to camera frustums or other regions that also intersect a "mask" shape should be displayed.

I don't want to deal with complex boolean operations ordering so I think it's more than sufficient to consider all masked regions to be the same. This means that visibility from non-mask regions should be evaluate and then filtered down based on the mask values (or vice versa).

src/three/plugins/LoadRegionPlugin.js
gkjohnson: > Because for me current API doesn't support inView=null anywhere (so I guess this is future API)

That's right this isn't used anywhere at the moment. This would mean adding a "null" option rather than a return value from this function.

> But, based on my experiments before these changes, a tile with inView=false can still be rendered. So it would mean introducing a breaking change where inView=false no indicates that the tile must not be rendered?

Yeah that's true - admittedly `null` and `false` is a bit ambiguous, as well. I suppose ideally we might allow for exporting enums like `IN_REGION`, `OUT_OF_REGION`, and `MASKED_OUT` where the last option means "don't render this tile even if it's in the camera frustum". My explanation above would map to "true", "null", and "false". "null" seems to map to "OUT_OF_REGION" in my mind because it basically means "don't do anything" and just follow the camera frustums.

Regarding breaking changes - some of these plugin callbacks are still changing quite a bit as we find what the useful hooks are for the renderer and they're not documented so I generally feel okay with a breaking change to the return values from the "calculateTileViewError" here. But I'll ping @jsulli and @WilliamLiu-1997 since they're the two I know have been interested in this plugin specifically for some more thoughts.

xawill: Okey, all makes sense now. Thanks a lot for the confirmation.

I have now updated the API as proposed. I just assumed the enum would be an improvement and not yet for this PR. However, I find it very error prone to have a boolean with a meaningful `null` value. I therefore added a few warnings to make sure bugs are not introduced in the future by forgetting about this. The enum would indeed be cleaner and make everything more explicit. Let me know if you want that in this PR. But first please just check the existing logic :)

src/three/plugins/LoadRegionPlugin.js
xawill: Ok, I reverted. Any chance you can elaborate on how end users (me, in that case) could handle hit filtering, particularly with GlobeControls?

I don't think current public API allows anything like this without a custom GlobeControls, right? As a first step, I tried by using `globeControls.raycaster.layers.set(targetTilesetLayerId)`, but from what I can see `GlobeControls.raycaster` is private API (at least not exposed in types definitions), right?

gkjohnson:
Sorry for the delays on this - I've just taken a deeper look at everything and made some changes. Let me know what you think:

- The plugin function now returns "false" if the in view value should be considered a no op, probably similar to where the PR started (sorry for the back and forth).
- Mask values are now correctly applied _after_ all other shape "in view" calculations (`inView = null || inView` was otherwise always a no-op line).
- Updated the demo to add a flag for testing "mask" behavior. If "mask" is true and "regionOnly" is false then tiles will load up to the camera's error target while tiles outside the masked region will not be loaded.

xawill:
Thank you for the feedback.

I am trying to test it myself but struggle with the `master` branch merges. I personally prefer rebasing, but whatever

It looks like d5d7a7ac8d5c18a36ccfbb50c913108c572d4267 broke something with Typescript. I am getting a hard time to have my project find the correct imports and compile. I couldn't find the root cause yet. Any idea what could mess up so much the imports in this commit? It was fine before, after my fixes in #1289.

gkjohnson:
Sorry if my merges caused issues - I'll try to use rebase in the future.

> It looks like d5d7a7a broke something with Typescript. I am getting a hard time to have my project find the correct imports and compile. I couldn't find the root cause yet. Any idea what could mess up so much the imports in this commit?

#1275 brought quite a few changes to simplify the internal imports between subpackages. The [package.json definition](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/package.json#L25-L58) _seems_ well-formed but I'm not a typescript expert (and have more than once been tripped up by how package.json module resolution works).

Can you tell me more about what's going wrong? Is there an error? Is this only happening in local package development? Or when importing the package in another project? Is this happening in master? It may be worth making a dedicated issue to discuss.

xawill:
Ok so I have been able to test on my side and LGTM! 

Thank you for your improvements; definitely cleaner.

-> Ready to be merged 🎉

gkjohnson:
Great, thanks for taking a look!

PR #1263, opened 2025-08-10
Fix Google Maps overlay

xawill:
A refactoring of the plugin introduced a small (but breaking) issue I fix here.

Moreover, I improved the constructor typings, as according to [the docs](https://developers.google.com/maps/documentation/tile/session_tokens?_gl=1*3o9lcg*_up*MQ..*_ga*ODU3NzEzMzIzLjE3NTQ4MzQ1MTQ.*_ga_NRWSTWS78N*czE3NTQ4MzQ1MTQkbzEkZzAkdDE3NTQ4MzQ1MTQkajYwJGwwJGgw), the `sessionOptions` params you listed are required. Other optional params can be listed.

### DIFF
--- a/src/three/plugins/images/ImageOverlayPlugin.d.ts
+++ b/src/three/plugins/images/ImageOverlayPlugin.d.ts
@@ -85,7 +85,7 @@ export class GoogleMapsOverlay extends ImageOverlay {
 		apiToken: string,
 		autoRefreshToken?: boolean,
 		logoUrl?: string,
-		sessionOptions?: null | { mapType: string, language: string, region: string },
+		sessionOptions: { mapType: string, language: string, region: string } & { [key: string]: any },
 
 		color: number | Color,
 		opacity: number,
--- a/src/three/plugins/images/ImageOverlayPlugin.js
+++ b/src/three/plugins/images/ImageOverlayPlugin.js
@@ -1615,7 +1615,7 @@ export class GoogleMapsOverlay extends ImageOverlay {
 			.then( json => {
 
 				this.imageSource.tileDimension = json.tileWidth;
-				this.imageSource = 'https://tile.googleapis.com/v1/2dtiles/{z}/{x}/{y}';
+				this.imageSource.url = 'https://tile.googleapis.com/v1/2dtiles/{z}/{x}/{y}';
 				return this.imageSource.init();
 
 			} );


src/three/plugins/images/ImageOverlayPlugin.d.ts
```diff
@@ -85,7 +85,7 @@ export class GoogleMapsOverlay extends ImageOverlay {
 		apiToken: string,
 		autoRefreshToken?: boolean,
 		logoUrl?: string,
-		sessionOptions?: null | { mapType: string, language: string, region: string },
+		sessionOptions: { mapType: string, language: string, region: string } & { [key: string]: any },
```
gkjohnson: We can remove "null" since it's needed for the loading images from Google Cloud but why add `{ [key: string]: any }`? As far as I understand the "mapType" etc structure is required.

xawill: Yeah, sorry, my description was too concise. This part is to allow any other fields, because the API allows more options like for instance `styles` which gives freedom to customize the layout (visibility, color, etc.) of map features like road, country names, lakes, etc... Without it we cannot add more fields in an object literal than the three required fields already listed. Of course we could replace the `any` type with a much more detailed type definition, but honestly I don't think anyone would want to maintain such a data structure.

So my proposal is to make sure than the user fills at least the required fields and give them the possibility to add more if they want.

gkjohnson: Would something like this work to allow for extra fields? This how it's handled elsewhere in the project.

```js
{
  mapType: string,
  language: string,
  region: string,
  [key: string]?: string,
}
```

Or explicitly allowing the other fields allowed by the API as optional?

```js
{
  mapType: string,
  language: string,
  region: string,
  style?: string,
}
```

xawill: No. Based on Google's docs linked above, `styles` doesn't take just a string, but an object which could contain a lot of fields (all optional). Provided example in Google's docs is:

```
{
  "mapType": "satellite",
  "language": "en-US",
  "region": "us",
  "layerTypes": [ "layerRoadmap", "layerStreetview" ],
  "overlay":  true,
  "scale": "scaleFactor1x",
  "styles": [
    {
      "stylers": [
        { "hue": "#00ffe6" },
        { "saturation": -20 }
      ]
    },{
      "featureType": "road",
      "elementType": "geometry",
      "stylers": [
        { "lightness": 100 },
        { "visibility": "simplified" }
      ]
    }
  ]
}
```

So IMO it's either the `any` or going through the whole API, exhaustively typing it and maintaining it over time in case Google changes the API.

gkjohnson: Got it, thanks - lets change to this, then, and change it [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/plugins/GoogleCloudAuthPlugin.d.ts#L8), as well:

```js
{
  mapType: string,
  language: string,
  region: string,
  [key: string]?: any,
}
```

src/three/plugins/GoogleCloudAuthPlugin.d.ts
```diff
@@ -5,7 +5,7 @@ export class GoogleCloudAuthPlugin {
 		autoRefreshToken?: boolean,
 		logoUrl?: string,
 		useRecommendedSettings?: boolean;
-		sessionOptions?: null | { mapType: string, language: string, region: string },
+		sessionOptions: { mapType: string, language: string, region: string, [key: string]: any },
```
gkjohnson: To be clear: `null` in still needed as an option in this case since "null" indicates that the 3d tiles api should be used whereas the 2d image api requires these session options. I'll have to update the docs to include info on the session options field.

xawill: Got it. Didn't realize this AuthPlugin was used for both 3d and 2d tiles.

gkjohnson:
Looks great, thanks!

Issue #1266, opened 2025-08-11
New build step for exports breaks TS


xawill:
### Description

Following your last changes regarding the build step for exports (in c545fdc8), the type definitions are not properly exported anymore. I am first opening an issue to make sure that we align on how to fix it because IMO there are two possibilities:
1) Just include `import` and `types` in each module's `export` entry in package.json with the existing src path. Example below.
2) Also export the type definitions in the build folder and use this path for the `types` path (see 1 above). This I didn't immediately manage to solve, because I encounter strange tsc issues and wanted to check with you first.

Bottomline, I think you are the best to know how you want to have your library built, so I suggest you fix it yourself, it will be easier. But if you prefer, I can try to make it work and submit a PR.

### Reproduction steps

1. Build library with `npm run build-lib`.
2. Notice no type definitions are exported.
3. Notice any TS project using the library won't compile because the type definitions are not found.

### Code

```
"exports": {
".": {
  "import": "./build/index.js",
  "types": "./src/index.d.ts"
},
// Include other plugins; similar setup
}
```

### Live example

_I think this is trivial enough to not need that, but let me know if you disagree._

### Library Version

ce39630eca68e79d5d90ac6bff8122fefaed00b2

### Three.js Version

r179.1

gkjohnson:
Thanks for the catch - I'm not a typescript expert so I'm open to recommendations but just adding the "types" into the exports field seems like the most simple setup if you'd like to make a PR.

Jut for context I've added a build so that internal files can use the bare-imports (eg `import { ... }from '3d-tiles-renderer/core'`), which will make internal dependencies easier to manage, without breaking things for users who may be importing via cdn and also make the cdn imports faster to load.

There may be other ways around it, like replacing any imports before publish, but this seems like the most reasonable option for now.

PR #1271, opened 2025-08-12
Include type declarations in build step

xawill:
Proposal fix for #1266, as discussed

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "@typescript-eslint/parser": "^7.14.1",
     "@vitejs/plugin-react": "^4.3.2",
     "babel-jest": "^27.5.1",
-    "cesium": "^1.131.0",
+    "cesium": "^1.132.0",
     "concurrently": "^6.2.1",
     "eslint": "^8.57.1",
     "eslint-config-mdcs": "^5.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -23,14 +23,38 @@
...
-    "./plugins": "./build/index.plugins.js",
-    "./core": "./build/index.core.js",
-    "./three": "./build/index.three.js",
-    "./r3f": "./build/index.r3f.js",
-    "./core/plugins": "./build/index.core-plugins.js",
-    "./three/plugins": "./build/index.three-plugins.js",
-    "./src/*": "./src/*"
+    ".": {
+		"import": "./build/index.js",
+		"types": "./build/types/index.d.ts"
```
gkjohnson: Amazing, thanks! Can you explain why all the d.ts files need to be copied to the `/build` folder rather than be directly referenced in the `/src` folder?

xawill: They don't. Both ways work; just a "design" choice. It was my initial proposal in the issue to keep consistency with your build folder approach to have there all the files exported by the library. This includes the types IMO. But you could also reference the files in the src folder and remove the copying step in the vite config file. Your choice :)

gkjohnson: Thanks for the clarification! I think I'd like to err on the side of less code and keep the builds more simple where possible so I've just pushed the changes to reference the types directly in the /src folder. If this looks good to you then we can merge it!

xawill: A bit more difficult to say with the last merge commit, but yes, LGTM. Thanks!

PR #1287, opened 2025-08-20
Add support for NodeMaterial in TilesFadePlugin

shotamatsuda:
This PR adds preliminary support for `NodeMaterial` in `TilesFadePlugin`. I consider this preliminary because it doesn't support `BatchedMesh`, as I have not been able to use `BatchedMeshPlugin` with `WebGPURenderer` yet.

A possible incompatibility is that it will not work if the user is using importmaps and doesn't have a route to `three/tsl`.

An example page was added for demonstration purposes, but it's mostly a duplicate. It would be ideal to add the ability to switch the renderer in `fadingTiles.html`. I'll be happy to remove it, or to work on it if this PR is legitimate.

### DIFF
--- a/example/webgpu/fadingTiles.html
+++ b/example/webgpu/fadingTiles.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
+		<meta charset="utf-8"/>
+
+		<title>Dither Fade Tiles</title>
+		<link rel="stylesheet" href="./styles.css">
+    </head>
+    <body class="light">
+		<div id="info">
+			Demonstration of tiles using a dither fade to change, smoothing out the transition.
+		</div>
+        <script src="./fadingTiles.js" type="module"></script>
+    </body>
+</html>
--- a/example/webgpu/fadingTiles.html
+++ b/example/webgpu/fadingTiles.html
@@ -5,7 +5,7 @@
 		<meta charset="utf-8"/>
 
 		<title>Dither Fade Tiles</title>
-		<link rel="stylesheet" href="./styles.css">
+		<link rel="stylesheet" href="../styles.css">
     </head>
     <body class="light">
 		<div id="info">
--- a/example/webgpu/fadingTiles.js
+++ b/example/webgpu/fadingTiles.js
@@ -1,0 +1,243 @@
+import {
+	Scene,
+	PerspectiveCamera,
+	OrthographicCamera,
+	Group,
+} from 'three';
+import { MeshBasicNodeMaterial, WebGPURenderer } from 'three/webgpu';
+import { TilesFadePlugin } from '3d-tiles-renderer/plugins';
+import { EnvironmentControls, TilesRenderer, CameraTransitionManager } from '3d-tiles-renderer';
+import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
+
+let controls, scene, renderer;
+let groundTiles, skyTiles, tilesParent, transition;
+
+const params = {
+
+	reinstantiateTiles,
+	fadeRootTiles: false,
+	useFade: true,
+	errorTarget: 6,
+	fadeDuration: 0.5,
+	renderScale: 1,
+	fadingGroundTiles: '0 tiles',
+
+	orthographic: false,
+	transitionDuration: 0.25,
+
+};
+
+init().then( () => {
+
+	render();
+
+} );
+
+async function init() {
+
+	// renderer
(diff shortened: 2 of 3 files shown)


gkjohnson:
Thanks @shotamatsuda! And sorry for the delay in reply.

I suppose some (all?) of my concerns from https://github.com/mrdoob/three.js/issues/30185 are finally coming to pass, so I'm trying to decide what to do here 😅 It's also exposing a few other issues I wasn't expecting, like the [custom ring shader for EnvironmentControls anchor](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/8f5cfb95d5f364e2dac34a7f11a40e84c0e04c0e/src/three/renderer/controls/PivotPointMesh.js#L34-L105) not rendering and logging an error with WebGPURenderer. Generally I want to support WebGPU more directly but unfortunately with the way three.js is handling things that will seem to mean maintaining two copies of every shader in the project since most if not all of my other users and clients are on WebGLRenderer and complicate the APIs (no one else has complained about WebGPU support yet).

cc @mrdoob are there any more concrete plans for how to more gracefully handle the WebGL -> WebGPU material transition? This is exactly the kind of issue I anticipated running in to and trying to avoid. These types of issues are going to make maintaining the project much more difficult and much less enjoyable. To lay it all out:

- All custom materials must be written and maintained twice in order to support WebGL & WebGPU.
- Code must contain an internal switch to detect node materials but even then it's not fool-proof since users can assign non-node materials and have them work gracefully in WebGPURenderer (the strategy used in this PR is not resilient to this case). So the only robust option is to require users to pass a "webGPU: true" option into everything that needs to know the target rendering system. The solution in this PR is to require the user to [manually manage and replace all loaded materials](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/1287/files#diff-911cf4670354b3f1b515ede10ab8285366b4769aaf66c9e411a3116bdf61baabR120-R158) because there's otherwise no way to gracefully support both renderers.
- TSL code will be imported into webgl projects that do not need it and vice-versa for webgpu projects importing shader lib code, unnecessarily inflating project sizes.
- The three.js project is encouraging users to replace the `"three"` import with the `"three/webgpu"` variant which means any internal dependency that expects "three" to export ShaderMaterial etc will break if users follow the guidance implied by the official three.js examples (see [this fiddle from three.js bug reports](https://jsfiddle.net/xno7bmw0/), and the [official webgpu example imports](https://github.com/mrdoob/three.js/blob/dev/examples/webgpu_backdrop.html#L18)).

shotamatsuda:
Thanks for sparing your time on this PR!

I'm perfectly fine with you and Three.js taking time to make the best decision, especially since you haven't had any requests for WebGPU support from your clients. Support for `NodeMaterial` in `TilesFadePlugin` [can be implemented in user code](https://github.com/takram-design-engineering/three-geospatial/tree/webgpu/storybook-webgpu/src/plugins/fade), thanks to the plugin architecture. As to the pivot mesh, which I had not noticed, I agree that having two copies of shaders is redundant and makes it harder to maintain, and it makes sense to wait for Three.js to support `NodeMaterial` in `WebGLRenderer`.

gkjohnson:
@shotamatsuda - I'm starting to take a look at what's involved in migrating to TSL for some of the plugins. One concern I have is that modifying even a single node on a built-in material, even if it's modified in exactly the same was as others, will cause a rebuild of the TSL graph even if it's structured exactly the same as another (eg adding the same dither fade to all materials). Have you seen any performance issues with this? It's the kind of thing that will happen every time a new tile gets loaded.

shotamatsuda:
@gkjohnson I'm not sure we have the same problem, but if I construct nodes like `fadeIn` and `fadeOut` per material in this PR, it affects performance when new tiles are loaded, to the extent I'd like to avoid it.

The expensive part is `material.build`, which is triggered when a part of its graph changes, as I outlined here: https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1573#issuecomment-4464308289. Even if the changing node is a leaf in the graph, the entire graph will be rebuilt. Similarities of subgraphs don't make it faster. I haven't been following TSL updates for a while since r185, so there might be some improvements now.

gkjohnson:
> if I construct nodes like fadeIn and fadeOut per material in this PR, it affects performance when new tiles are loaded, to the extent I'd like to avoid it.
> ...
> The expensive part is material.build, which is triggered when a part of its graph changes

Yeah this was my concern... I made an issue here (https://github.com/mrdoob/three.js/issues/33821) that discusses it a bit - it may be worth making a separate issue for it, though. You can't share shader graph builds between any materials even if you're just changing a uniform. It's a pretty big design flaw in TSL, in my opinion. This is probably the primary thing keeping me from migrating this project at the moment...

PR #1289, opened 2025-08-21
Fix loaders' working path creation

### LINKED ISSUE #1282: LoaderBase: Improve "resolveExternalURL" resolution
When loading [this](https://3d.geo.admin.ch/ch.swisstopo.vegetation.3d/v1/20250728/10/205/328.i3dm) (technically a CMPT) the i3dm loader throws an error that it cannot load the constructed external urls - possibly because of embedded path navigation (`..`) or double slashes (`//`).

xawill:
Resolves #1282.

Trailing slash was not always added to `workingPath` and `resolveExternalURL` was not accounting for relative path (such as "../../tree.glb" in the tile mentioned in #1282).


gkjohnson:
Great, thank you! I moved a few things around but this looks good.

Issue #122, opened 2020-10-04
Example: Add BatchTable demonstration to I3DM example


gkjohnson:

xawill:
I can help if this is something you still want. If yes, please let me know if you want a new example or extend an existing one.

As a new example, I can propose Swiss quantized mesh terrain overlayed with Swiss aerial imagery, with i3dm vegetation layer with color change on hover (as initially asked) and scale of individual tree read from batch table.


--
Edit: We should probably wait for #1281 to be fixed before adding this example.

gkjohnson:
> I can help if this is something you still want. If yes, please let me know if you want a new example or extend an existing one.

That would be great, if you'd like to! My preference would be to adjust the `i3dmExample` to load a single i3dm file with batch table and display hover information with hover effect as is done in the [b3dm example](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/b3dmExample.html) to demonstrate how accessing the batched table info works in the instance case. I don't want to add a full 3d tile set example for this demonstration - just a single piece of geometry like in the b3dm case so we can probably just grab and display a single vegetation tile from the swiss model.

PR #1290, opened 2025-08-21
Add BatchTable demonstration to I3DM example

xawill:
Following discussion in #122.

I mostly reused your work with the B3DM example.

I found an I3DM tile in the swisstopo vegetation tileset (since most are actually CMPT) and used it for the example. It adds a bit more info in the batch table than the Cesium one used up to now. 

However the trees orientation is all messed up. Not exactly sure why because I have no issue when loading the whole tileset, but I imagine this is because of a transformation at the tileset level. I don't think we can do anything about it in the example.

**⚠️ Is blocked by #1289. Merge #1289 first.** (Thus the two commits, including 4b220ec which is common to the two stacked PRs. I am not sure how you prefer to handle stacked PRs...)


gkjohnson:
Excellent, thank you again! I made some changes to make the b3dm example more consistent and made BatchTable easier to work with by pre-filtering the "extensions" field in the BatchTable header, as well.

PR #1293, opened 2025-08-25
Feat: Refactor getUrlExtension for performance improvement

junhaotong:
### Summary
This pull request refactors the getUrlExtension function to improve its performance and robustness.

### Motivation
The previous version of getUrlExtension used multiple regular expressions to parse the URL. While functional, this approach can be slow due to the overhead of string manipulation and regex engine execution. 

The new implementation avoids regular expressions entirely. It uses highly optimized native string methods like lastIndexOf and substring to directly find the necessary parts of the URL. This method is significantly faster.

### Performance Benchmark
A comprehensive benchmark was run to compare the "New" refactored function against the "Old" one. The benchmark measured operations per second (Ops/sec), where a higher number indicates better performance. The results below demonstrate the significant speed improvements across various URL types.

| URL                                                                         | New Ops/sec  | Old Ops/sec  | Winner | Performance |
| --------------------------------------------------------------------------- | ------------ | ------------ | ------ | ----------- |
| foo.json                                                                    | 13854932.75  | 4027995.31   | New    | 3.44x       |
| /foo/bar.json                                                               | 12364101.01  | 3216382.82   | New    | 3.84x       |
| foo.json?a=b                                                                | 12217549.29  | 3396461.00   | New    | 3.60x       |
| /foo.json                                                                   | 12747627.76  | 3601058.44   | New    | 3.54x       |
| /foo/bar.json                                                               | 12054030.50  | 3001826.58   | New    | 4.02x       |
| png.svg.json                                                                | 12906596.14  | 2399542.87   | New    | 5.38x       |
| https://nasa.gov                                                            | 15483397.31  | 10160478.34  | New    | 1.52x       |
| https://nasa.gov/                                                           | 15738192.77  | 8207638.54   | New    | 1.92x       |
| https://nasa.gov/tileset                                                    | 9818308.52   | 3927720.85   | New    | 2.50x       |
| https://nasa.gov/foo/bar.baz/tileset?foo=bar                                | 9483266.56   | 3588111.63   | New    | 2.64x       |
| https://nasa.gov/foo/bar.baz/tileset?a.b=c.d                                | 9722803.77   | 3679619.44   | New    | 2.64x       |
| https://nasa.gov/tileset                                                    | 9863827.37   | 4059355.31   | New    | 2.43x       |
| https://nasa.gov/tileset.                                                   | 9962048.81   | 3561356.87   | New    | 2.80x       |
| Pleiades                                                                    | 13592597.05  | 3373336.89   | New    | 4.03x       |
| .                                                                           | 16875075.12  | 10451281.57  | New    | 1.61x       |
| ..                                                                          | 16863101.17  | 9575891.81   | New    | 1.76x       |
|                                                                             | 172493493.45 | 176531397.27 | Old    | 0.98x       |
| https://nasa.gov/foo/bar.baz/tileset.json                                   | 8662649.43   | 2262693.95   | New    | 3.83x       |
| https://nasa.gov/foo/bar.baz/tileset.json?foo=bar                           | 9165839.66   | 2202666.31   | New    | 4.16x       |
| https://nasa.gov/foo/bar.baz/tileset.json?a.b=c.d                           | 9147051.05   | 2181744.78   | New    | 4.19x       |
| https://nasa.gov/foo/bar.baz/tileset.json?a.b=query/path/to/file            | 8352799.16   | 2150514.86   | New    | 3.88x       |
| https://nasa.gov/foo/bar.baz/tileset.json?a.b=c.d#e.f                       | 11792347.76  | 2153265.37   | New    | 5.48x       |
| https://nasa.gov/tileset.json                                               | 9971810.54   | 2319395.40   | New    | 4.30x       |
| https://nasa.gov//tileset.json                                              | 9882156.90   | 2315658.22   | New    | 4.27x       |
| file:///Users/JaneScientist/code/3DTilesRendererJS/example/b3dmExample.json | 7809895.08   | 537171.78    | New    | 14.54x      |
| undefined                                                                   | 106769184.51 | 109174899.95 | Old    | 0.98x       |
---

### DIFF
--- a/src/core/renderer/utilities/urlExtension.js
+++ b/src/core/renderer/utilities/urlExtension.js
@@ -13,59 +13,31 @@ export function getUrlExtension( url ) {
 
 	// Find the last occurrence of '?' and '#' to handle query params and fragments
 	let endIndex = url.length;
-	const queryIndex = url.lastIndexOf( '?' );
-	const fragmentIndex = url.lastIndexOf( '#' );
-
+	const queryIndex = url.indexOf( '?' );
+	const fragmentIndex = url.indexOf( '#' );
 	if ( queryIndex !== - 1 ) {
 
 		endIndex = Math.min( endIndex, queryIndex );
 
 	}
+
 	if ( fragmentIndex !== - 1 ) {
 
 		endIndex = Math.min( endIndex, fragmentIndex );
 
 	}
 
-	// If this looks like a URL without a path (just domain), return null
-	// Check if we're dealing with a domain-only URL by looking for protocol
-	if ( url.includes( '://' ) ) {
-
-		const urlAfterProtocol = url.substring( url.indexOf( '://' ) + 3, endIndex );
-		const slashIndex = urlAfterProtocol.indexOf( '/' );
-
-		// If there's no slash after the domain, or the slash is at the very end,
-		// then we're dealing with just a domain name, not a file
-		if ( slashIndex === - 1 || slashIndex === urlAfterProtocol.length - 1 ) {
-
-			return null;
-
-		}
-
-	}
-
-	// Find the last slash to get the filename part
+	// Check if the string is just a hostname or whether the path does not end in an extension
+	const lastPeriodIndex = url.lastIndexOf( '.', endIndex );
 	const lastSlashIndex = url.lastIndexOf( '/', endIndex );
-	const filename = url.substring( lastSlashIndex + 1, endIndex );
-
-	// If there's no filename (empty string or just a slash), return null
-	if ( ! filename ) {
-
-		return null;
-
-	}
-
-	// Find the last dot in the filename
-	const lastDotIndex = filename.lastIndexOf( '.' );
-
-	if ( lastDotIndex === - 1 || lastDotIndex === 0 ) {
+	const protocolIndex = url.indexOf( '://' );
+	const isHostOnly = protocolIndex !== - 1 && protocolIndex + 2 === lastSlashIndex;
+	if ( isHostOnly || lastPeriodIndex === - 1 || lastPeriodIndex < lastSlashIndex ) {
 
 		return null;
 
 	}
 
-	const extension = filename.substring( lastDotIndex + 1 );
-
-	return extension || null;
+	return url.substring( lastPeriodIndex + 1, endIndex ) || null;
 
 }
--- a/test/urlExtension.test.js
+++ b/test/urlExtension.test.js
@@ -4,6 +4,8 @@ describe( 'getUrlExtension', () => {
 
 	it.each( [
 		'https://nasa.gov/foo/bar.baz/tileset.json',
+		'https://nasa.gov/foo/bar.baz/tileset.json#a.b',
+		'https://nasa.gov/foo/bar.baz/tileset.json#/a.b',
 		'https://nasa.gov/foo/bar.baz/tileset.json?foo=bar',
 		'https://nasa.gov/foo/bar.baz/tileset.json?a.b=c.d',
 		'https://nasa.gov/foo/bar.baz/tileset.json?a.b=query/path/to/file',
@@ -26,6 +28,9 @@ describe( 'getUrlExtension', () => {
 	it.each( [
 		'https://nasa.gov',
 		'https://nasa.gov/',
+		'https://nasa.gov#a.b',
+		'https://nasa.gov#/a.b',
+		'https://nasa.gov/foo/bar.baz/tileset.json/#/a.b',
 		'https://nasa.gov/tileset',
 		'https://nasa.gov/foo/bar.baz/tileset?foo=bar',
 		'https://nasa.gov/foo/bar.baz/tileset?a.b=c.d',


junhaotong:
The lint errors are not introduced by this PR. They were already present due to a previous commit ([028c307](https://github.com/NASA-AMMOS/3DTilesRendererJS/commit/028c3079666ba95faedf29c3754bd5f51abcba6c)
), which is why npm run lint fails here.

gkjohnson:
Hello! Thank you for the contribution - can you explain a bit more about what the motivation for this change was? Is it somehow a bottleneck? I understand the function can be faster but it otherwise doesn't look like it's so slow that it should be causing issuee. I ask because the new implementation is quite a bit more complex than the previous one.

> which is why npm run lint fails here.

Yes the linting in master is broken while some changes for React 19 are made. This should be fixed once #1286 is merged.

junhaotong:
Thank you for the review and for raising this point.

The motivation behind this change is performance. While the previous regex-based implementation is concise and correct, regular expressions introduce a non-trivial amount of parsing overhead, especially when this function is executed repeatedly or in performance-sensitive code paths. By switching to native string operations, I avoid regex backtracking and significantly reduce execution time.

I agree that the new implementation is more verbose and complex, but the trade-off was made in favor of measurable runtime efficiency. To help offset the added complexity, I can add inline comments to clarify the logic if that would be helpful.

gkjohnson:
> The motivation behind this change is performance. 

Thank you - I understand this point but there is almost _always_ a faster way to implement something so at some point you have to prioritize code readability and maintainability. The percentage improvements look high here but in practice the amount of actual evaluation time is incredibly small. So I'm wondering if you've seen any real world performance issues from the `getURLExtension` function? Are you seeing stutters, or framerate issues etc that this PR fixes?

junhaotong:
Thank you, that’s a very fair point. 
I haven’t observed visible stutters or frame drops that were directly attributable to getUrlExtension, so this change isn’t fixing a user-facing performance bug.

My motivation came mainly from noticing that this helper is called very frequently in the tile-loading code path of our GIS application. When users pan or zoom quickly — especially with multiple tile layers loaded — thousands of tile requests can be triggered in a short time. In such scenarios, even small per-call savings can add up and help reduce the risk of frame drops.

I fully agree that readability and maintainability should take priority, so I’m happy to add inline comments to make the implementation clearer, or adjust the approach — or even close the PR if we feel the regex version is the better trade-off for now.

gkjohnson:
I think generally I'd expect these kinds of changes to come with some kind of practical impact so the value of the change can be evaluated against real world usage- eg "This was taking 5ms in the profiler when loading a tile set before and now it's taking 1ms". But I also understand the desire to optimize functions that are being called a lot. It's also the case that the use of regex objects may create new memory but I'm not completely sure.

Here's my proposed change based on the new function in this PR that's a bit shorter and more readable (to me, at least). If this looks good then if you update the PR with this function implementation we can get it merged:

```js
export function getUrlExtension( url ) {

	if ( ! url ) {

		return null;

	}

	// Find the last occurrence of '?' and '#' to handle query params and fragments
	let endIndex = url.length;
	const queryIndex = url.indexOf( '?' );
	const fragmentIndex = url.indexOf( '#' );
	if ( queryIndex !== - 1 ) {

		endIndex = Math.min( endIndex, queryIndex );

	}

	if ( fragmentIndex !== - 1 ) {

		endIndex = Math.min( endIndex, fragmentIndex );

	}

	// Check if the string is just a hostname or whether the path does not end in an extension
	const lastPeriodIndex = url.lastIndexOf( '.', endIndex );
	const lastSlashIndex = url.lastIndexOf( '/', endIndex );
	const protocolIndex = url.indexOf( '://' );
	const isHostOnly = protocolIndex !== - 1 && protocolIndex + 2 === lastSlashIndex;
	if ( isHostOnly || lastPeriodIndex === - 1 || lastPeriodIndex < lastSlashIndex ) {

		return null;

	}

	return url.substring( lastPeriodIndex + 1, endIndex ) || null;

}
```

Let me know what you think - note I haven't benchmarked it. While we're here if you could also add a few more lines to the tests fur this function to ensure that the extension extraction will work with a hash (using "#") that would be great.

junhaotong:
It looks very good, more shorter and readable, I'll copy this function and run the benchmark and get back to you

junhaotong:
Thanks for the suggestion! I agree that your version is cleaner and likely faster, especially since this function can be called quite often.

I did notice, though, this updated version the function doesn’t correctly handle URLs like:

**https://nasa.gov/foo/bar.baz/tileset.json/#/a.b**


In that case the trailing slash before the fragment causes the extension check to fail and I get null instead of "json". I’ll add a small guard to handle this case and also extend the tests to cover # fragments so this doesn’t regress in the future.

gkjohnson:
> I did notice, though, this updated version the function doesn’t correctly handle URLs like:
> `https://nasa.gov/foo/bar.baz/tileset.json/#/a.b`
> ...
>  I get null instead of "json"

I wouldn't consider this a valid extension / file url. Adding a trailing slash after a file name causes the static servers I just tested to fail. Returning "null" should be the correct behavior.

junhaotong:
Your version is perfect then — I’m happy to go with it as-is.

| URL                                                                         | New Ops/sec  | Old Ops/sec  | Winner | Performance |
| --------------------------------------------------------------------------- | ------------ | ------------ | ------ | ----------- |
| foo.json                                                                    | 19126559.68  | 3915221.35   | New    | 4.89x       |
| /foo/bar.json                                                               | 18778259.10  | 3034644.50   | New    | 6.19x       |
| foo.json?a=b                                                                | 17525606.21  | 3249789.77   | New    | 5.39x       |
| /foo.json                                                                   | 18099466.98  | 3381336.31   | New    | 5.35x       |
| /foo/bar.json                                                               | 18395716.89  | 3206002.75   | New    | 5.74x       |
| png.svg.json                                                                | 18049430.74  | 2339448.98   | New    | 7.72x       |
| https://nasa.gov                                                            | 21916461.40  | 9889374.20   | New    | 2.22x       |
| https://nasa.gov/                                                           | 22939448.93  | 8161788.82   | New    | 2.81x       |
| https://nasa.gov/tileset                                                    | 20885871.01  | 3842703.31   | New    | 5.44x       |
| https://nasa.gov/foo/bar.baz/tileset?foo=bar                                | 19220958.54  | 3473750.53   | New    | 5.53x       |
| https://nasa.gov/foo/bar.baz/tileset?a.b=c.d                                | 19101236.49  | 3490096.85   | New    | 5.47x       |
| https://nasa.gov/tileset                                                    | 20923627.68  | 3831253.27   | New    | 5.46x       |
| https://nasa.gov/tileset.                                                   | 20964273.96  | 3444650.10   | New    | 6.09x       |
| Pleiades                                                                    | 21761460.96  | 3246256.68   | New    | 6.70x       |
| .                                                                           | 24321013.79  | 10371644.95  | New    | 2.34x       |
| ..                                                                          | 24053957.30  | 9382191.90   | New    | 2.56x       |
|                                                                             | 173688927.39 | 169568916.97 | New    | 1.02x       |
| https://nasa.gov/foo/bar.baz/tileset.json                                   | 19925673.72  | 2147959.61   | New    | 9.28x       |
| https://nasa.gov/foo/bar.baz/tileset.json?foo=bar                           | 19266300.30  | 2085900.01   | New    | 9.24x       |
| https://nasa.gov/foo/bar.baz/tileset.json?a.b=c.d                           | 19365102.69  | 2086724.69   | New    | 9.28x       |
| https://nasa.gov/foo/bar.baz/tileset.json?a.b=query/path/to/file            | 19356146.67  | 2058074.26   | New    | 9.40x       |
| https://nasa.gov/foo/bar.baz/tileset.json?a.b=c.d#e.f                       | 27193006.57  | 2083594.79   | New    | 13.05x      |
| https://nasa.gov/tileset.json                                               | 27658938.32  | 2205097.71   | New    | 12.54x      |
| https://nasa.gov//tileset.json                                              | 27702798.65  | 2197845.83   | New    | 12.60x      |
| file:///Users/JaneScientist/code/3DTilesRendererJS/example/b3dmExample.json | 27030955.10  | 512161.81    | New    | 52.78x      |
| undefined                                                                   | 101084280.27 | 102529885.99 | Old    | 0.99x       |
| https://nasa.gov/foo/bar.baz/tileset.json#a.b   | 19401145.58 | 1455567.56  | New    | 13.33x      |
| https://nasa.gov/foo/bar.baz/tileset.json#/a.b  | 19549559.32 | 5338004.08  | New    | 3.66x       |
| https://nasa.gov/foo/bar.baz/tileset.json/#/a.b | 20880406.42 | 5341577.94  | New    | 3.91x       |
---

gkjohnson:
Great, thanks!

PR #1302, opened 2025-09-02
Added WMS format, with relative capabilities loader, Image source, Ti…

SoftwareMechanic:
…lesPlugin, OverlayPlugin and 2 examples.

Note: seems that wms requests are not triggered with planar rendering, only WMS which support EPSG:3857 or EPSG:4326 are supposed to work for now.

### DIFF
--- a/example/wmsOverlay.js
+++ b/example/wmsOverlay.js
@@ -1,0 +1,309 @@
+import { Scene, WebGLRenderer, PerspectiveCamera } from "three";
+import {
+	TilesRenderer,
+	GlobeControls,
+	EnvironmentControls,
+} from "3d-tiles-renderer";
+import { GUI } from "three/examples/jsm/libs/lil-gui.module.min.js";
+import {
+	CesiumIonAuthPlugin,
+	GLTFExtensionsPlugin,
+	ImageOverlayPlugin,
+	WMSTilesOverlay,
+	WMTSCapabilitiesLoader,
+    WMSCapabilitiesLoader
+} from "3d-tiles-renderer/plugins";
+import { XYZTilesPlugin } from "3d-tiles-renderer/plugins";
+import * as THREE from "three";
+
+import { DRACOLoader } from "three/examples/jsm/loaders/DRACOLoader.js";
+
+const dracoLoader = new DRACOLoader().setDecoderPath(
+	"https://www.gstatic.com/draco/v1/decoders/",
+);
+
+let capabilities, controls, scene, renderer, tiles, camera, gui;
+let wmsOverlay, overlayPlugin;
+let params;
+
+(async () => {
+	await init();
+	render();
+})();
+
+async function init() {
+	renderer = new WebGLRenderer({ antialias: true });
+	renderer.setPixelRatio(window.devicePixelRatio);
+	renderer.setSize(window.innerWidth, window.innerHeight);
+	renderer.setClearColor(0x111111);
+	renderer.setAnimationLoop(render);
+
+	document.body.appendChild(renderer.domElement);
+
+	scene = new Scene();
+	camera = new PerspectiveCamera(
+		60,
+		window.innerWidth / window.innerHeight,
+		0.001,
+		10000,
+	);
+
+	params = {
+		planar: false,
+		wmsOpacity: 0.7,
+	};
+
+	console.log("initializing");
+	await updateCapabilities(); // Wait for capabilities to load
+
+	onWindowResize();
+	window.addEventListener("resize", onWindowResize, false);
+}
+function rebuildGUI() {
+	if (gui) gui.destroy();
+
+	const layer = capabilities.layers.find((l) => l.name === params.layer);
+
+	gui = new GUI();
+	// gui.add(params, "planar").onChange(rebuildTiles); // Disabled: WMS does not work in planar mode
+	// NOTE: Planar mode is disabled because WMS overlays do not render correctly in planar mode.
+
+	gui.add(params, "wmsOpacity", 0, 1, 0.01).onChange(updateOverlayParams);
+	gui
+		.add(
+			params,
+			"layer",
+			capabilities.layers.map((l) => l.name),
+		)
+		.onChange(updateOverlayParams);
+	gui
+		.add(
+			params,
+			"styles",
+			layer.styles.map((s) => s.name),
+		)
+		.onChange(updateOverlayParams);
+	gui.add(params, "crs", layer.crs).onChange(updateOverlayParams);
+	gui
+		.add(params, "format", ["image/png", "image/jpeg"])
+		.onChange(updateOverlayParams);
+	gui.add(params, "tileDimension", [256, 512]).onChange(updateOverlayParams);
+	gui.add(params, "version", ["1.1.1", "1.3.0"]).onChange(updateOverlayParams);
+}
+
+function rebuildTiles() {
+	if (tiles) tiles.dispose();
+	if (controls) {
+		controls.dispose();
+		controls = null;
+	}
+
+	tiles = new TilesRenderer();
+	tiles.setCamera(camera);
+	scene.add(tiles.group);
+
+	// Base map plugin (XYZ)
+	tiles.registerPlugin(
+		new XYZTilesPlugin({
+			bounds: [-180, -90, 180, 90],
+			levels: 18,
+			center: true,
+			shape: "ellipsoid", // seems that planar is no working with wms
+			url: "https://tile.openstreetmap.org/{z}/{x}/{y}.png",
+		}),
+	);
+
+	// tiles.registerPlugin(
+	// 	new GLTFExtensionsPlugin({
+	// 		dracoLoader: dracoLoader,
+	// 	}),
+	// );
+
+	// google tiles for testing with wms
+	// tiles.registerPlugin(
+	// 	new CesiumIonAuthPlugin({
+	// 		apiToken: `YOUR_CESIUM_TOKEN_HERE`,
+	// 		assetId: "2275207",
+	// 		autoRefreshToken: true,
+	// 	}),
+	// );
+
+	// Example: Add Authorization header for WMS (Basic Auth)
+	// const username = "your_username";
+	// const password = "your_password";
+	// const basicAuth = "Basic " + btoa(`${username}:${password}`);
+
+	const overlayBounds = [-180, -90, 180, 90];
+
+	const crsParam = params.version === "1.1.1" ? "SRS" : "CRS";
+
+	wmsOverlay = new WMSTilesOverlay({
+		baseUrl: "https://geoservizi.regione.liguria.it/geoserver/M2660/wms",
+		layer: params.layer,
+		crs: params.crs,
+		crsParam,
+		format: params.format,
+		tileDimension: params.tileDimension,
+		//styles: params.styles,
+		version: params.version,
+		bounds: overlayBounds,
+		levels: 10,
+		opacity: params.wmsOpacity,
+		// extraHeaders: {
+		// 	//"Authorization": "Bearer your_token_here"
+		// 	Authorization: "Basic " + btoa("your_username:your_password"),
+		// },
+	});
+
+	overlayPlugin = new ImageOverlayPlugin({
+		overlays: [wmsOverlay],
+		renderer,
+		resolution: 256,
+	});
+	tiles.registerPlugin(overlayPlugin);
+
+	// Controls setup
+	if (params.planar) {
+		controls = new EnvironmentControls(scene, camera, renderer.domElement);
+		controls.enableDamping = true;
+		controls.minDistance = 1e-4;
+		controls.maxDistance = 5;
+		controls.cameraRadius = 0;
+		controls.fallbackPlane.normal.set(0, 0, 1);
+		controls.up.set(0, 0, 1);
+		controls.camera.position.set(0, 0, 2);
+		controls.camera.quaternion.identity();
+
+		camera.near = 1e-4;
+		camera.far = 10;
+		camera.updateProjectionMatrix();
+	} else {
+		tiles.group.rotation.x = -Math.PI / 2;
+		controls = new GlobeControls(scene, camera, renderer.domElement);
+		controls.setEllipsoid(tiles.ellipsoid, tiles.group);
+		controls.enableDamping = true;
+		moveCameraToLonLat(8.9463, 44.4056);
+		controls.minDistance = 150;
+	}
+}
+
+function lonLatHeightToCartesian(lon, lat, height = 0) {
+    const radius = 6378137.0 + height;
+    const lambda = (lon * Math.PI) / 180;
+    const phi = (lat * Math.PI) / 180;
+    const x = radius * Math.cos(phi) * Math.cos(lambda);
+    const y = radius * Math.cos(phi) * Math.sin(lambda);
+    const z = radius * Math.sin(phi);
+    return { x, y, z };
+}
+
+function moveCameraToLonLat(lon, lat, height = 0, cameraDistance = 1e7) {
+    // 1. Get the marker position
+    const { x, y, z } = lonLatHeightToCartesian(lon, lat, height);
+
+    // 2. Apply the same rotation as the globe group
+    const rotationMatrix = new THREE.Matrix4().makeRotationX(-Math.PI / 2);
+    const rotatedPos = new THREE.Vector3(x, y, z).applyMatrix4(rotationMatrix);
+
+    // 3. Place camera along the same direction, at a distance
+    const direction = rotatedPos.clone().normalize();
+    camera.position.copy(direction.multiplyScalar(cameraDistance));
+    
+    // 4. Focus camera on the rotated marker position
+    camera.lookAt(rotatedPos);
+}
+
+async function updateCapabilities() {
+	const url =
+		"https://geoservizi.regione.liguria.it/geoserver/M2660/wms?REQUEST=GetCapabilities";
+
+	const loader = new WMSCapabilitiesLoader()
+
+
+	capabilities = await loader.load(url);
+
+	console.log("capabilities:", capabilities);
+
+	const defaultLayer = capabilities.layers[0];
+	let selectedCRS = "EPSG:4326";
+	if (defaultLayer.crs.includes("EPSG:3857")) {
+		selectedCRS = "EPSG:3857";
+	}
+
+	params = {
+		// planar: false, // Disabled: WMS does not work in planar mode
+		wmsOpacity: 0.7,
+		optimizeWMS: false,
+		layer: defaultLayer.name,
+		style: defaultLayer.styles[0]?.name || "",
+		crs: selectedCRS,
+		format: "image/png",
+		tileDimension: 256,
+		bounds: defaultLayer.boundingBoxes[0]?.bounds || [-180, -90, 180, 90],
+		version: capabilities.version || "1.3.0",
+		styles: defaultLayer.styles[0]?.name || "",
+	};
+
+	rebuildGUI();
+	rebuildTiles();
+}
+
+function updateOverlayParams() {
+	// Remove old overlay and plugin
+	if (overlayPlugin) {
+		tiles.unregisterPlugin(overlayPlugin);
+		overlayPlugin = null;
+	}
+	if (wmsOverlay) {
+		wmsOverlay.dispose?.();
+		wmsOverlay = null;
+	}
+
+	// Recreate overlay with updated params
+	const overlayBounds = [-180, -90, 180, 90];
+	const crsParam = params.version === "1.1.1" ? "SRS" : "CRS";
+
+	wmsOverlay = new WMSTilesOverlay({
+		baseUrl: "https://geoservizi.regione.liguria.it/geoserver/M2660/wms",
+		layer: params.layer,
+		crs: params.crs,
+		crsParam,
+		format: params.format,
+		tileDimension: params.tileDimension,
+		version: params.version,
+		bounds: overlayBounds,
+		levels: 10,
+		opacity: params.wmsOpacity,
+		// styles: params.styles,
+		// extraHeaders: { ... }
+	});
+
+	overlayPlugin = new ImageOverlayPlugin({
+		overlays: [wmsOverlay],
+		renderer,
+		resolution: 256,
+	});
+	tiles.registerPlugin(overlayPlugin);
+
+	tiles.update();
+}
+
+function onWindowResize() {
+	const aspect = window.innerWidth / window.innerHeight;
+	camera.aspect = aspect;
+	camera.updateProjectionMatrix();
+	renderer.setSize(window.innerWidth, window.innerHeight);
+}
+
+function render() {
+	if (controls) {
+		controls.update();
+		camera.updateMatrixWorld();
+	}
+	if (tiles) {
+		tiles.setCamera(camera);
+		tiles.setResolutionFromRenderer(camera, renderer);
+		tiles.update();
+	}
+	renderer.render(scene, camera);
+}
--- a/example/wmsTiles.js
+++ b/example/wmsTiles.js
@@ -1,0 +1,274 @@
+import { Scene, WebGLRenderer, PerspectiveCamera } from "three";
+import {
+	TilesRenderer,
+	GlobeControls,
+	EnvironmentControls,
+} from "3d-tiles-renderer";
+import {
+	CesiumIonAuthPlugin,
+	TilesFadePlugin,
+	UpdateOnChangePlugin,
+	WMSCapabilitiesLoader,
+	WMSTilesPlugin,
+} from "3d-tiles-renderer/plugins";
+import { GUI } from "three/examples/jsm/libs/lil-gui.module.min.js";
+import { XYZTilesPlugin } from "3d-tiles-renderer/plugins";
+import * as THREE from "three";
+
+const url =
+	window.location.hash.replace(/^#/, "") ||
+	"https://geoservizi.regione.liguria.it/geoserver/M2660/wms?request=GetCapabilities";
+
+let controls, scene, renderer;
+let xyzTiles, wmsTiles, camera, gui;
+let params, capabilities;
+
+init();
+render();
+
+function init() {
+	renderer = new WebGLRenderer({ antialias: true });
+	renderer.setPixelRatio(window.devicePixelRatio);
+	renderer.setSize(window.innerWidth, window.innerHeight);
+	renderer.setClearColor(0x111111);
+	renderer.setAnimationLoop(render);
+
+	document.body.appendChild(renderer.domElement);
+
+	scene = new Scene();
+	camera = new PerspectiveCamera(
+		60,
+		window.innerWidth / window.innerHeight,
+		0.001,
+		10000,
+	);
+
+	updateCapabilities();
+
+	onWindowResize();
+	window.addEventListener("resize", onWindowResize, false);
+	window.addEventListener("hashchange", () => location.reload());
+}
+
+async function updateCapabilities() {
+    capabilities = await new WMSCapabilitiesLoader().loadAsync(url);
+
+    const defaultLayer = capabilities.layers[0];
+    let selectedCRS = "EPSG:4326";
+    if (defaultLayer.crs.includes("EPSG:3857")) {
+        selectedCRS = "EPSG:3857";
+    }
+
+    params = {
+        baseUrl: url, // Remove GetCapabilities params
+        layer: defaultLayer.name,
+        style: defaultLayer.styles[0]?.name || "",
+        crs: selectedCRS,
+        format: "image/png",
+        tileDimension: 256,
+        planar: false,
+        bounds: defaultLayer.boundingBoxes[0]?.bounds || [-180, -90, 180, 90],
+        version: capabilities.version || "1.3.0",
+        styles: defaultLayer.styles[0]?.name || "",
+    };
+
+    rebuildGUI();
+    rebuildTiles();
+}
+
+function rebuildGUI() {
+	if (gui) gui.destroy();
+
+	const layer = capabilities.layers.find((l) => l.name === params.layer);
+
+	gui = new GUI();
+	//gui.add(params, "planar").onChange(rebuildTiles); wms doesn't show up in planar mode
+	gui
+		.add(
+			params,
+			"layer",
+			capabilities.layers.map((l) => l.name),
+		)
+		.onChange(() => {
+			const selectedLayer = capabilities.layers.find(
+				(l) => l.name === params.layer,
+			);
+			params.crs = selectedLayer.crs[0] || "EPSG:3857";
+			params.bounds = selectedLayer.boundingBoxes[0]?.bounds || [
+				-180, -90, 180, 90,
+			];
+			params.styles = selectedLayer.styles[0]?.name || "";
+			rebuildGUI();
+			rebuildTiles();
+		});
+	gui
+		.add(
+			params,
+			"styles",
+			layer.styles.map((s) => s.name),
+		)
+		.onChange(rebuildTiles);
+	gui.add(params, "crs", layer.crs).onChange(rebuildTiles);
+	gui.add(params, "format", ["image/png", "image/jpeg"]).onChange(rebuildTiles);
+	gui.add(params, "tileDimension", [256, 512]).onChange(rebuildTiles);
+}
+function rebuildTiles() {
+	if (xyzTiles) xyzTiles.dispose();
+	if (wmsTiles) wmsTiles.dispose();
+	if (controls) {
+		controls.dispose();
+		controls = null;
+	}
+
+	// XYZ base layer
+	xyzTiles = new TilesRenderer();
+	xyzTiles.registerPlugin(new TilesFadePlugin());
+	xyzTiles.registerPlugin(new UpdateOnChangePlugin());
+	xyzTiles.registerPlugin(
+		new XYZTilesPlugin({
+			center: true,
+			shape: params.planar ? "planar" : "ellipsoid",
+			url: "https://tile.openstreetmap.org/{z}/{x}/{y}.png",
+		}),
+	);
+	xyzTiles.setCamera(camera);
+	scene.add(xyzTiles.group);
+
+	// WMS overlay layer
+	wmsTiles = new TilesRenderer();
+	wmsTiles.registerPlugin(new TilesFadePlugin());
+	wmsTiles.registerPlugin(new UpdateOnChangePlugin());
+	wmsTiles.registerPlugin(
+		new WMSTilesPlugin({
+			shape: params.planar ? "planar" : "ellipsoid",
+			center: true,
+			baseUrl: url.replace(/\?.*$/, ""),
+			layer: params.layer,
+			crs: params.crs,
+			format: params.format,
+			tileDimension: params.tileDimension,
+			styles: params.styles,
+			version: params.version,
+			bounds: params.bounds,
+		}),
+	);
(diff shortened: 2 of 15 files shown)


gkjohnson:
Thank you for the work! These are some really great additions. I was going to ask you to make an issue to discuss the GeoJSON addition but what you've made is pretty close to what I was imagining!

A couple comments on the PR:
- I've pulled your branch and run the `wmsTiles` example but I'm not seeing in WMS tiles and I'm getting errors in the console. I only see XYZ tiles displaying - is there a reason to have XYZ tile are included? I'd expect the example to focus on just displaying the WMS tile format. Once it's working I can give some more detailed feedback.
- The GeoJSON addition is great but would you mind if we separate it into a separate PR? Then we can discuss that addition separately and keep this PR focused on WMS - I think it will make everything easier for follow for me.

SoftwareMechanic:
You are welcome, I look forward to implement even more formats.

you are right, that error is due to that specific wms I think, because with other wms services it works without errors, and using the same even in cesium viewer, gives the same errors (but still works in the end, probably its because that is a heavy WMS, and that could be the reason why you don't see it, it takes some time to load also). I will change the WMS used for the examples 👍 

for what concerns having separate pull requests, yes you are right, give me just some time and I will fix everything

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1,14 +1,29 @@
...
+	WebGLRenderTarget,
+	Color,
+	SRGBColorSpace,
+	BufferAttribute,
+	Matrix4,
+	Vector3,
+	Box3,
+	Triangle,
+	CanvasTexture,
+} from 'three';
```
gkjohnson: Can we revert these style changes to this file and others that aren't related to this PR? Ideally only the code relevant changes for WMS would be included. Otherwise it's a lot of noise to comb through to make sure I'm understanding everything that's been added.

example/wmsOverlay.js
```diff
@@ -0,0 +1,360 @@
+import { Scene, WebGLRenderer, PerspectiveCamera } from 'three';
```
gkjohnson: I think it's okay for us to just have one WMS example to test the WMS tiling system. Why don't we delete "overlay" and keep the "wmsTiles" example.

example/wmsTiles.js
```diff
@@ -0,0 +1,307 @@
...
+	// xyzTiles.registerPlugin( new UpdateOnChangePlugin() );
+	// xyzTiles.registerPlugin(
+	// 	new XYZTilesPlugin( {
+	// 		center: true,
+	// 		shape: params.planar ? 'planar' : 'ellipsoid',
+	// 		url: 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
+	// 	} ),
+	// );
+	// xyzTiles.setCamera( camera );
+	// scene.add( xyzTiles.group );
```
gkjohnson: Lets remove some of the code and comments that are no longer bing used in the example.

example/wmsTiles.js
```diff
@@ -0,0 +1,307 @@
...
+	const rotatedPos = new THREE.Vector3( x, y, z ).applyMatrix4( rotationMatrix );
+
+	// 3. Place camera along the same direction, at a distance
+	const direction = rotatedPos.clone().normalize();
+	camera.position.copy( direction.multiplyScalar( cameraDistance ) );
+
+	// 4. Focus camera on the rotated marker position
+	camera.lookAt( rotatedPos );
+
+}
```
gkjohnson: Lets using the `Ellipsoid` API for this:

```js
tiles.ellipsoid.getObjectFrame( lat, lon, height, az, el, roll, camera.matrixWorld, CAMERA_FRAME );
camera.matrixWorld.decompose( camera.position, camera.quaternion, camera.scale );
```

example/wmsTiles.js
```diff
@@ -0,0 +1,307 @@
...
+
+	params = {
+
+		baseUrl: url, // Remove GetCapabilities params
+		layer: defaultLayer.name,
+		style: defaultLayer.styles[ 0 ]?.name || '',
+		crs: selectedCRS,
+		format: 'image/png',
+		tileDimension: 256,
+		planar: false,
```
gkjohnson: It looks like this option is never added to the GUI?

src/three/plugins/images/utils/ProjectionScheme.js
```diff
@@ -38,6 +38,7 @@ export class ProjectionScheme {
 				break;
 
 			default:
+				console.error( 'Unknown projection scheme:', scheme );
```
gkjohnson: Lets move this warning to the "error" message, instead:

```js
throw new Error( `ProjectionScheme: Unknown projection scheme "${ scheme }"` );
```

gkjohnson (commented):
Thanks for the changes! I've made a few initial comments - mostly about some unnecessary code style change which makes the new additions a bit more difficult to follow.

in the mean time I'll take a deeper look at the WMSCapabilitiesLoader this weekend. Thanks again!

src/three/plugins/loaders/WMSCapabilitiesLoader.js
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	async load( url ) {
+
+		const response = await fetch( url, this.fetchOptions );
+		if ( ! response.ok ) throw new Error( 'Failed to fetch WMS Capabilities' );
+		const buffer = await response.arrayBuffer();
+		return this.parse( buffer );
+
+	}
```
gkjohnson: A "loadAsync" function is already implemented in "LoaderBase" so we can remove this function from the loader and just implement "parse".

src/three/plugins/loaders/WMSCapabilitiesLoader.js
```diff
@@ -0,0 +1,136 @@
...
+			const service = parseService( xml.querySelector( 'Service' ) );
+			const capability = xml.querySelector( 'Capability' );
+			const layers = capability
+				? Array.from( capability.querySelectorAll( 'Layer' ) ).map( parseLayer )
+				: [];
+			resolve( { service, layers } );
+
+		} );
+
+	}
```
gkjohnson: Is there a reason this needs to return a promise?

src/three/plugins/loaders/WMSCapabilitiesLoader.js
```diff
@@ -0,0 +1,136 @@
...
+}
+
+export class WMSCapabilitiesLoader extends LoaderBase {
+
+	constructor( fetchOptions = {} ) {
+
+		super();
+		this.fetchOptions = fetchOptions;
+
+	}
```
gkjohnson: For consistency, let's remove the constructor and assign fetchOptions directly on the instance like the other loaders.

src/three/plugins/loaders/WMSCapabilitiesLoader.js
```diff
@@ -0,0 +1,136 @@
...
+		exGeographicBoundingBox: exGeoBBox,
+		styles,
+		subLayers,
+	};
+
+}
+
+function parseService( el ) {
+
+	if ( ! el ) return {};
```
gkjohnson: If there's no element we should still return an object with all the expected fields but empty so the object is reliably shaped.

src/three/plugins/loaders/WMSCapabilitiesLoader.js
```diff
@@ -0,0 +1,136 @@
...
+
+		return new Promise( ( resolve ) => {
+
+			const str = new TextDecoder( 'utf-8' ).decode( new Uint8Array( buffer ) );
+			const xml = new DOMParser().parseFromString( str, 'text/xml' );
+			const service = parseService( xml.querySelector( 'Service' ) );
+			const capability = xml.querySelector( 'Capability' );
+			const layers = capability
+				? Array.from( capability.querySelectorAll( 'Layer' ) ).map( parseLayer )
+				: [];
```
gkjohnson: We should probably parse all the "request" information, such as "GetMap" along with the supported image formats, url base, etc so these can be used when requesting tiles or other data rathe than hardcoding the fields.

src/three/plugins/loaders/WMSCapabilitiesLoader.d.ts
```diff
@@ -0,0 +1,41 @@
...
+export interface WMSLayer {
+    name: string;
+    title: string;
+    abstract: string;
+    queryable: boolean;
+    opaque: boolean;
+    keywords: string[];
+    crs: string[];
+    boundingBoxes: WMSBoundingBox[];
+    exGeographicBoundingBox: { bounds: [number, number, number, number] } | null;
```
gkjohnson: We should be able to shape this the same as "WMSBoundingBox" - the implicit "CRS" would be "EPSG:4326". It also doesn't look like we're using this "ex" bounding box, though. Do you think it's needed?

src/three/plugins/loaders/WMSCapabilitiesLoader.js
```diff
@@ -0,0 +1,136 @@
...
+	);
+	const bboxEls = Array.from( el.querySelectorAll( 'BoundingBox' ) );
+	const boundingBoxes = bboxEls.map( parseBoundingBox );
+	const exGeoBBox = parseEXGeographicBoundingBox(
+		el.querySelector( 'EX_GeographicBoundingBox' ),
+	);
+	const styles = Array.from( el.querySelectorAll( 'Style' ) ).map( parseStyle );
+
+	// Recursively parse sublayers
+	const subLayers = Array.from( el.querySelectorAll( 'Layer' ) ).map( parseLayer );
```
gkjohnson: Can you help me understand what the difference is between the "nested" layers and the root layer object? They seem to have a lot of the same information?

src/three/plugins/loaders/WMSCapabilitiesLoader.js
```diff
@@ -0,0 +1,136 @@
...
+function parseStyle( el ) {
+
+	return {
+		name: el.querySelector( 'Name' )?.textContent || '',
+		title: el.querySelector( 'Title' )?.textContent || '',
+		legendUrl:
+			el
+				.querySelector( 'LegendURL OnlineResource' )
+				?.getAttribute( 'xlink:href' ) || '',
+	};
```
gkjohnson: We should include the image format here, as well.

src/three/plugins/loaders/WMSCapabilitiesLoader.js
```diff
@@ -0,0 +1,136 @@
...
+
+function parseBoundingBox( el ) {
+
+	if ( ! el ) return null;
+	const crs = el.getAttribute( 'CRS' ) || el.getAttribute( 'crs' );
+	const minx = parseFloat( el.getAttribute( 'minx' ) || '0' );
+	const miny = parseFloat( el.getAttribute( 'miny' ) || '0' );
+	const maxx = parseFloat( el.getAttribute( 'maxx' ) || '0' );
+	const maxy = parseFloat( el.getAttribute( 'maxy' ) || '0' );
+	return { crs, bounds: [ minx, miny, maxx, maxy ] };
```
gkjohnson: We should "correct" the bounding boxes as we area in [WMTSCapabilitiesLoader](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/4b8495373758e5df5bc7ed00e82e10ee883c8852/src/three/plugins/loaders/WMTSCapabilitiesLoader.js#L36-L63) (feel free to move some of those utility functions to a separate file). Basically the loader converts all degrees to radians (the units 3d tiles renderer uses), ensures all bounding box definitions are in `[ lon, lat, lon, lat ]` order, converts "CRS:84" to "EPSG:4326", and converts the odd [EPGS:3857 meters units to cartographic units](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/4b8495373758e5df5bc7ed00e82e10ee883c8852/src/three/plugins/loaders/WMTSCapabilitiesLoader.js#L48-L63).

I'm realizing it may be best to handle this elsewhere, such as in the "TilingScheme" or "ProjectionScheme" classes but this is how the project is currently structured so I think any larger changes will have to wait for another PR. Worth discussing in another issue, I think.

src/three/plugins/images/sources/WMSImageSource.js
```diff
@@ -0,0 +1,152 @@
...
+			} else {
+
+				// WMS 1.3.0: lat/lon order
+				const minx = this.convertProjectionToLongitude( bbox[ 0 ] );
+				const miny = this.convertProjectionToLatitude( bbox[ 1 ] );
+				const maxx = this.convertProjectionToLongitude( bbox[ 2 ] );
+				const maxy = this.convertProjectionToLatitude( bbox[ 3 ] );
+				bboxParam = [ miny, minx, maxy, maxx ]; // lat,lon,lat,lon
+
+			}
```
gkjohnson: This is fun 😅 Is there a reference anywhere discussing the bounding box parameter order change? I think we can fix this in the loader, too.

Is this true of all WMS 1.1.1 data sets? Or just those from a specific tool?

src/three/plugins/images/sources/WMSImageSource.js
```diff
@@ -0,0 +1,152 @@
...
+		super();
+		this.url = baseUrl;
+		this.layer = layer;
+		this.crs = crs;
+		this.format = format;
+		this.tileDimension = tileDimension;
+		this.styles = styles;
+		this.version = version;
+		this.bounds = bounds;
+		this.levels = levels;
```
gkjohnson, line 30: I see that the WMS capabilities file doesn't seem to define a "max level" for the tiles. Is it the case that WMS lets you zoom in to the data set an arbitrary amount?

src/three/plugins/images/EPSGTilesPlugin.js
```diff
@@ -52,23 +53,59 @@ export class WMTSTilesPlugin extends EllipsoidProjectionTilesPlugin {
...
+
+		this.name = 'WMS_TILES_PLUGIN';
+		this.imageSource = new WMSImageSource( {
+			baseUrl,
+			layer,
+			crs,
+			format,
+			tileDimension,
+			styles,
+			version
```
gkjohnson: It looks like a number of arguments (like "bounds") are not passed here.

src/three/plugins/images/ImageOverlayPlugin.js
SoftwareMechanic: Of course, I am sorry for that, I probably run a prettifier and didn't know esling was used, will fix this!

example/wmsOverlay.js
SoftwareMechanic: yes I could remove overlay, of course. but just for the moment I keep it, because I am having issues with planar rendering (only with overlay plugin), where wms fetch everything correctly, but I can't see the overlay, I would like to understand why, maybe this can lead to a bug in WMS implementation

example/wmsTiles.js
SoftwareMechanic: in wmsTiles example I was having the issue with planar rendering, but as I solved it locally, this will be fixed (note: still having issue with the overlay in planar rendering)

src/three/plugins/loaders/WMSCapabilitiesLoader.js
SoftwareMechanic: you are right, there is no reason for it

src/three/plugins/loaders/WMSCapabilitiesLoader.d.ts
SoftwareMechanic: I made some test with it, because initially I thought this could be used to define the bounds for WMSImageSource, but I was wrong, seems something we actually don't need, I could remove it eventually

src/three/plugins/loaders/WMSCapabilitiesLoader.js
SoftwareMechanic: Totally agree, thank you for pointing that out.

src/three/plugins/loaders/WMSCapabilitiesLoader.js
SoftwareMechanic: with services I used in my use cases, the root layers are just a way to define a group of layers, therefore I found out that root layers are not queryable, but considering the layer structure from the WMS capabilities, I would not consider that true in every case.
Ideally in a project, if a user want to focus on WMS, he should create a tree representation of the layers, and make selectable only those with the property 'IsQueryable' set to 1 from the XML.
But I think It could happen that even a root layer is selectable (maybe achieving the same result as activating all the sublayers with less requests to the server).

in the meantime I just filtered out from the GUI params the selectable layers.

src/three/plugins/loaders/WMSCapabilitiesLoader.js
SoftwareMechanic: I agree, in the meantime I followed the WMTS approach for this PR

src/three/plugins/images/sources/WMSImageSource.js
SoftwareMechanic: here we have some refs talking about it, it is true for every WMS 1.1.1
I report a statement here: 

> v1.1.1
Released in Jan 2002
    Use SRS for 1.1.1
    In WMS 1.1.1 EPSG:4326 is wrongly defined as having long/lat coordinate axes. See v1.3.0 documentation for details.


[doc](https://docs.foursquare.com/analytics-products/docs/data-formats-wms)
[geo topic](https://gis.stackexchange.com/questions/23347/getmap-wms-1-1-1-vs-1-3-0)

src/three/plugins/images/sources/WMSImageSource.js
SoftwareMechanic: personally I tried using 5 levels and also 100, it worked as expected, with different wms services.
probably there is a limit server side, but I am not sure about it.

src/three/plugins/images/EPSGTilesPlugin.js
SoftwareMechanic: I was using arbitrary bounds for testing in the beginning, trying to take them from the layer, I thought that could optimize the wms requests, now they are defined like for the WMTS, without needing for them to be defined externally.

SoftwareMechanic:
in the new PR I fixed most of the things, I will answer here to your questions:


> seems that wms requests are not triggered with planar rendering

> Is this still an issue?

now in planar rendering requests are triggered correctly, in wmsTiles example you can see it working now!
 but, in the wmsOverlay example, I can't make it work, even if requests are triggered (using the frame property to WMSTilesOverlay when planar is ture), I don't see the images projected on the plane. still have to figure out why that happen.  ( I found out that there are no examples using planar mode and an overlay, is this an issue maybe? )


> using EPSG:3857 seems to work better

> Can you elaborate on what is better about it?

when I was doing the first PR I realized that using EPSG:3857, I could zoom out a lot, and the layer was still visible, while using the EPSG:4326 it was disappearing, but that was related to a previous version, now it should be fixed, as you can try change the crs (between these two) from GUI and the WMS will work fine.


thanks a lot for your time,
for anything else, just let me know

example/wmsOverlay.js
gkjohnson: The "planar projection" isn't so straightforward and require defining a specific frame in tile-space to use for the planar projection. It still needs some work and documentation for improve the clarity and behavior of things but overall I'm seeing that WMSTilesPlugin works just fine with planar projection when modifying [this r3f demo](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/r3f/projection.html):

Image

So to me it looks like everything is okay and we can remove this file, I think.

src/three/plugins/images/sources/WMSImageSource.js
gkjohnson: It looks like the WMS capabilities file doesn't define a "max level" though. It looks like there may be some `MinScaleDenominator`, `MaxScaleDenominator`, and `TileLevel` tags that may be available at times but not in the data we're looking at here. I think it's okay to rely on manual level setting for now.

src/three/plugins/images/sources/WMSImageSource.js
gkjohnson: Great, thanks! I think it would be good to include a link to one of these topics inline as a comment so the rationale is clear.

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1474,6 +1475,49 @@ export class XYZTilesOverlay extends ImageOverlay {
...
+
+			const { extraHeaders } = this.imageSource;
+			options.headers = options.headers || {};
+			Object.entries( extraHeaders ).forEach( ( [ key, value ] ) => {
+
+				options.headers[ key ] = value;
+
+			} );
+
+		}
```
gkjohnson: If you need extra auth headers for your requests then these should be handled in a separate plugin unless there's something WMS-specific about this. You should be able to handle this like so:

```js
tiles.registerPlugin( {
  fetch( url, options = {} ) {
    if ( /* test for url pattern */ ) {
      options.headers = options.headers || {};
      options.headers = {
        ...options.headers,
        ...extraHeaders,
      };
    }
  }
} );
```

Or it may be enough to just set the `fetchOptions` on the tiles renderer itself so they're included in all requests:

```js
tiles.fetchOptions.headers = extraHeaders;
```

src/three/plugins/loaders/WMSCapabilitiesLoader.d.ts
gkjohnson: Okay lets remove it for now, then.

src/three/plugins/loaders/WMSCapabilitiesLoader.js
gkjohnson: I'm not sure if I'm fully following.

> with services I used in my use cases, the root layers are just a way to define a group of layers, therefore I found out that root layers are not queryable,

It doesn't seem to be the case always. There are some other datasets [here](https://www.gebco.net/data-products/gebco-web-services/web-map-service), as well, that have some more complex capabilities XML. The root layer is called "GEBCO_Grid" and and can be loaded as a map (see [this link](https://wms.gebco.net/mapserv?request=getMap&service=wms&version=1.3.0&layers=GEBCO_Grid&CRS=EPSG:4326&BBOX=-90,-180,90,180&FORMAT=image/png&WIDTH=1024&HEIGHT=1024)). The data set used as a demo in this PR seems a bit oddly formed with a lot of `<![CDATA[WMS]]>` tokens throughout the file - and is even used as the root "layer name" which causes the service to fail when you try to load it. Is it possible there's an issue with this particular service and it's serving bad data?

Interestingly the above services _only_ defines `EX_GeographicBoundingBox` options with no other bounds. So it seems likely that we have to keep continue parsing that bounding box definition.

> make selectable only those with the property 'IsQueryable' set to 1 from the XML.

From the [WMS wkipedia page](https://en.wikipedia.org/wiki/Web_Map_Service#:~:text=Request%20types%20that%20WMS%20providers,coordinate%20of%20the%20map%20image.) it seems to be that "queryable" only refers to the ability to call "GetFeatureInfo" and isn't related to whether map data can be retrieved.

Regarding nested layers it seems that each layer should be individually loadable and the nested serves as both a form of organization and as a means for child layers to inherit some properties from the parent. See [this SO answer here](https://gis.stackexchange.com/questions/482376/in-a-wms-getcapabilities-response-is-layer-allowed-multiple-times-directly-wi)

gkjohnson (commented):
Thanks for the changes - I pushed some updates myself to remove more of the code style adjustments. I've also made a few more comments. Let me know what you think and I can plan to help out with some of the remaining changes, as well, if needed.

example/wmsOverlay.js
SoftwareMechanic: I thought it would be possible to have let's say the XYZ plugin, in planar mode, with the WMS as overlay (lots of WMS services don't cover the entire globe like this one used in examples, but just a portion of the map).
but in any case for my project I can use just the globe, it's not really an issue for me.
I delete the example as requested :)

src/three/plugins/loaders/WMSCapabilitiesLoader.js
SoftwareMechanic: > It doesn't seem to be the case always. There are some other datasets [here](https://www.gebco.net/data-products/gebco-web-services/web-map-service), as well, that have some more complex capabilities XML. The root layer is called "GEBCO_Grid" and and can be loaded as a map (see [this link](https://wms.gebco.net/mapserv?request=getMap&service=wms&version=1.3.0&layers=GEBCO_Grid&CRS=EPSG:4326&BBOX=-90,-180,90,180&FORMAT=image/png&WIDTH=1024&HEIGHT=1024)). 

according to this, we need to understand better when allow the user to select a layer for getting the relative images.

> The data set used as a demo in this PR seems a bit oddly formed with a lot of <![CDATA[WMS]]> tokens throughout the file - and is even used as the root "layer name" which causes the service to fail when you try to load it.

After the XML is parsed I can see the correct titles and names, I didn't face a failure from wms while loading a layer, but I found out that after parsing I end up with 2 layers with the same name ('0', which seems to be the identifier for the layer used, so to me there is something wrong), and one is nested in the root.

>  Is it possible there's an issue with this particular service and it's serving bad data?

I don't exclude that, I used the first service I found after changing it from the one I was using (which was too heavy), but also with that service, if I remember correctly I had this issue of multiple layer with same name.


Eventually if you want I can set up a different WMS for the example.
✌️ 

> Interestingly the above services only defines EX_GeographicBoundingBox options with no other bounds. So it seems likely that we have to keep continue parsing that bounding box definition.

Ok, I am going to keep that property!

src/three/plugins/images/ImageOverlayPlugin.js
SoftwareMechanic: in the case of using the WMSTilesOverlay, considering that a user could have different WMS layers from different services with different auth headers, overlayed on top of other tiles (cesium globe, photorealistic google tiles, xyz etc...), how do I use 'extraHeaders' in that situation for each overlay? 
should I register an ImageOverlayPlugin for each overlay I want?
and how do I register headers for each of them eventually?

in the case of using WMSTilesPlugin with extra headers, I don't understand your example,
I tried to mimic your code, registering a fetch method, but never logs anything.

```
	wmsTiles.registerPlugin( {
  		fetch( url, options = {} ) {
		console.log("url", url);
		console.log("options", options);
		}
	} )
```

> If you need extra auth headers for your requests then these should be handled in a separate plugin

if you mean to have an AuthPlugin for WMS, I don't know honestly, in my use cases I only had to add an Authorization header, and still I have not figured yet how the AuthPlugins work in this library and how they manipulate the plugin requests.

probably I am missing something about it, I am sorry for that.

I must be sure that auth headers for WMS are managed in the most flexible way

SoftwareMechanic:
Ok for now I updated just the comment for the differencies between WMS 1.1.1 and 1.3.0, still have to understand better the header management, and the WMS selectable layers for the example.

thank you a lot for your time also, still available if you need.

example/wmsOverlay.js
gkjohnson: > I thought it would be possible to have let's say the XYZ plugin, in planar mode, with the WMS as overlay (lots of WMS services don't cover the entire globe like this one used in examples, but just a portion of the map).

I see - I think this would be something good to support in the future but isn't the intent for "planar" projection. Planar projection is designed to support projecting images on top of terrain - eg small aerial patches, robotics camera images, etc. Perhaps it's worth making another issue to figure out the best way to handle cartographic-aligned planar projection, as well.

src/three/plugins/images/ImageOverlayPlugin.js
gkjohnson: I see, now - the ImageOverlayPlugins are not hooked up to the tiles renderer fetch callback correctly 🤔 Do you mind making a separate issue for this? You're right this is something that needs to be supported. I would just like to come up with a more universal solution than adding separate authentication logic for every plugin and overlay that needs it if possible.

SoftwareMechanic: Yes exactly, that is the point which I better described in a new issue as you suggested.
#1308

gkjohnson:
I just pushed a few more changes to align the capabilities loader with the expectations of the [WMS Schema](https://schemas.opengis.net/wms/1.3.0/capabilities_1_3_0.xsd) and added a few extra fields to simplify the example a bit (I may change the WMTSLoader similarly, as well). Overall this is really great - my only concern is that the way the project has been handling CRS' may not be appropriate for WMS. Basically the previous loaders have been converting everything into cartographic lat / lon (`[-180, 180], [-90, 90]`) since there was no need to operate in anything else. But it looks like the WMS requests may require different units / scales, which are lost during the conversion. I'll need a bit of time to think about this and will also want to test on a few other data sets (including [this one](https://www.gebco.net/data-products/gebco-web-services/web-map-service) I referenced previously).

Thanks for bearing with me on this. I'm quite glad to have this feature added - there are just a few assumptions in the project that make it not so straight forward. I'll need a bit more time to think about what the right thing to do is. Thank you again.

src/three/plugins/loaders/WMSCapabilitiesLoader.js
```diff
@@ -0,0 +1,360 @@
...
+	if ( dcp.length === 0 ) {
+
+		const online = opEl.querySelector( ':scope > OnlineResource' );
+		if ( online ) {
+
+			dcp.push( { type: 'HTTP', get: readOnlineResourceHref( online ), post: '' } );
+
+		}
+
+	}
```
gkjohnson: For these lines here - do you have examples of some of these non-standard ways of representing the "OnlineResource" fields? According to [this scheme](https://schemas.opengis.net/wms/1.3.0/capabilities_1_3_0.xsd) the "OnlineResource" field is supposed to _always_ be inside a "Get" and "Post" element. But it looks like you're checking for the url in a few different places. My preference is to only check for the expected location as defined by the specification but if its common for some of the services to place it in odd locations then we can keep the check here - I just want to make sure it's needed.

gkjohnson:
Okay I've got [these data sets](https://www.gebco.net/data-products/gebco-web-services/web-map-service#requests) working with both versions 1.3.0 and 1.1.1 (though the requests are quite slow):

image


And made some changes to how CRSs and bounding boxes are handled as well as updated some WMTS issues I found along the way. I think this should be good to go, now but let me know what you think. We can deal with the request headers issue with #1308 and another PR.

I'll also plan to make another few issues for aligning the APIs of WMTS and WMS a bit more. I would like them to be structured so you can easy drop in parsed data from the "Capabilities" file _or_ construct the parameters with literal values so you don't have load the capabilities file (which can be huge) if you don't want to. We'll have to figure out how to deal with that in vanilla and r3f, for example.

_edit: it looks like the previously-linked wms capabilities file includes "MaxScaleDenominator" field which may be able to be used to infer the max number of loadable levels?_

SoftwareMechanic:
thank you a lot for all the tests and commits. I appreciate!!

> it looks like the previously-linked wms capabilities file includes "MaxScaleDenominator" field which may be able to be used to infer the max number of loadable levels?

I found a pdf which explain it more precisely, doesn't look like simple to implement the management of MinScaleDenominator and MaxScaleDenominator, but certainly has to do with the zoom levels and visibility.
there are more info from page 26 of the following pdf, can you tell me what you think about it, should we implement it, and eventually where could be ideal to manage that information in the WMS workflow?

[doc](http://portal.opengeospatial.org/files/?artifact_id=1188)


> my only concern is that the way the project has been handling CRS' may not be appropriate for WMS. Basically the previous loaders have been converting everything into cartographic lat / lon ([-180, 180], [-90, 90]) since there was no need to operate in anything else. But it looks like the WMS requests may require different units / scales, which are lost during the conversion.

please let me know what you plan to do about that, I am still not 100% into this library so I have to understang some things, like what you mean by "the way the project has been handling CRS'", but eventually, I would be happy to help if possible.

in any case I think this is already a great result, improvable of course,  but I tested different services and to me, as a user (not very expert on WMS) seems work nice as other libraries, given the capabilities, the user can also play a lot with WMS relative functions as well, like displaying the legends, or reading feature Info [simple example](url) by getting clicked coordinates etc...

I will also look at the other issues you opened!

the potential for this library is 🚀 🚀

gkjohnson:
> I found a pdf which explain it more precisely, doesn't look like simple to implement the management of MinScaleDenominator and MaxScaleDenominator, but certainly has to do with the zoom levels and visibility.
> ...
>  can you tell me what you think about it, should we implement it

I think it's okay to skip implementing it for now. We can add it if it becomes necessary in the future.

> please let me know what you plan to do about that

I think I've figured out what to do about the CRS' in this PR. For the most part the CRS values are used to infer the type of distortion that's used in an image so it can be accounted for in projection (until there's something else they need to be used for). So the CRS ranges are converted to a typical cartographic range in a consistent lon / lat order ([-180, -90, 180, 90] in radians) and then the WMS plugin converts that to the appropriate order and units for request (meters for web mercator, for example).

The bounding boxes from the "BoundingBox" tags are also not used at all at the moment since from what I can tell they're just defining the ranges of the given CRS? And the `LatLonBoundingBox` or `EX_GeographicBoundingBox` define the actual geographic range of the data. This [SO answer](https://gis.stackexchange.com/questions/18894/valid-ranges-for-bounding-box-in-wms-1-1-1) seems to indicate that, at least.

Overall I think this can be merged as it is. I'll do another review in the next day or so before merging it, though.

---

Also - I'd include both the Geojson & WMS overlay contributions in an upcoming blog post for Cesium and of course would like to give credit if you're comfortable with it. If you could get in touch over email (in my profile) we can discuss it a bit more.

SoftwareMechanic:
> I think it's okay to skip implementing it for now. We can add it if it becomes necessary in the future.

I think this is fine, in the end is working now :)

> The bounding boxes from the "BoundingBox" tags are also not used at all at the moment since from what I can tell they're just defining the ranges of the given CRS? And the LatLonBoundingBox or EX_GeographicBoundingBox define the actual geographic range of the data. This [SO answer](https://gis.stackexchange.com/questions/18894/valid-ranges-for-bounding-box-in-wms-1-1-1) seems to indicate that, at least.

in the beginning I was asking myself if they could be used in order to optimize the rendering, maybe considering the geographic bounding box in the camera view. 
but to mee seems already working fine and eventually difficult to implement yet.

> Overall I think this can be merged as it is. I'll do another review in the next day or so before merging it, though.

🚀 🚀🚀 awesome

> Also - I'd include both the Geojson & WMS overlay contributions in an upcoming blog post for Cesium and of course would like to give credit if you're comfortable with it. If you could get in touch over email (in my profile) we can discuss it a bit more.

Thank you so much, yes I would appreciate that a lot!
I will contact you soon by email!

src/three/plugins/loaders/WMSCapabilitiesLoader.js
```diff
@@ -0,0 +1,388 @@
...
+
+	const formats = Array.from( opEl.querySelectorAll( 'Format' ) ).map( f => f.textContent.trim() );
+	const dcp = Array.from( opEl.querySelectorAll( 'DCPType' ) ).map( dcp => {
+
+		const httpEl = dcp.querySelector( 'HTTP' );
+		if ( ! httpEl ) {
+
+			return { type: 'UNKNOWN', get: '', post: '' };
+
+		}
```
gkjohnson: Do any of the data sets you encountered actually include a DCP type other than "HTTP"? From the schema it sounds like only HTTP is really defined:

```xml
<element name="DCPType">
  <annotation>
    <documentation> Available Distributed Computing Platforms (DCPs) are listed here. At present, only HTTP is defined. </documentation>
  </annotation>
  ...
</element>
```

I'm thinking we can remove this check for now until we have an example of it being an issue?

SoftwareMechanic: I think you are right here, I've only met HTTP  as DCP type, and seems to be the only possible value, I remove this

gkjohnson:
This is great! Thanks again for the addition. And bearing with me through the change requests 😅

PR #1305, opened 2025-09-05
Add GeoJSON format support

SoftwareMechanic:
added geojson support as overlay and tilesplugin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - LoadRegionPlugin: Added support for "mask" regions that allow for only loading tiles within the shape.
 - Added WMSTilesOverlay & WMSTilesPlugin for displaying and overlaying WMS image tiles.
+- I3DMLoader: Add support for quantized position data.
+
+### Fixed
+- Fixed types not being defined correctly in the package.json.
 
 ## [0.4.16] - 2025.09.10
 ### Fixed
--- a/example/loadRegion.js
+++ b/example/loadRegion.js
@@ -30,6 +30,7 @@ const params = {
 	region: 'SPHERE',
 	regionErrorTarget: 0.1,
 	regionOnly: true,
+	mask: false,
 	displayBoxBounds: false,
 
 };
@@ -110,6 +111,7 @@ function init() {
 	gui.add( params, 'region', [ 'SPHERE', 'BOX', 'RAY' ] ).onChange( updateRegion );
 	gui.add( params, 'regionErrorTarget' ).min( 0 ).max( 1 );
 	gui.add( params, 'animate' );
+	gui.add( params, 'mask' );
 	gui.add( params, 'regionOnly' ).onChange( v => {
 
 		if ( ! v ) {
@@ -192,7 +194,8 @@ function animate() {
 		sphereMesh.position.set( Math.sin( time ) * 20, 0, Math.cos( time ) * 20 );
 		sphereMesh.scale.setScalar( sphereRegion.sphere.radius );
 
-		sphereRegion.errorTarget = params.regionErrorTarget;
+		sphereRegion.mask = params.mask;
+		sphereRegion.errorTarget = params.mask ? Infinity : params.regionErrorTarget;
 		sphereRegion.sphere.center
 			.copy( sphereMesh.position )
 			.applyMatrix4( tiles.group.matrixWorldInverse );
@@ -201,7 +204,8 @@ function animate() {
 
 		rayMesh.position.set( Math.sin( time * 2 ) * 20, 50, Math.cos( time * 2 ) * 20 );
 
-		rayRegion.errorTarget = params.regionErrorTarget;
+		rayRegion.mask = params.mask;
+		rayRegion.errorTarget = params.mask ? Infinity : params.regionErrorTarget;
 		rayRegion.ray.direction
 			.set( 0, - 1, 0 )
 			.transformDirection( tiles.group.matrixWorldInverse );
@@ -216,7 +220,8 @@ function animate() {
 		boxMesh.updateMatrixWorld();
 		boxMesh.geometry.computeBoundingBox();
 
-		boxRegion.errorTarget = params.regionErrorTarget;
+		boxRegion.mask = params.mask;
+		boxRegion.errorTarget = params.mask ? Infinity : params.regionErrorTarget;
 		boxRegion.obb.box.copy( boxMesh.geometry.boundingBox );
 		boxRegion.obb.transform.copy( boxMesh.matrixWorld ).premultiply( tiles.group.matrixWorldInverse );
 		boxRegion.obb.update();
--- a/package.json
+++ b/package.json
@@ -24,36 +24,36 @@
   ],
   "exports": {
     ".": {
-      "import": "./build/index.js",
-      "types": "./src/index.d.ts"
+      "types": "./src/index.d.ts",
+      "import": "./build/index.js"
     },
     "./plugins": {
-      "import": "./build/index.plugins.js",
-      "types": "./src/plugins.d.ts"
+      "types": "./src/plugins.d.ts",
+      "import": "./build/index.plugins.js"
     },
     "./core": {
-      "import": "./build/index.core.js",
-      "types": "./src/core/renderer/index.d.ts"
+      "types": "./src/core/renderer/index.d.ts",
+      "import": "./build/index.core.js"
     },
     "./three": {
-      "import": "./build/index.three.js",
-      "types": "./src/three/renderer/index.d.ts"
+      "types": "./src/three/renderer/index.d.ts",
+      "import": "./build/index.three.js"
     },
     "./r3f": {
-      "import": "./build/index.r3f.js",
-      "types": "./src/r3f/index.d.ts"
+      "types": "./src/r3f/index.d.ts",
+      "import": "./build/index.r3f.js"
     },
     "./core/plugins": {
-      "import": "./build/index.core-plugins.js",
-      "types": "./src/core/plugins/index.d.ts"
+      "types": "./src/core/plugins/index.d.ts",
+      "import": "./build/index.core-plugins.js"
     },
     "./three/plugins": {
-      "import": "./build/index.three-plugins.js",
-      "types": "./src/three/plugins/index.d.ts"
+      "types": "./src/three/plugins/index.d.ts",
+      "import": "./build/index.three-plugins.js"
     },
     "./src/*": {
-      "import": "./src/*",
-      "types": "./src/*.d.ts"
+      "types": "./src/*.d.ts",
+      "import": "./src/*"
     }
   },
   "scripts": {
--- a/src/three/plugins/LoadRegionPlugin.d.ts
+++ b/src/three/plugins/LoadRegionPlugin.d.ts
@@ -4,8 +4,9 @@ import { Sphere, Ray } from 'three';
 
 declare class BaseRegion {
 
-	constructor( errorTarget?: number );
 	errorTarget: number;
+	mask: boolean;
+	constructor( options?: { errorTarget?: number, mask?: boolean } );
 	intersectsTile( boundingVolume: TileBoundingVolume, tile: Tile, tilesRenderer: TilesRenderer ): boolean
 	calculateError( tile: Tile, tilesRenderer: TilesRenderer ): number
 
@@ -23,20 +24,20 @@ export class LoadRegionPlugin {
 export class OBBRegion extends BaseRegion {
 
 	obb: OBB;
-	constructor( errorTarget?: number, obb?: OBB );
+	constructor( options?: { errorTarget?: number, mask?: boolean, obb?: OBB } );
 
 }
 
 export class RayRegion extends BaseRegion {
 
 	ray: Ray;
-	constructor( errorTarget?: number, ray?: Ray );
+	constructor( options?: { errorTarget?: number, mask?: boolean, ray?: Ray } );
 
 }
 
 export class SphereRegion extends BaseRegion {
 
 	sphere: Sphere;
-	constructor( errorTarget?: number, sphere?: Sphere );
+	constructor( options?: { errorTarget?: number, mask?: boolean, sphere?: Sphere } );
 
 }
--- a/src/three/plugins/LoadRegionPlugin.js
+++ b/src/three/plugins/LoadRegionPlugin.js
@@ -50,27 +50,41 @@ export class LoadRegionPlugin {
 
 	}
 
+	// Calculates shape intersections and associated error values to use. If "mask" shapes are present then
+	// tiles are only loaded if they are within those shapes.
 	calculateTileViewError( tile, target ) {
 
 		const boundingVolume = tile.cached.boundingVolume;
 		const { regions, tiles } = this;
 
-		let visible = false;
+		let inShape = false;
+		let inMask = null;
 		let maxError = - Infinity;
 		for ( const region of regions ) {
 
+			// Check if the tile is intersecting the shape and calculate the
+			// view and error values.
 			const intersects = region.intersectsTile( boundingVolume, tile, tiles );
-			if ( intersects ) {
+			inShape = inShape || intersects;
+			maxError = Math.max( region.calculateError( tile, tiles ), maxError );
 
-				visible = true;
-				maxError = Math.max( region.calculateError( tile, tiles ), maxError );
+			// Store whether the tile is in a "mask" shape if they exist. If "inMask" is
+			// still "null" by the end of the loop then there are no mask elements.
+			if ( region.mask ) {
+
+				inMask = inMask || intersects;
 
 			}
 
 		}
 
-		target.inView = visible;
+		// A tile should only be visible if it intersects a shape and a mask shape or there
+		// are no masks.
+		target.inView = inShape && inMask !== false;
 		target.error = maxError;
+
+		// Returning "false" indicates "no operation" and all results will be ignored.
+		return target.inView || inMask !== null;
 
 	}
 
@@ -85,9 +99,22 @@ export class LoadRegionPlugin {
 // Definitions of predefined regions
 export class BaseRegion {
 
-	constructor( errorTarget = 10 ) {
+	constructor( options = {} ) {
+
+		if ( typeof options === 'number' ) {
+
+			console.warn( 'LoadRegionPlugin: Region constructor has been changed to take options as an object.' );
+			options = { errorTarget: options };
+
+		}
+
+		const {
+			errorTarget = 10,
+			mask = false,
+		} = options;
 
 		this.errorTarget = errorTarget;
+		this.mask = mask;
 
 	}
 
@@ -103,9 +130,21 @@ export class BaseRegion {
 
 export class SphereRegion extends BaseRegion {
 
-	constructor( errorTarget = 10, sphere = new Sphere() ) {
+	constructor( options = {} ) {
 
-		super( errorTarget );
+		if ( typeof options === 'number' ) {
+
+			console.warn( 'SphereRegion: Region constructor has been changed to take options as an object.' );
+			options = {
+				errorTarget: arguments[ 0 ],
+				sphere: arguments[ 1 ],
+			};
+
+		}
+
+		const { sphere = new Sphere() } = options;
+
+		super( options );
 		this.sphere = sphere.clone();
 
 	}
@@ -120,9 +159,21 @@ export class SphereRegion extends BaseRegion {
 
 export class RayRegion extends BaseRegion {
 
-	constructor( errorTarget = 10, ray = new Ray() ) {
+	constructor( options = {} ) {
 
-		super( errorTarget );
+		if ( typeof options === 'number' ) {
+
+			console.warn( 'RayRegion: Region constructor has been changed to take options as an object.' );
+			options = {
+				errorTarget: arguments[ 0 ],
+				ray: arguments[ 1 ],
+			};
+
+		}
+
+		const { ray = new Ray() } = options;
+
+		super( options );
 		this.ray = ray.clone();
 
 	}
@@ -137,9 +188,21 @@ export class RayRegion extends BaseRegion {
 
 export class OBBRegion extends BaseRegion {
 
-	constructor( errorTarget = 10, obb = new OBB() ) {
+	constructor( options = {} ) {
 
-		super( errorTarget );
+		if ( typeof options === 'number' ) {
+
+			console.warn( 'RayRegion: Region constructor has been changed to take options as an object.' );
+			options = {
+				errorTarget: arguments[ 0 ],
+				obb: arguments[ 1 ],
+			};
+
+		}
+
+		const { obb = new OBB() } = options;
+
+		super( options );
 		this.obb = obb.clone();
 		this.obb.update();
 
--- a/src/three/plugins/README.md
+++ b/src/three/plugins/README.md
@@ -727,6 +727,10 @@ You can define a custom region by providing functions to determine Intersection 
 
 ```js
 {
+	// Indicates whether the region should be considered a "mask". If true then tiles outside the shape will not
+	// be loaded even if they are in the camera view.
+	mask: boolean,
+
 	// Custom function that determines whether a tile's bounding volume intersects this region.
 	intersectsTile: ( boundingVolume: TileBoundingVolume, tile: Tile, tilesRenderer: TilesRenderer ) => boolean,
 
--- a/src/three/renderer/loaders/I3DMLoader.js
+++ b/src/three/renderer/loaders/I3DMLoader.js
@@ -79,7 +79,10 @@ export class I3DMLoader extends I3DMLoaderBase {
 					loader.parse( gltfBuffer, workingPath, model => {
 
 						const INSTANCES_LENGTH = featureTable.getData( 'INSTANCES_LENGTH' );
-						const POSITION = featureTable.getData( 'POSITION', INSTANCES_LENGTH, 'FLOAT', 'VEC3' );
+						let POSITION = featureTable.getData( 'POSITION', INSTANCES_LENGTH, 'FLOAT', 'VEC3' );
+						const POSITION_QUANTIZED = featureTable.getData( 'POSITION_QUANTIZED', INSTANCES_LENGTH, 'UNSIGNED_SHORT', 'VEC3' );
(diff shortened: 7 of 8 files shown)


example/geojson.js
```diff
@@ -0,0 +1,252 @@
...
+	XYZTilesPlugin,
+} from '3d-tiles-renderer/plugins';
+import * as THREE from 'three';
+
+import { DRACOLoader } from 'three/examples/jsm/loaders/DRACOLoader.js';
+const dracoLoader = new DRACOLoader().setDecoderPath(
+	'https://www.gstatic.com/draco/v1/decoders/',
+);
+
+const geojson = {
```
gkjohnson, line 11: What's the source for this polygon data?

example/geojson.js
```diff
@@ -0,0 +1,252 @@
...
+	// 	} ),
+	// } );
+
+	// Use WebMercator lat clamp for EPSG:3857 tiles
+	const WEBMERCATOR_LAT_LIMIT = 85.0511287798066;
+	const overlayBounds = [ - 180, - WEBMERCATOR_LAT_LIMIT, 180, WEBMERCATOR_LAT_LIMIT ];
+
+	const geojsonOverlay = new GeoJSONTilesOverlay( {
+
+		levels: 14, // choose max zoom to rasterize overlay ( 12- 14 typical )
```
gkjohnson: This is an interesting new case - this is a good solution for now. It may be worth adjusting overlays so that they can load up to whatever maximum resolution the underlying tile set is at in order to more easily support these kinds of dynamically generated / flexible types of data sets.

_edit created #1310_

gkjohnson (commented):
This is great to see - and a really nice approach to overlaying and generating it.

Can you explain your use case a bit? Are you planning to have multiple polygons with multiple colors? One potential issue I see is that multiple colored geojson data sets will require separate layers of images which could cause a bit of an explosion of memory.

One thought I had is that the geojson could be converted to SVG and then you'd have all the flexibility of SVG scalability and CSS styling. This [geojson2svg](https://github.com/gagan-bansal/geojson2svg) package supports Geojson > SVG conversion. Who know's how the performance may be affected, though 🤔 Something to explore in the future.

_edit_ And sorry for the commit rework. I wanted to remove the dependency on #1302 so we can get this merged separately if needed.

example/geojson.js
SoftwareMechanic: [bacino_varenna_aggregato4326.json](https://github.com/user-attachments/files/22235072/bacino_varenna_aggregato4326.json)
I have this local file geojson (in order to load it here on Github, I had to change the extension from ._geojson_ to ._json_), I wanted to try to have both an url or a geojson object.
the reasons for geojson object was because I wanted to start thinking about a "drawPlugin",
but also because maybe the front end developer could handle a conversion of it to the accepted EPSG,
in any case if just URL is provided, we should make sure that the geojson is defined in supported EPSG codes looking at 'crs.properties.name' of the object.

what do you think about this?
we can just keep the url if you prefer, to make it coherent with the library

example/geojson.js
SoftwareMechanic: absolutely agree, I think that somehow we should maybe have access to the levels defined for the plugin the geojson is onto in order to manage that dynamically.
or hardcode an high number of levels if the prop is not provided?

SoftwareMechanic:
> This is great to see - and a really nice approach to overlaying and generating it.
thank you so much, I appreciate.

> Can you explain your use case a bit? 

In my use case, for the project I'm working on, I'd like the user to upload any geojson for a specific project (in supported EPSG, of course; the platform I'm working on could eventually convert them to the desired EPSG in the future), and these geojsons should be overlaid on a globe/basemap (photorealistic tiles, cesium globe, or whatever "base layer" the user wishes). One of the purposes of the geojson is to highlight, for example, the valley affected by a specific infrastructure (such as a bridge), so that the user can not only better see the area for whatever reason, but also understand, for example, whether a heavy rainfall (provided by a WMS) is within the area of ​​interest or not, perhaps causing a flood, and so on.

> Are you planning to have multiple polygons with multiple colors? One potential issue I see is that multiple colored geojson data sets will require separate layers of images which could cause a bit of an explosion of memory.

that would be possible, I didn't think about it honestly, we could have colors for each feature, similarly to what is possible to do with your geojson library for three.js.

about having multiple datasets, yes I understand that drawing on multiple canvas can be heavy if multiple data sets are there, in that case, I would consider, if possible, to have only one data set with multiple datasets, in order to create and draw using only one canvas per time, maybe having an Api from GeoJSONTilesOverlay/Plugin, to add a dataset and manage the "merge" of datasets behind the scene, or I would consider to test your suggestion about SVG.
if the SVG solution would work and is more performant, I'd keep that solution as the definitive, and even more simple maybe.

honestly I already made some tests with SVG but I have to dig more, I will consider that converter you linked me, or drawing them directly as I did in my last commit.

I think that the advantage you see in using SVG is to don't have to worry about levels and redrawing images, in order to gain performance, am I right? I agree with that eventually 🚀 , but with current implementation svgs are created whenever it would create other images format, eventually we should keep track of the memory and cpu usage to see what is the best approach, and how to optimize it.

gkjohnson:
> in supported EPSG, of course; the platform I'm working on could eventually convert them to the desired EPSG in the future

As far as I understand, GeoJSON does not support custom CRS. The values should always be in latitude / longitude which will be universal. To that end the "GeoJSONImageSource" shouldn't need a custom CRS other than to adjust the implicit tiling scheme. Though I think that's probably not needed for this.

> that would be possible, I didn't think about it honestly, we could have colors for each feature

For reference it looks like Mapbox provides a custom [styling system](https://docs.mapbox.com/style-spec/guides/) for things like this but it would be nice to avoid adding an overly complex styling system.

> I think that the advantage you see in using SVG is to don't have to worry about levels and redrawing images, in order to gain performance, am I right?

Unfortunately in order to render SVG in WebGL (or WebGPU afaik) you still have to rasterize the SVG data to a texture, so you will still have to deal with levels and the texture generation. But SVG is a good standard for these kinds of vector images and supports declarative styling out of the box rather than requiring a custom one. Once concern I have is that this will be slower, though, because of the complexity of the styles, but it will require some testing.

> Merge branch 'feature/geojson' of https://github.com/SoftwareMechanic/3DTilesRendererJS into feature/geojson

It looks like you've pulled and merged the cherry picked and rebased version of the branch I made into your local version, which has brought back all the styling changes I fixed. I can force push my version of the branch again but you'll need to pull it and overwrite your local version. Alternatively we can wait until #1302 is merged and you can fix up the styling issues in this branch then. The diff is otherwise too complex to understand now, unfortunately. Let me know what you'd like to do!

SoftwareMechanic:
> As far as I understand, GeoJSON does not support custom CRS. The values should always be in latitude / longitude which will be universal. To that end the "GeoJSONImageSource" shouldn't need a custom CRS other than to adjust the implicit tiling scheme. Though I think that's probably not needed for this.

Once it happened that I received a .geojson file in EPSG:3003, but that was in fact an exception,  and not even the Cesium Viewer can render it by default, so I agree to treat them as latitude longitude coordinates, maybe we can just write a line in the docs or in a comment in the example about this

> For reference it looks like Mapbox provides a custom [styling system](https://docs.mapbox.com/style-spec/guides/) for things like this but it would be nice to avoid adding an overly complex styling system.

I agree, I would not add another 'rendering' layer, not for now at least. the more is simple the better it is imo.

> Unfortunately in order to render SVG in WebGL (or WebGPU afaik) you still have to rasterize the SVG data to a texture, so you will still have to deal with levels and the texture generation. But SVG is a good standard for these kinds of vector images and supports declarative styling out of the box rather than requiring a custom one. Once concern I have is that this will be slower, though, because of the complexity of the styles, but it will require some testing.

that was in fact my doubt about webGL, but I agree that using SVG is better for drawing and styling, and can lead us to have a good and modular management of the styling eventually.


> It looks like you've pulled and merged the cherry picked and rebased version of the branch I made into your local version, which has brought back all the styling changes I fixed. I can force push my version of the branch again but you'll need to pull it and overwrite your local version. Alternatively we can wait until https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/1302 is merged and you can fix up the styling issues in this branch then. The diff is otherwise too complex to understand now, unfortunately. Let me know what you'd like to do!

🙈 I am sorry I messed up with GH desktop.
 you can forse push and I will overwrite my local version, my bad!

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,423 @@
+import * as THREE from 'three';
```
gkjohnson: nit: Can we use specific imports to grab imports instead of `*`, similar to the other files? Like so:

```js
import { CanvasTexture, /* ... */ } from 'three';
```

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,423 @@
...
+import { ProjectionScheme } from '../utils/ProjectionScheme.js';
+
+export class GeoJSONImageSource extends TiledImageSource {
+
+	constructor( {
+
+		geojson = null,
+		url = null, // URL or GeoJson object can be provided
+		tileDimension = 256,
+		bounds = [ - 180, - 90, 180, 90 ],
```
gkjohnson: Instead of passing "bounds" into the image source it would be best to calculate the bounds of the geojson  object so it's a tight bounds (including some margin to account for stroke thickness, etc.

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,423 @@
...
+		levels = 18,
+		projection = null, // 'EPSG:3857' or 'EPSG:4326' or null to auto- detect
```
gkjohnson: I think we can remove "projection" - the geojson spec specifies that [all values are in the WGS84 coordinate system](https://stevage.github.io/geojson-spec/#section-4). It sounds like it used to be the case that GeoJSON allowed for CRS specifications but I don't expect this to be common and I'd prefer to ask the user to convert their geojson to the common representation before passing it to the library for the sake of simplicity.

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,423 @@
...
+				// allow both Response and plain object (in case overlay fetch returns parsed JSON)
+				if ( res && typeof res.json === 'function' ) {
+
+					this.geojson = await res.json();
+
+				} else if ( res ) {
+
+					this.geojson = res;
+
+				}
```
gkjohnson: What case are you imagining where `fetchData` will return pre-parsed JSON? I think it should be okay to assume this is performing a typical fetch in this case.

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,423 @@
...
+
+			} catch ( e ) {
+
+				console.warn( 'GeoJSONImageSource: failed to fetch geojson from url', this.url, e );
+				// continue, but without geojson no tiles will be produced
+
+			}
```
gkjohnson: I think we should just let this init function throw an error so it can be handled elsewhere if needed. This is what the other image sources do.

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,423 @@
...
+
+		}
+
+		return Promise.resolve( );
```
gkjohnson: Since the function is marked as `asnyc` we shouldn't need to return a promise.

gkjohnson (commented):
Added some initial comments. I'll take a look at the drawing code later this week.

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,423 @@
...
+		const boundsNormalized = this.tiling.getTileBounds(
+			x,
+			y,
+			level,
+			true,
+			false,
+		); // [ minx,miny,maxx,maxy ] normalized
+
+		// compute tile extent in projection units ( degrees for 4326, meters for 3857 )
+		const tileInfo = this._tileExtentFromNormalized( boundsNormalized );
```
gkjohnson: You can get the lat / lon for the tile using the `getTileBounds` and passing `normalized = false`.

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,423 @@
...
+		canvas.width = this.tileDimension;
+		canvas.height = this.tileDimension;
+		const ctx = canvas.getContext( '2d' );
+
+		// clear
+		ctx.clearRect( 0, 0, canvas.width, canvas.height );
+
+		// find features that ( quick bbox ) intersect tile
+		const features =
+			this.geojson && this.geojson.features ? this.geojson.features : [ ];
```
gkjohnson: There's no guarantee that a geojson object has a root "features" object unless it's a "FeatureCollection". The GeoJSON root can otherwise be any of the GeoJSON geometry or feature types so we'll need to handle this more generically, as well.

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,423 @@
...
+
+		for ( let i = 0; i < features.length; i ++ ) {
+
+			const f = features[ i ];
+			if ( ! f || ! f.geometry ) continue;
+			if ( ! this._featureIntersectsTile( f, tileInfo ) ) continue;
```
gkjohnson: I wonder if we can cache the bounding boxes for each feature instead of recalculating it ever draw.

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,423 @@
...
+	// Compute pixel from projected coords and tileInfo
+	_projectedToPixel( pxProjX, pxProjY, tileInfo, width, height ) {
+
+		const { minX, minY, maxX, maxY } = tileInfo;
+		const rangeX = maxX - minX || 1;
+		const rangeY = maxY - minY || 1;
+		const x = ( ( pxProjX - minX ) / rangeX ) * width;
+		// canvas y origin is top, projection y increases north - > flip
+		const y = height - ( ( pxProjY - minY ) / rangeY ) * height;
+		return [ x, y ];
```
gkjohnson: This may be simplified by using the following:

```js
import { MathUtils } from 'three';

// ...

const px = MathUtils.mapLinear( lon, tileMinLon, tileMaxLon, 0, width );
const py = height - MathUtils.mapLinear( lat, tileMinLat, tileMaxLat, 0, height );
```

These will map the geojson lat / lon values from the tile ranges to the pixel coordinates (and flip the y value).

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: absolutely!

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: for the moment in this new commit I will set the bounds like for other image sources using the crs bounds, but if using the bounds limited to the entity can optimize the process, I would like to implement this, even in #1302 , a lot of WMS services like geojson use only a small portion of the map to display data in there.

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: I tried doing that, but logging the result I get different values from getTileBounds not normalized and from _tileExtentFromNormalized, and the drawing will not work anymore.

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: that could be possible, maybe creating an instance map, I will keep this conversation open for the moment, in order to come back to this optimization

src/three/plugins/images/sources/GeoJSONImageSource.js
gkjohnson: > if using the bounds limited to the entity can optimize the process, I would like to implement this

Calling `TilingScheme.setContentBounds` tells the tiling instance that any tiles that do not include that content bounds in its range is non existent and should not be loaded. This prevents unnecessary data loading and prevents render targets for overlays from being generated - saving memory and performance.

> even in #1302 , a lot of WMS services like geojson use only a small portion of the map to display data in there.

I'd added this in some of the WMSImageSource updates - the constructor takes a `contentBounds` field (which can be derived from the EX_GeographicBoundingBox in the WMS capabilities XML) and is used to set the content bounds of tiling [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/7fa6d531853f0b734b17f305540ee023e3cae1fa/src/three/plugins/images/sources/WMSImageSource.js#L45-L53) (with the full CRS range used as a fallback).

src/three/plugins/images/sources/GeoJSONImageSource.js
gkjohnson: Yeah I think it's okay to do this later, too.

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,662 @@
...
+			const tilesX = 2 ** i;
+			const tilesY = 2 ** i;
+			this.tiling.setLevel( i, {
+				tilePixelWidth: this.tileDimension,
+				tilePixelHeight: this.tileDimension,
+				tileCountX: tilesX,
+				tileCountY: tilesY,
+			} );
+
+		}
```
gkjohnson: We can just use the `generateLevels` function here, which will generate a basic set of quad-split tiles down to the desired level:

```js
tiling.generateLevels(
  levels,
  projection.tileCountX,
  projection.tileCountY,
  {
    tilePixelWidth: tileDimension,
    tilePixelHeight: tileDimension,
  },
);
```

src/three/plugins/images/sources/GeoJSONImageSource.js
gkjohnson: Lets start by setting all the CRSs to something that's easier to follow. Right now the "ProjectionScheme" is being set to `EPSG:3857`, which isn't right since the GeoJSON data just specifies WGS84 lat / lon values and these are just being linearly interpolated to convert to canvas coordinates. These interpretations of the data need to be the same.

EPSG:3857 is the "Web Mercator" CRS, which has a nonlinear projection distortion along the latitude axis. That means that while latitude and longitude will correspond to the same spots on a globe they will be at different points on the map - meaning the "normalized values" bounding the tile sections will be different `EPSG:3857` and `EPSG:4326`. It's possible to use the web mercator projection here but generally it's a bit easier to work with `EPSG:4326` and either way the important part is being consistent.

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: this is awesome!! thank you for confirm me this

I tried setting the bounds, but seems that if I don't use very large offset, the geojson won't be even rendered, in the last commit I added a method to get the bounds from geojson, but it's usage is commented out as it's not working as expected,  I would expect to don't even need to use an offset, but maybe I didn't get something about this

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: I found out that it works parsing the data but also without parsing it, you can try i.e. with this public geojson used in 'url' parameter: [italy municipalities](https://github.com/openpolis/geojson-italy/blob/master/geojson/limits_IT_municipalities.geojson)
(it's a bit heavy as geojson, but I just used the first one I found)

but I think that the best thing to do is to parse it anyway, what do you think?

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: this is done by the next commit, I absolutely agree with keeping consistency.

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: this should be fixed now, within '__drawFeatureOnCanvas_' method

gkjohnson: Can you elaborate? It doesn't look like this has been addressed. I still see the `this.geojson && this.geojson.features ? this.geojson.features : [ ];` condition in the in the `_drawFeatureOnCanvas` means that only "featureCollection" geojson will draw. This valid GeoJSON does not work, for example:

```json
{
  "type": "LineString",
  "coordinates": [
    [ 100, 0 ],
    [ 201, 1 ],
    [ 201, 10 ],
    [ 100, 10 ],
    [ 100, 0 ]
  ]
}
```

src/three/plugins/images/sources/GeoJSONImageSource.js
gkjohnson: I'm not sure if I completely follow. Calling `fetchData` is expected to return a `Request` object, not a normal Javascript Object that could be GeoJSON - so the expectation is that we will _always_ need to call "res.json()" here unless I'm missing something?

src/three/plugins/images/sources/GeoJSONImageSource.js
gkjohnson: It looks like there may actually be an issue with this that's appeared after some recent changes. I made #1320 to track it can can take care of it after this is merged.

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,759 @@
...
+	// main fetch per tile - > returns .Texture
+	async fetchItem( tokens, signal ) {
+
+		//return this.drawCanvasImage(tokens);
+		return await this.drawSVGImage( tokens, signal );
+
+	}
+
+
+	async drawSVGImage( tokens, signal ) {
```
gkjohnson: In the interest of getting this PR merged lets keep this PR focused on the canvas approach for rendering the GeoJSON. We can discuss an SVG approach in another issue but this code path is noticeably slower and I expect there are more complexities to it.

src/three/plugins/images/sources/GeoJSONImageSource.js
gkjohnson: It looks like this line can be switched to `tiling.getTileBounds( x, y, level, false, false );` now, right?

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,759 @@
...
+		if ( maxY === minY ) maxY = minY + 1;
+
+		const x = MathUtils.mapLinear( pxProjX, minX, maxX, 0, width );
+		// canvas y origin is top, projection y increases north -> flip
+		const y = height - MathUtils.mapLinear( pxProjY, minY, maxY, 0, height );
+
+		// clamp to finite numbers
+		return [ isFinite( x ) ? x : 0, isFinite( y ) ? y : 0 ];
+
+	}
```
gkjohnson: It may be worth rounding these returned values to integer values for the sake of performance. See [this MDN article on Canvas performance](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Optimizing_canvas#avoid_floating-point_coordinates_and_use_integers_instead) indicating that integer values are better to pass to the canvas functions.

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: now I got it, I am sorry I didn't understand what you mean (sometimes my english lacks), bear with me
that was a distraction error in the code, will be fixed!

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: for this commit I commented it out so we can have a reference eventually for the future, but if you prefer we can remove it, using a huge geojson to me seems to work slower with canvas, but I think after this PR we can think about optimization in general also. and I have not made real comparison yet

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: thank you for pointing that out

gkjohnson:
Thanks for the updates! I'll plan to make a few changes to simplify the code in some places. For now I think we can remove the SVG path, as well. We can always come back to this PR to get it in the history if needed.

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,386 @@
...
+				polygon.forEach( ring => ring.forEach( c => consume( c[ 0 ], c[ 1 ] ) ) ),
+			);
+
+		}
+
+		if ( minLon === Infinity ) {
+
+			return null;
+
+		}
```
gkjohnson: A few final questions: is this a case you have run into or is this just a defensive check?

src/three/plugins/images/sources/GeoJSONImageSource.js
```diff
@@ -0,0 +1,386 @@
...
+	}
+
+	// Compute pixel from projected coords and tileInfo
+	_projectedToPixel( pxProjX, pxProjY, bounds, width, height ) {
+
+		let [ minX, minY, maxX, maxY ] = bounds;
+
+		// avoid zero-range (prevents division by zero in mapLinear)
+		if ( maxX === minX ) maxX = minX + 1;
+		if ( maxY === minY ) maxY = minY + 1;
```
gkjohnson: Likewise - can you explain this case? In order for this case to be encountered the "tile bounds" would have to have a total width of "0" in either dimension, right? I don't think it's possible for that to happen?

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: I think this can be removed, it was a defensive check as you said

src/three/plugins/images/sources/GeoJSONImageSource.js
SoftwareMechanic: also this one was a defensive check, but it is probably never going to happen

gkjohnson:
Thanks! I'm going to merge this and move it into the `examples/src` folder, for now. ~I think there are still some open questions to be worked out regarding multiple geojson assets, styling, etc. Once some of these things are worked out we can move it in to core!~ I'll actually just leave this undocumented, for now, until some of the issues are worked out.

Thank you again for all the work on this.

SoftwareMechanic:
thanks to you for all the support and the help!

Issue #1316, opened 2025-09-10
Unable to import anything with TS


pailhead:
### Description

When trying to import from `3d-tiles-renderer` and `3d-tiles-renderer/plugins` TS complains:
```
Cannot find module '3d-tiles-renderer/plugins' or its corresponding type declarations.
```
Downgrading to `0.4.14` seems to fix the issue with `3d-tiles-renderer` but `/plugins` still fail. 

 

### Reproduction steps

1. `npm i 3d-tiles-renderer`
2. `import { GeoUtils } from '3d-tiles-renderer'`
3. Error with typescript. 

### Code

```js
// code goes here
```

### Library Version

v0.4.16

### Three.js Version

r178

dbuck:
Just came to post the same, I'm using typescript with moduleResolution: "node". I checked the "bundler" type also, as the error message shown when it can't find the type mentions it, but that's a larger undertaking on an existing project since many existing packages in the wild don't correctly set up their package.json.exports blocks yet.

Errors:
```
error TS2307: Cannot find module '3d-tiles-renderer/three' or its corresponding type declarations.
  There are types at '{REPO_ROOT}/node_modules/3d-tiles-renderer/src/three/renderer/index.d.ts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.

1 import { TilesRenderer } from '3d-tiles-renderer/three';
```

Workaround attempt: You can **partially** work around it by importing directly
```
import { TilesRenderer } from '3d-tiles-renderer/src/three/renderer/index.js';
```

<img width="718" height="61" alt="Image" src="https://github.com/user-attachments/assets/38ada461-1ffd-4dbd-ada4-ad62bdad73bf" />

but that doesn't cover it, my TilesRenderer is then presenting that the registerPlugin function doesn't exist since the typedef for TilesRenderer.d.ts uses `import { Tile, TilesRendererBase } from '3d-tiles-renderer/core';` and the typedef of `extends TilesRendererBase` doesn't propagate?

`export class TilesRenderer<TEventMap extends TilesRendererEventMap = TilesRendererEventMap> extends TilesRendererBase implements EventDispatcher<TEventMap> {`

dbuck:
introduced in version https://github.com/NASA-AMMOS/3DTilesRendererJS/releases/tag/v0.4.12 if you're looking for how far to fall back to avoid the issue @pailhead

[...]

Issue #1318, opened 2025-09-11
Typescript not properly resolving type definitions when build locally


xawill:
### Description

Opening an issue following https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/1261#issuecomment-3272568914.

I have a local fork of the library that I build with `npm run build-lib`. Before #1275 I was able to build the library and my project would correctly pickup the type definitions. However since then, it seems to take the .js files in the build folder instead (see attached screen capture with `TilesRenderer` example).

Note that when the build folder is deleted, the type definitions are fine again.

It is also fine if I install the repository from npm (v0.4.15 or v0.4.16).

Still reproducible in `master`.

It looks like the issue #1275 introduced is that in the index.d.ts now uses package import instead of relative paths. TypeScript now resolves it through Node module resolution using your package.json "exports" field, but does not automatically follow deep re-exports if the .d.ts is not at the root of the exported module path.

[With the help of chatgpt]: TypeScript sees '3d-tiles-renderer/core' → resolves to "exports" entry → loads the .d.ts → but this .d.ts is inside src/core/renderer, which is not the path TS expects based on how your exports are structured. So the chain is broken.

I am not exactly sure how we should fix it. In #1271 I had submitted a proposal to copy the types alongside the .js in the build folder. Maybe this could fix it, although I was not immediately able to make it work.

Simplest solution is probably to revert to relative paths in .d.ts. I personally don't immediately grasp the added value of the change, so maybe we can revert.

<img width="2208" height="718" alt="Image" src="https://github.com/user-attachments/assets/7f0a2dcf-3133-4077-9251-ab0de78e9ff7" />

### Reproduction steps

1. Build lib locally
2. Link it to your TS project: `npm link` in the library and then `npm link 3d-tiles-renderer` in the TSproject
3. Notice the type declaration are not properly being picked up.

### Code

NA

### Live example

NA

### Screenshots & Tile Set

NA

### Library Version

v0.4.16

### Three.js Version

NA

gkjohnson:
It sounds like this is a duplicate of #1316, is that right? Or is there something not covered in that issue?

xawill:
It probably is and the more they are talking about the issue the more it looks like that. But when I opened the issue it was not clear and the corresponding PR was closed for a reason I didn’t understand. 
[...]

PR #1319, opened 2025-09-12
fix: typescript detection should be fixed for ESM Node 16+

crutchcorn:
This PR updates the `"types"` and the `"import"` file order to avoid bugs in TS detection per https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1316

It even adds `publint` and `@arethetypeswrong/cli` to allow us to catch mistakes like this more easily in the future.

However, it does nothing to change CJS or Node10 file resolution. I can make a second PR to fix this, but it would be a larger change that would require a bundling modification

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "0.4.16",
       "license": "Apache-2.0",
       "devDependencies": {
-        "@arethetypeswrong/cli": "^0.18.2",
         "@babel/core": "^7.21.8",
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thanks! From #1318 it sounds like this fixes things. I've made some commits to adjust the indentation and removed some of the added dev dependencies. If these kinds of typing or package issues continue to be a problem we can revisit adding these kinds of checks.

Issue #1307, opened 2025-09-08
I3DMLoader: Add support for quantized position


mqrty:
### Description

I recently encountered a tileset containing I3DM with quantized positions. The `I3DMLoader` currently does not support this feature and fails [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/4b8495373758e5df5bc7ed00e82e10ee883c8852/src/three/renderer/loaders/I3DMLoader.js#L111) if the I3DM has no regular `POSITION`. It would be nice if such tilesets/I3DM can be loaded.

### Solution

This could be supported by applying the following formula in the `I3DMLoader` (see [spec](https://docs.ogc.org/cs/22-025r4/22-025r4.html#toc28)):

> POSITION = POSITION_QUANTIZED * QUANTIZED_VOLUME_SCALE / 65535.0 + QUANTIZED_VOLUME_OFFSET

The spec also mentions:
> If both POSITION and POSITION_QUANTIZED are defined for an instance, the higher precision POSITION will be used.

Perhaps [this tileset](https://github.com/CesiumGS/cesium/tree/2723c72c0169647c6ac9ee8c02c3d7151f02899f/Specs/Data/Cesium3DTiles/Instanced/InstancedQuantized) could be used for testing.

### Alternatives

Not being able to load such I3DM :(

gkjohnson:
Hello! Yes there are still a few I3DM features that are not supposed. Would you like to make a PR to add support? You can do so by constructing the "POSITION" buffer [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/4b8495373758e5df5bc7ed00e82e10ee883c8852/src/three/renderer/loaders/I3DMLoader.js#L105-L109) if the quantized buffers are present.

xawill:
Duplicate of #83. I was in the process of submitting a PR for this.

If @mqrty you need the feature urgently, I recommend you submit the PR yourself, otherwise I'll handle it in the next couple of weeks.

In any case, thank you for the tileset example for testing.

PR #1322, opened 2025-09-15
I3DMLoader: support quantized position

mqrty:
Hi, I went ahead and implemented support for I3DM with quantized positions.

I used [this tileset](https://raw.githubusercontent.com/CesiumGS/cesium/2723c72c0169647c6ac9ee8c02c3d7151f02899f/Specs/Data/Cesium3DTiles/Instanced/InstancedQuantized/tileset.json) for testing (as mentioned in #1307). Loading it with my changes looks like this:

image

### DIFF
--- a/src/three/renderer/loaders/I3DMLoader.js
+++ b/src/three/renderer/loaders/I3DMLoader.js
@@ -104,40 +104,15 @@ export class I3DMLoader extends I3DMLoaderBase {
 						} );
 
 						// use quantized position if position is missing
-						if ( ! POSITION ) {
+						if ( ! POSITION && POSITION_QUANTIZED ) {
 
-							if ( POSITION_QUANTIZED ) {
+							POSITION = new Float32Array( INSTANCES_LENGTH * 3 );
 
-								if ( ! QUANTIZED_VOLUME_OFFSET || ! QUANTIZED_VOLUME_SCALE ) {
+							for ( let i = 0; i < INSTANCES_LENGTH; i ++ ) {
 
-									[
-										'QUANTIZED_VOLUME_OFFSET',
-										'QUANTIZED_VOLUME_SCALE',
-									].forEach( feature => {
-
-										if ( ! ( feature in featureTable.header ) ) {
-
-											console.warn( `I3DMLoader: FeatureTable using "POSITION_QUANTIZED" but missing "${ feature }".` );
-
-										}
-
-									} );
-
-								}
-
-								POSITION = new Float32Array( INSTANCES_LENGTH * 3 );
-
-								for ( let i = 0; i < INSTANCES_LENGTH; i ++ ) {
-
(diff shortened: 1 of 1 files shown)


src/three/renderer/loaders/I3DMLoader.js
```diff
@@ -103,6 +103,46 @@ export class I3DMLoader extends I3DMLoaderBase {
...
+
+										if ( ! ( feature in featureTable.header ) ) {
+
+											console.warn( `I3DMLoader: FeatureTable using "POSITION_QUANTIZED" but missing "${ feature }".` );
+
+										}
+
+									} );
+
+								}
```
gkjohnson: Lets remove these checks for necessary. Generally I think we need to expect that the incoming data is well-formed otherwise we'll wind up checking for every possible mistake that an exporter could make.

src/three/renderer/loaders/I3DMLoader.js
```diff
@@ -103,6 +103,46 @@ export class I3DMLoader extends I3DMLoaderBase {
...
+									POSITION[ i * 3 + 1 ] = QUANTIZED_VOLUME_OFFSET[ 1 ] + ( POSITION_QUANTIZED[ i * 3 + 1 ] / 65535.0 ) * QUANTIZED_VOLUME_SCALE[ 1 ];
+									POSITION[ i * 3 + 2 ] = QUANTIZED_VOLUME_OFFSET[ 2 ] + ( POSITION_QUANTIZED[ i * 3 + 2 ] / 65535.0 ) * QUANTIZED_VOLUME_SCALE[ 2 ];
+
+								}
+
+							} else {
+
+								console.warn( 'I3DMLoader: FeatureTable missing "POSITION" and "POSITION_QUANTIZED".' );
+
+							}
```
gkjohnson: Same for here - this shouldn't be needed.

src/three/renderer/loaders/I3DMLoader.js
```diff
@@ -103,6 +103,46 @@ export class I3DMLoader extends I3DMLoaderBase {
 
 						} );
 
+						// use quantized position if position is missing
+						if ( ! POSITION ) {
+
+							if ( POSITION_QUANTIZED ) {
```
gkjohnson: I think we can simplify this to just check for `if ( POSITION_QUANTIZED )`

mqrty: Here I'm not sure. If we just check for `POSITION_QUANTIZED` the behaviour would be off-spec for I3DM that also have `POSITION` as it mentions to use the latter in such cases (see section 7.6.4.1.1 under [I3DM section](https://docs.ogc.org/cs/22-025r4/22-025r4.html#toc28)).

> If both POSITION and POSITION_QUANTIZED are defined for an instance, the higher precision POSITION will be used.

gkjohnson:
Looks great, thanks!

PR #1323, opened 2025-09-16
new example with Dutch WMTS sources

Ylannl:
The result of me trying to get local WMTS services working with that nice new WMTS plugin. This PR adds 

1. some small fixes to the WMTSCapabilitiesLoader:
   - setting the crs in case of a layer `WGS84BoundingBox`. Per the [WMTS spec](https://portal.ogc.org/files/?artifact_id=35326) (fig. 6) this can only be `urn:ogc:def:crs:CRS::84` in this case
   - handling the absence of optional elements in the layer `Style` capabilities (see same fig. in spec)
2. small enhancement to WMTSImageSource:
   - call `tiling.setContentBounds()` automatically using layer bounding box
3. a new example with several public Dutch WMTS services (available in EPSG:3857, some also in EPSG:4326)
   - based on the WMTSTiles example,
   - Notice: some WMTS sources require the camera to be close to the globe for tiles to show up, due to a lack of the lower Matrix levels.
   
Happy to split this up in several PRs if needed.


src/three/plugins/loaders/WMTSCapabilitiesLoader.js
```diff
@@ -199,15 +199,15 @@ function parseDimension( el ) {
...
+function parseBoundingBox( el, isWGS84 = false ) {
 
 	if ( ! el ) {
 
 		return null;
 
 	}
 
-	let crs = el.getAttribute( 'crs' );
+	let crs = isWGS84 ? 'urn:ogc:def:crs:CRS::84' : el.getAttribute( 'crs' );
```
gkjohnson: Would it be enough to default the crs to CRS84 rather than passing an argument in? Or checking the element tag name?

src/three/plugins/loaders/WMTSCapabilitiesLoader.js
```diff
@@ -244,9 +244,10 @@ function parseBoundingBox( el ) {
 // parse layer <Style> tag
 function parseStyle( el ) {
 
-	const title = el.querySelector( 'Title' ).textContent;
+	const title = el.querySelector( 'Title' )?.textContent || null;
 	const identifier = el.querySelector( 'Identifier' ).textContent;
-	const isDefault = el.getAttribute( 'isDefault' ) === 'true';
+	const isDefaultAttr = el.getAttribute( 'isDefault' );
+	const isDefault = isDefaultAttr ? isDefaultAttr === 'true' : null;
```
gkjohnson: Can you explain this change? "isDefault" should be a boolean value - "true" or "false". This change adjusts the result so it is either "true" or "null".

src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -81,7 +81,17 @@ export class WMTSImageSource extends TiledImageSource {
...
+
+		if ( layer.boundingBox !== null ) {
+
+			tiling.setContentBounds( ...layer.boundingBox.bounds );
+
+		} else {
+
+			tiling.setContentBounds( ...tiling.projection.getBounds() );
+
+		}
```
gkjohnson, line 93: For WMTS specifically this may not be necessary. Each "Tile Matrix" contains a dedicate bounding box describing the bounds of that layer of image data. The [schema](https://schemas.opengis.net/wmts/1.0/wmtsGetCapabilities_response.xsd), in fact, doesn't even permit a bounding box definition in the <Layer> element so it's possible we should remove the parsing logic for layer bounding boxes.

src/three/plugins/loaders/WMTSCapabilitiesLoader.js
Ylannl: As far as I understand this evaluates to false if there is value and it is not equal to "true", as it was before. Only now it becomes "null", if there is no `isDefault` element, which can happen since it is not required by the WMTS spec. This gives the user just a little bit more information as to if that element was present or not (compared to just setting it to false). 
That was my rationale, but happy to simplify it to always be a boolean if you prefer that.

src/three/plugins/loaders/WMTSCapabilitiesLoader.js
Ylannl: Yes, of course. I implemented the second option. Technically a BoundingBox without the WGS84 prefix could lack a crs element, in which case it would be incorrect to assume a default. Although this could still be the more pragmatic thing to do, but I don't have enough experience with WMTS to know.

src/three/plugins/images/sources/WMTSImageSource.js
Ylannl: If everything can be derived from the tile matrix sets, I agree it would be better to ignore the Layer BoundingBox completely, and just look at the tile matrices. 

Weirdly enough some of the WMTS services (eg, "PDOK Luchtfoto") that I was looking at serve tiles (many completely white or transparent) for the entire globe, even though the actual data only spans one country. This is why I did this change. I'm assuming the implementation already takes into account the extent of the tile matrix when requesting tiles?
In such cases this Layer BoundingBox can still be helpful to pinpoint where the data is (eg for automatically centering the camera).

Regarding if it is part of the schema or not: as far as I understand it is an optional element that can be used if the WMTS services also adheres to the OWS standard. But I'm no expert. I do get the impression that in practice it is quite common to have a Layer BoundingBox.

Ylannl:
Thanks for taking the time to look at my changes! I answered your comments. 

I can remove the example, and port the ergonomy changes to the existing WMTS example. I had not noticed you can pass a url like that.
However, one issue is that some WMTS services only work on very small extents and/or on very high 'zoom' levels. So you might not see anything when loading a url. And it would be virtually impossible to pan and zoom to the data without any map reference.
Do you think we can use the BoundingBox from the Layer in combination with the lowest available tile matrix level to automatically set the camera to where the data is visible?

src/three/plugins/loaders/WMTSCapabilitiesLoader.js
gkjohnson: The default value of the "isDefault" attribute should be "false" so I think we should reflect that in the parsed data. The parser should be disambiguating and filling in these default values otherwise the end-user (or end-plugin) will have to make assumptions about what the implicit value should be when this is "null".

gkjohnson:
> Do you think we can use the BoundingBox from the Layer in combination with the lowest available tile matrix level to automatically set the camera to where the data is visible?

Yeah I think automatically focusing the camera on the region with data is a good improvement to make, as well!

Ylannl:
Alright, I fixed everything, removed my example, and made a few improvements to the WMTS example:

- don't crash when no default layer is set
- auto select a layer with a compatible CRS on initialisation
- auto select a TileSetMatrix with a compatible CRS on GUI rebuild

Automatically focusing the camera is not implemented yet, I can do this in another PR.

src/three/plugins/loaders/WMTSCapabilitiesLoader.js
```diff
@@ -244,9 +244,9 @@ function parseBoundingBox( el ) {
 // parse layer <Style> tag
 function parseStyle( el ) {
 
-	const title = el.querySelector( 'Title' ).textContent;
+	const title = el.querySelector( 'Title' )?.textContent || null;
 	const identifier = el.querySelector( 'Identifier' ).textContent;
-	const isDefault = el.getAttribute( 'isDefault' ) === 'true';
+	const isDefault = el.hasAttribute( 'isDefault' ) ? el.getAttribute( 'isDefault' ) === 'true' : false;
```
gkjohnson, line 249: nit: this line shouldn't have to change, `el.getAttribute( 'isDefault' ) === 'true'` will always return the correct "true" or "false" value.

example/wmtsTiles.js
```diff
@@ -8,6 +8,7 @@ import { TilesFadePlugin, UpdateOnChangePlugin, WMTSCapabilitiesLoader, WMTSTile
 import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
 
 const url = window.location.hash.replace( /^#/, '' ) || 'https://gibs.earthdata.nasa.gov/wmts/epsg4326/best/wmts.cgi?SERVICE=WMTS&request=GetCapabilities';
+const compatibleCRSList = [ 'EPSG:4326', 'EPSG:3857', 'OGC:1.0:GoogleMapsCompatible', 'urn:ogc:def:crs:CRS::84' ];
```
gkjohnson, line 11: This is a good idea. I believe the WMTSCapabilitiesLoader should be implicitly converting urn:ogc:def:crs:CRS::84 to EPSG:4326 in the interest of consistency within the tiles renderer classes. If "GoogleMapsCompatible" is a real CRS definition then we can probably convert that to EPSG:3857, as well, to help simplify some of this CRS management. We should just be able to check for this in the [isWebMercator](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/8ecd0bd583d60aaeededcafde7d26591ae4680e6/src/three/plugins/loaders/WMTSCapabilitiesLoader.js#L21-L25) function.

src/three/plugins/loaders/WMTSCapabilitiesLoader.js
Ylannl: Apologies, I am still learning javascript. Thanks for being patient with this line!

example/wmtsTiles.js
Ylannl: indeed, I hadn't realised that. Simplified this to check only the two EPSG codes. Upon closer inspection the 'GoogleMapsCompatible' is not actually a CRS value, I got confused by the TileSetMatrix naming of one WMTS service.

gkjohnson:
Thank you!

PR #1326, opened 2025-09-19
GeoJSONOverlay: Add support for drawing polygons with holes

gkjohnson:
Based on #1325 

Uses "evenodd" fill for drawing polygon holes.

**Before**
image

**After**
image

cc @SoftwareMechanic


SoftwareMechanic:
wow thank you for all these development!  I  do really appreciate!!

PR #1335, opened 2025-10-06
Remove conditional import for TextureUtils

Desplandis:
The conditional import of the `TextureUtils` as
```js
const { TextureUtils } = THREE;
if ( ! TextureUtils ) {

  return 0;

}
```
causes bundler to disabling entirely tree-shaking (i.e. dead code removal) on `three.js`.

Since `TextureUtils` is available since r166 and `3d-tiles-renderer` depends on `three` > r170, I think it is safe to use the standard ESM import.

### DIFF
--- a/src/three/plugins/images/sources/TiledImageSource.js
+++ b/src/three/plugins/images/sources/TiledImageSource.js
@@ -1,7 +1,6 @@
 import { DataCache } from '../utils/DataCache.js';
 import { TilingScheme } from '../utils/TilingScheme.js';
-import { SRGBColorSpace, Texture } from 'three';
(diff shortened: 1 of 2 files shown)


gkjohnson:
Thanks! 

> 3d-tiles-renderer depends on three > r170

Just a small note that the project reports as [being compatible with r166 and up](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/package.json#L115) - it's just that 170 is used as the development dependency. Either way this looks safe, though!

Desplandis:
My bad, I hadn't looked to peer dependencies !
Thanks for merging it =)

PR #1336, opened 2025-10-07
Added custom fetchOptions management for overlays, added a proxyUrl parameter

SoftwareMechanic:
The reason for the proxyUrl is that usually the front end can not directly communicate with external servers, causing CORS issues.
so a proxy is needed in order to manage the request server side, regardless of the need for custom fetch options.

the proxy is also used by the CesiumJS Library, here just a code example:

```
        const imageryProvider = new CesiumJs.WebMapServiceImageryProvider({
          url: new CesiumJs.Resource({
            url: wmsUrl,
            headers,
            proxy: new CesiumJs.DefaultProxy("/api/proxy?url="),
          }),
          layers: layerName,
          enablePickFeatures: true,
          parameters: {
            transparent: true,
            format: "image/png",
            version: "1.3.0",
          },
        });
```

for the custom headers, the ImageOverlay class now takes also fetchOptions parameter and use them while fetching the imageSource.

just 2 things:

1. if you want, I can send you privately via e-mail an example which uses the authentication (and a proxy that you should create separately). I can give you the credentials, but of course this would be just to let you test it eventually.
2. I noticed something that was happening also while I was trying to implement WMS initially, when using EPSG:4326, (and not 3857) the layer is not correctly applied to the map, with my example with authentication, you can clearly see this. we could solve it for this PR, or otherwise i could create a new Issue for that. as you prefer :)

just for reference here an image of what I mean in the point 2.
immagine

if you notice something, don't hesitate to tell me!

### DIFF
--- a/src/three/plugins/images/ImageOverlayPlugin.d.ts
+++ b/src/three/plugins/images/ImageOverlayPlugin.d.ts
@@ -20,6 +20,8 @@ export class ImageOverlay {
 	color: number | Color;
 	opacity: number;
 	frame?: Matrix4 | null;
+	proxyUrl?: string;
+	fetchOptions?: any;
 
 }
 
@@ -71,7 +73,6 @@ export class WMSTilesOverlay extends ImageOverlay {
 		levels?: number,
 		transparent?: boolean,
 		contentBoundingBox?: [ number, number, number, number ],
-
 		color: number | Color,
 		opacity: number,
 		frame?: Matrix4 | null,
--- a/src/three/plugins/images/ImageOverlayPlugin.js
+++ b/src/three/plugins/images/ImageOverlayPlugin.js
@@ -961,7 +961,9 @@ export class ImageOverlayPlugin {
 	_initOverlay( overlay ) {
 
 		const { tiles } = this;
-		overlay.imageSource.fetchOptions = tiles.fetchOptions;
+		//overlay.imageSource.fetchOptions = tiles.fetchOptions;
+		overlay.imageSource.fetchOptions = { ...tiles.fetchOptions, ...overlay.fetchOptions };
+
 		if ( ! overlay.isInitialized ) {
 
 			overlay.imageSource.fetchData = ( ...args ) => tiles
@@ -1407,8 +1409,12 @@ class ImageOverlay {
 			opacity = 1,
 			color = 0xffffff,
 			frame = null,
+			fetchOptions = {},
+			proxyUrl = null,
 		} = options;
 		this.imageSource = null;
+		this.fetchOptions = fetchOptions;
+		this.proxyUrl = proxyUrl;
 		this.opacity = opacity;
 		this.color = new Color( color );
 		this.frame = frame !== null ? frame.clone() : null;
@@ -1428,9 +1434,12 @@ class ImageOverlay {
 
 	}
 
-	fetch( ...args ) {
+	// Apply proxy prefix and merge options
+	fetch( url, options = {} ) {
 
-		return fetch( ...args );
+		const finalUrl = this.proxyUrl ? `${ this.proxyUrl }${ encodeURIComponent( url ) }` : url;
+		const finalOptions = { ...this.fetchOptions, ...options };
+		return fetch( finalUrl, finalOptions );
 
 	}
 
--- a/src/three/plugins/images/sources/WMSImageSource.js
+++ b/src/three/plugins/images/sources/WMSImageSource.js
@@ -140,7 +140,6 @@ export class WMSImageSource extends TiledImageSource {
 			REQUEST: 'GetMap',
 			VERSION: version,
 			LAYERS: layer,
-			STYLES: styles,
 			[ crsParam ]: crs,
 			BBOX: bboxParam.join( ',' ),
 			WIDTH: tileDimension,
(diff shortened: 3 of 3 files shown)


src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1428,9 +1434,12 @@ class ImageOverlay {
 
 	}
 
-	fetch( ...args ) {
+	// Apply proxy prefix and merge options
+	fetch( url, options = {} ) {
 
-		return fetch( ...args );
+		const finalUrl = this.proxyUrl ? `${ this.proxyUrl }${ encodeURIComponent( url ) }` : url;
```
gkjohnson: Instead of "proxyURL" lets add an optional "preprocessURL" callback function similar to the plugins that can transform the url. Then users can format the url however they need. Something like so:

```js
if ( this.preprocessURL ) {

  url = this.preprocessURL( url );

}
```

SoftwareMechanic: managed this! for the capabilities loader, do you want a similar managemement? in my private example with credentials I am using the proxy url + the endpoint url of the wms capabilities in order to get the capabilities. but if you prefer I could add a preprocess url callback also for the capabilities loader

gkjohnson: If there's no issue then I'd prefer not to add it since the user can construct it themselves before using the loader which otherwise isn't the case here. Is that right?

SoftwareMechanic: I think yes, it's fine 🚀

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1428,9 +1433,17 @@ class ImageOverlay {
...
+	fetch( url, options = {} ) {
+
+		if ( this.preprocessURL ) {
+
+			url = this.preprocessURL( url );
+
+		}
 
-		return fetch( ...args );
+		const finalOptions = { ...this.fetchOptions, ...options };
```
gkjohnson: `this.fetchOptions` and the `options` passed here are the same ones since `imageSource.fetchOptions` is assigned  above, right? I think it's okay to just use `options` here.

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1407,8 +1408,12 @@ class ImageOverlay {
 			opacity = 1,
 			color = 0xffffff,
 			frame = null,
+			fetchOptions = {},
+			preprocessURL = null,
 		} = options;
 		this.imageSource = null;
+		this.fetchOptions = fetchOptions;
```
gkjohnson: Since the options on `imageSource` are really the ones being used here it may make most sense to add getters / setters for those options instead of storing them multiple places:

```js
get fetchOptions() { return this.imageSource.fetchOptions; }
set fetchOptions( v ) { this.imageSource.fetchOptions = v; }
```

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -961,7 +961,8 @@ export class ImageOverlayPlugin {
 	_initOverlay( overlay ) {
 
 		const { tiles } = this;
-		overlay.imageSource.fetchOptions = tiles.fetchOptions;
+		overlay.imageSource.fetchOptions = overlay.fetchOptions;
+
```
gkjohnson, line 964: This line can just be removed, now.

src/three/plugins/images/ImageOverlayPlugin.d.ts
```diff
@@ -21,6 +21,8 @@ export class ImageOverlay {
 	color: number | Color;
 	opacity: number;
 	frame?: Matrix4 | null;
+	proxyUrl?: string;
+	fetchOptions?: any;
```
gkjohnson, line 25: Lets update the types to reflect the new preprocessURL function.

src/three/plugins/images/ImageOverlayPlugin.js
SoftwareMechanic: you are right, I checked and yes I was merging the same headers basically, using _options_ is fine 👍

src/three/plugins/images/ImageOverlayPlugin.js
SoftwareMechanic: Agree, but in order to make _fetchOptions_ still usable as a parameter in the constructor from the client (and so for usage in R3F) I had to set the imageSource fetchOptions in the **WMSTilesOverlay** using the options, otherwise I can't set from ImageOverlay the fetch Options before the imageSource is defined.

```
export class WMSTilesOverlay extends ImageOverlay {

	constructor( options = {} ) {

		super( options );
		
		this.imageSource = new WMSImageSource( options );

		if (options.fetchOptions) {

			this.imageSource.fetchOptions = options.fetchOptions;

		}

		this.imageSource.fetchData = ( ...args ) => this.fetch( ...args );

	}
... rest of the code


```
if this the way to go I would consider implementing the same logic for all other overlays

SoftwareMechanic:
sorry to keep you waiting, I was sick these days.

I just left one conversation open which I think is good point to discuss.

src/three/plugins/images/ImageOverlayPlugin.js
gkjohnson: Sorry for the delay - I've just made a few changes that allow for "fetchOptions" to be passed down into the TiledImageSource as an option parameter so this should "just work" across all image overlays, now. I think this should be ready, now, but let me know how it looks!

SoftwareMechanic: I tried using my wms service with authentication and works very fine also with these modifications 😃 

I think too now this is fine.

eventually, if I face errors, I will open new issues/pull requests.

PR #1357, opened 2025-11-12
GLTFExtensionsPlugin: update typing

sguimmara:


src/three/plugins/GLTFExtensionsPlugin.d.ts
```diff
@@ -1,7 +1,10 @@
 import { DRACOLoader } from 'three/examples/jsm/loaders/DRACOLoader.js';
 import { KTX2Loader } from 'three/examples/jsm/loaders/KTX2Loader.js';
+import { MeshoptDecoder as MeshoptDecoderObj } from 'three/examples/jsm/libs/meshopt_decoder.module.js';
 import { GLTFLoaderPlugin, GLTFParser } from 'three/examples/jsm/loaders/GLTFLoader.js';
 
+export type MeshoptDecoder = typeof MeshoptDecoderObj;
+
```
gkjohnson: Why is necessary to re-export this type? The user should be getting the type from the original type definitions the same way this file is, no?

sguimmara: Updated without the re-export

gkjohnson:
Great, thanks!

PR #1361, opened 2025-11-13
Fix unregisterPlugin not working with name overload

dsafa:
When calling `TilesRendererBase.unregisterPlugin` with a name, it doesn't remove the plugin because it is using the wrong variable


gkjohnson:
Thanks! I'm surprised the linter didn't catch the non existent variable.

dsafa:
Its unfortunately because of the [global name](https://developer.mozilla.org/en-US/docs/Web/API/Window/name) property. I've had this happen a few times myself 😅

Issue #746, opened 2024-09-15
Globe Tiles: Hide tiles based on facing direction


gkjohnson:
If traversing tiles with known lat / lon values we can mark them as non traversable / non visible based on camera facing direction.

gkjohnson:
Tiles don't seem to have a large enough "non-renderable" view cone for this to be worthwhile.

PR #1368, opened 2025-11-17
Added a plugin for clipping the globe geometry starting from geojson geometries

### LINKED ISSUE #1365: Hide Regions of 3D Tiles or Terrain with Clipping Polygons
### Description

Hello!

I was looking at this interesting [feature ](https://cesium.com/learn/cesiumjs-learn/cesiumjs-clipping-polygons/)in CesiumJS, and I think we need it also in this library 😄 

basically we need a way to cut some part of the tiles out, in order to highlight our assets, and avoid artefacts from the tiles, when a model is aligned. especially with photorealistic ones.

### Solution

I think somehow this is doable, but I don't know where to start for it's implementation.
do you think we should use a CSG library, or is there another way?

### Alternatives

Don't know

### Additional context

in the link, you can also see they start from the geojson, which now this library supports, hopefully that can help us.

eventually I would like to help for this

SoftwareMechanic:
Fix #1365

this could be useful to represent assets on tiles, especially photorealistic tiles, without using the TileFlatteningPlugin.
the purpose is to render our models in their original position, removing artifacts given by the photorealistic tiles.

this pull request is aimed at solving (partially) the issue #1365 

I say **partially** because as we have discussed in the issue, there could be other ways, with other pros/cons to achieve the same effect.

questions:

1. do we want it to extend the geojson class as it is now in this initial state of the pr, or a new class only for it?
2. is it ok to modify the _wrapOverlaysMaterials_ like has beeen done, or we somehow want to use the alphamap default property from the materials?

### DIFF
--- a/src/three/plugins/images/ImageOverlayPlugin.d.ts
+++ b/src/three/plugins/images/ImageOverlayPlugin.d.ts
@@ -57,33 +57,15 @@ export class GeoJSONOverlay extends ImageOverlay {
 
 		frame?: Matrix4 | null,
 
-	} );
-
-}
-
-export class GeoJSONClipOverlay extends GeoJSONOverlay {
-
-	constructor( options: {
-
-		isClipOverlay : boolean,
-		clipThreshold : number,
-		clipInside : boolean,
-
-		// rasterize GeoJSON per tile (forwarded to GeoJSONImageSource)
-		geojson?: any, // FeatureCollection or null (if url provided)
-		url?: string, // optional URL alternative to geojson object
-		tileDimension?: number, // tile size in px (runtime name: tileDimension)
-		levels?: number, // max rasterization zoom
-		pointRadius?: number,
-		strokeStyle?: string,
-		fillStyle?: string,
-		color?: number | Color,
-		opacity?: number,
-		frame?: Matrix4 | null,
+		// clip options
+		alphaMode?: 'layer' | 'global',
+		alphaTest?: number, // 0..1, 0 = no hard clip
+		alphaInvert?: boolean, // false = cut inside (keep outside); true = cut outside (keep inside)
 
 	} );
 
 }
+
 
 export class WMSTilesOverlay extends ImageOverlay {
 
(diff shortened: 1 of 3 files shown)


gkjohnson:
Thanks! I think for the API we can keep things more simple. Something like so:

```js
new GeoJSONOverlay( {
  // ...
  alphaTest: 0.5, // or [ 0 to 1 ] (default 0)
  alphaMode: 'global', // or 'layer' (default)
} );
```

"alphaTest" indicates the usage of alpha clipping and the associated alpha value to perform a hard clip at, similar to the material alpha test (eg "0" indicates normal transparency). "alphaMode" would indicate whether the alpha is used to blend to the layer below it (current behavior) or multiply the opacity by the accumulated alpha for the fragment (open to another name for this, as well). This should the change a lot more simple.

Let me know if you need any clarification or would like me to make any of those changes. This looks great, though.

_edit_

> this pull request is aimed at solving (partially) the issue #746

This issue is actually a bit outdated and unrelated to this issue. This was about trying to limit the downloaded tiles based on whether the camera could see the geometry (checking view direction to normals) in order to limit tile loads but it turns out it doesn't save a whole lot and most tiles don't have a large enough (or any) "non-renderable" angle for this to make sense.

SoftwareMechanic:
I tagged the wrong issue 🤦 I modified the message.

I also managed an "alphaInvert" parameter, to let the user decide what to occlude - the GeoJson Geometries, or everything except it.

if there is something which is not as you expected, don't hesitate to correct me

gkjohnson:
Thanks - I think I'm seeing a more simple way to handle this, now, but I think it's difficult to communicate the strategy. They might be significant but I'll make some changes to this PR when I get a chance and we can roll back if needed

SoftwareMechanic:
It is fine for me, I will keep an eye on this PR 💯

gkjohnson:
Closing in favor of #1372 - thanks for getting this started 🙏

PR #1372, opened 2025-11-20
Add "alphaMask" option for ImageOverlays

### LINKED ISSUE #1365: Hide Regions of 3D Tiles or Terrain with Clipping Polygons

gkjohnson:
Related to #1368 
Fix #1365

- Adds "alphaMask" option for "ImageOverlays". When "true" the color of the layer is ignored and alpha is applied globally to all layers below this layer.
- Adds "alphaTest" option for "ImageOverlays", which works similarly to three.js' "Material.alphaTest" option. When "0" no adjustment to alpha is done. When "true" alpha is clipped to [ 0, 1 ] with antialiasing for a clean mask an compatibility with alpha to coverage.
- Adds "alphaInvert" option which inverts the alpha of the given layer.

**Note**
- In order for the alpha to apply correctly the base materials have to be pre-configured to support alpha, in other words "transparent" or "alphaToCoverage" must be set to "true" by the user.
- In order to render both "color" geojson (or other) overlays in addition to an alpha mask you will have to use two different layers.

**TODO**
- Convert the new options to #defines and javascript-expanded loop to avoid additional if-statements.
- Consider approach for using the "red" channel as alpha so mask overlays can use a smaller texture size.

image
image

cc @SoftwareMechanic - I decided it was easier to make a separate PR, but I think this covers everything and should be a simpler change. Let me know what you think and if it looks good I'll handle the TODO items and get this merged.

### DIFF
--- a/src/three/plugins/images/ImageOverlayPlugin.js
+++ b/src/three/plugins/images/ImageOverlayPlugin.js
@@ -1325,11 +1325,6 @@ export class ImageOverlayPlugin {
 				params.layerMaps.value[ i ] = target !== null ? target.texture : null;
 				params.layerInfo.value[ i ] = overlay;
 
-				// mark per-layer defines
-				material.defines[ `LAYER_${ i }_EXISTS` ] = Number( target !== null );
-				material.defines[ `LAYER_${ i }_ALPHA_INVERT` ] = Number( overlay.alphaInvert );
-				material.defines[ `LAYER_${ i }_ALPHA_MASK` ] = Number( overlay.alphaMask );
-
 				material.defines.LAYER_COUNT = overlays.length;
 				material.needsUpdate = true;
 
--- a/src/three/plugins/images/overlays/wrapOverlaysMaterial.js
+++ b/src/three/plugins/images/overlays/wrapOverlaysMaterial.js
@@ -77,6 +77,9 @@ export function wrapOverlaysMaterial( material, previousOnBeforeCompile ) {
 					struct LayerInfo {
 						vec3 color;
 						float opacity;
+
+						int alphaMask;
+						int alphaInvert;
 					};
 
 					uniform sampler2D layerMaps[ LAYER_COUNT ];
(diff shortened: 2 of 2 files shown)


SoftwareMechanic:
thank you a lot for pointing me to this new solution.

I want to ask you, what approach you'd suggest for this:  

> In order for the alpha to apply correctly the base materials have to be pre-configured to support alpha, in other words "transparent" or "alphaToCoverage" must be set to "true" by the user.



thank you in advance,
your solution seems way more clean 🚀 

to me this pull request is fine ✌️

gkjohnson:
> In order for the alpha to apply correctly the base materials have to be pre-configured to support alpha, in other words "transparent" or "alphaToCoverage" must be set to "true" by the user.

You can adjust the materials for loaded geometry using the "load-model" event:

```jsx
<TilesRender /* ... */ onLoadModel={ ( { scene } ) => {
  scene.traverse( child => {
    if ( child.material ) {

      // apply settings here

    }
  } );
} } />
```

- Setting `transparent=true` and `depthWrite=false` will give you a typical transparency effect but have artifacts due to no depth write.
- Setting `alphaToCoverage=true`, `transparent=true`, and `depthWrite=true` will allow for antialised, crisp edges at transparency boundaries but doesn't allow for partial transparency and only works with AA render targets.

I'm going to have to think through "alphaTest" a bit more, I think. It may be best to rely on the original material "alphaTest" value here rather than providing a value per layer, in which case you'd set "alphaTest" in the above loop, as well.

gkjohnson:
Okay I've converted everything to use `#defines`. The last thing to do is adjust the overlays so the defines update when the flags are modified.

gkjohnson:
After looking around a bit it seems like the consensus is that modern gpus aren't impacted by if statements as significantly and even then there is really no dynamic branching since the pixels are all evaluating the same uniform. The if statements are much more simple so I'll plan to keep them here.

PR #1381, opened 2025-11-26
Fix access of `getPointerCount` by invocating instead of accessing `length`

Philipp-M:
Something I noticed which I don't believe was right.


gkjohnson:
Hello! Yes it certainly looks like a refactoring mistake! Thanks!

PR #1385, opened 2025-11-27
Adds support for external imagery providers in CesiumIonOverlay

### LINKED ISSUE #1382: CesiumIonOverlay with different imagery providers
For two months, Google Maps 2D tiles have also been available in Cesium by default next to Bing maps (https://cesium.com/blog/2025/10/02/introducing-google-maps-2d-tiles/). When I wanted to use the `CesiumIonOverlay` with the Bing maps imagery, I always got an error, but didn't investigate it further. Now, since I wanted to test the Google Maps 2D tiles, I tested it again, but the error remains.

Image

Now, I had a closer look. I'm basically try to use the [Quantized Mesh with Overlays](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/quantMeshOverlays.html) example just with another overlay.

```ts
baseOverlay = new CesiumIonOverlay( {
  assetId: '3830184',
  apiToken: import.meta.env.VITE_ION_KEY,
} );
```

However, the error is thrown within `TMSImageSource.init`. But main issue is in  `CesiumIonOverlay.init` with following line:

```ts
this.imageSource.url = json.url;
```

Looking at the endpoint requests, following data is returned:

```json
{
    "type": "IMAGERY",
    "externalType": "BING",
    "options": {
        "url": "https://dev.virtualearth.net",
        "mapStyle": "Aerial",
        "key": "cesium_provided_access_key"
    },
    "attributions": [ ... ]
}
```

```json
{
    "type": "IMAGERY",
    "externalType": "GOOGLE_2D_MAPS",
    "options": {
        "url": "https://assets.ion.cesium.com/proxy/3830184",
        "imageFormat": "png",
        "tileWidth": 256,
        "tileHeight": 256,
        "session": "cesium_provided_session_key",
        "key": "cesium_provided_access_key"
    },
    "attributions": [ ... ]
}
```

So, `json.url` is obviously not defined. But also `this.imageSource.url = json.url ?? json.options?.url;` or similar doesn't help either. Actually, each provider would require a special handling: these requests need to be sent with the provided access and session keys, url format may be different, etc. For Google Maps, the `CesiumIonOverlay` would probably need to use the `GoogleCloudAuth` class (maybe not needed if session key is already provided). Instead of `TMSImageSource`, it would need `XYZImageSource`. And so on.

So, I'm not sure if this is possible to support these different imagery providers via Cesium. I can also make a PR, if you want.

Brakebein:
Extends CesiumIonOverlay to support external imagery providers like Google Maps and Bing Maps, fetching metadata and adjusting tile URL structures accordingly.

Updates the example to showcase this new feature by adding base map options for different map providers.

Closes #1382


src/three/plugins/images/sources/XYZImageSource.js
```diff
@@ -49,3 +49,57 @@ export class XYZImageSource extends TiledImageSource {
...
+
+		return quadKey;
+
+	}
+
+	// Get random subdomain
+	_getSubdomain() {
+
+		return this.subdomains[ Math.floor( Math.random() * this.subdomains.length ) ];
+
```
gkjohnson: Can you explain this? What's the purpose? You're randoming fetching from a set of urls every time?

src/three/plugins/images/sources/XYZImageSource.js
```diff
@@ -49,3 +49,57 @@ export class XYZImageSource extends TiledImageSource {
...
+		return this.url
+			.replace( /{\s*subdomain\s*}/gi, this._getSubdomain() )
+			.replace( /{\s*quadkey\s*}/gi, this._tileToQuadKey( x, y, level ) );
+
+	}
+
+	_tileToQuadKey( x, y, level ) {
+
+		// zoom level 0 does not exist for QuadKey
+		if ( level === 0 ) return '0';
```
gkjohnson: What does "not exist" mean here? This will still result in fetching a tile, correct?

src/three/plugins/images/sources/XYZImageSource.js
```diff
@@ -49,3 +49,57 @@ export class XYZImageSource extends TiledImageSource {
 	}
 
 }
+
+// Bing Maps Tile System
+// https://learn.microsoft.com/en-us/bingmaps/articles/bing-maps-tile-system
```
gkjohnson: From this link - it looks like the Bing maps are going to be retired in a few years. Are there any other tiling schemes that use this quad key system?

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1612,30 +1617,72 @@ export class CesiumIonOverlay extends ImageOverlay {
...
+					default:
+						this.imageSource = new TMSImageSource( {
+							...this.options,
+							url: json.url,
+						} );
+						this.imageSource.fetchData = ( ...args ) => this.fetch( ...args );
+
+				}
+
+				// this.imageSource.url = json.url;
```
gkjohnson: nit: lets remove this comment if it's not needed

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1612,30 +1617,72 @@ export class CesiumIonOverlay extends ImageOverlay {
...
+			.then( async ( json ) => {
 
 				this._attributions = json.attributions.map( att => ( {
 					value: att.html,
 					type: 'html',
 					collapsible: att.collapsible,
 				} ) );
 
-				this.imageSource.url = json.url;
+				if ( json.type !== 'IMAGERY' ) throw new Error( 'Only IMAGERY is supported as overlay type' );
```
gkjohnson: nit: Lets prepend this with with the name of the plugin so users know where the error was thrown from:

```js
throw new Error( 'CesiumIonOverlay: Only IMAGERY is supported as overlay type' );
```

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1612,30 +1617,72 @@ export class CesiumIonOverlay extends ImageOverlay {
...
+						} );
+						break;
+
+					case 'BING': {
+
+						const response = await fetch(
+							`${json.options.url}/REST/v1/Imagery/Metadata/${json.options.mapStyle}?incl=ImageryProviders&key=${json.options.key}&uriScheme=https`,
+						).then( ( res ) => res.json() );
+						const metadata = response.resourceSets?.[ 0 ]?.resources?.[ 0 ];
+						if ( ! metadata ) throw new Error( 'Could not retrieve Bing Maps metadata' );
```
gkjohnson: Same with this error.

Also, is there a known case where this happens? If it's just defensive then I'd prefer to remove the check.

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1612,30 +1617,72 @@ export class CesiumIonOverlay extends ImageOverlay {
...
+		this.options = options;
 		this.assetId = assetId;
 		this.auth = new CesiumIonAuth( { apiToken, autoRefreshToken } );
-		this.imageSource = new TMSImageSource( options );
 
 		this.auth.authURL = `https://api.cesium.com/v1/assets/${ assetId }/endpoint`;
-		this.imageSource.fetchData = ( ...args ) => this.fetch( ...args );
 		this._attributions = [];
 
+		this.externalType = null;
```
gkjohnson: For clarity lets make this a boolean.

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1612,30 +1617,72 @@ export class CesiumIonOverlay extends ImageOverlay {
...
+
+				this.externalType = json.externalType;
+				switch ( json.externalType ) {
+
+					case 'GOOGLE_2D_MAPS':
+						this.imageSource = new XYZImageSource( {
+							...this.options,
+							url: `${json.options.url}/v1/2dtiles/{z}/{x}/{y}?session=${json.options.session}&key=${json.options.key}`,
+							tileDimension: json.options.tileWidth,
+						} );
```
gkjohnson, line 1677: Is there a "known" number of tile levels for Google Map Tiles we can set here?

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1612,30 +1617,72 @@ export class CesiumIonOverlay extends ImageOverlay {
...
+						break;
+
+					}
+
+					default:
+						this.imageSource = new TMSImageSource( {
+							...this.options,
+							url: json.url,
+						} );
+						this.imageSource.fetchData = ( ...args ) => this.fetch( ...args );
```
gkjohnson: It shouldn't be necessary to assign "fetchData" here since it's assigned above.

src/three/plugins/images/ImageOverlayPlugin.js
Brakebein: Yes, it's just defensive. The resources should always be set. Or in other words: I didn't encounter anything else.

src/three/plugins/images/ImageOverlayPlugin.js
Brakebein: This information is not specified within `options`. But Google says it goes up to 22 (https://developers.google.com/maps/documentation/tile/2d-tiles-overview?hl=en). So, it would be hard-coded here.

src/three/plugins/images/sources/XYZImageSource.js
Brakebein: It's explained here: https://learn.microsoft.com/en-us/bingmaps/rest-services/directly-accessing-the-bing-maps-tiles

> The `imageUrlSubdomains` property in the imagery metadata response provides a list of valid subdomain that can be used in the tile URL. Using a different one for each tile request you can increase performance by get around browser URL request limits i.e. many browsers allow up to 8 concurrent requests to the same domain. Using subdomains allows up to 8 requests per subdomain.

Those browser request limits shouldn't actually be a problem anymore with HTTP/2.

src/three/plugins/images/sources/XYZImageSource.js
Brakebein: As far as I know, the quadkey scheme is Microsoft-specific. At least for the next 3 years, we can still use Bing maps. But it looks like it's going to be replaced by Azure Maps, which will use the XYZ scheme (https://learn.microsoft.com/en-us/rest/api/maps/render/get-map-tile?view=rest-maps-2025-01-01&tabs=HTTP). But at one point, they also say: "Some mapping platforms use a quadkey indexing naming convention ..."  (https://learn.microsoft.com/en-us/azure/azure-maps/zoom-levels-and-tile-grid?tabs=csharp).

I haven't worked yet with Azure Maps. I only had a quick look because of this PR. And I probably won't work with it in near future. It's hard to find something in their documentation.

src/three/plugins/images/sources/XYZImageSource.js
Brakebein: This part is actually a little bit dirty. Zoom level 0 (basically the whole world) does not exist within the quadkey scheme. It starts with zoom level 1. So querying zoom level 0 is actually invalid (or empty string `''`). But the tile renderer goes through all levels starting from level 0. Here, it would result in a request error, so I return `'0'` (which is basically level = 1, x = 0, y = 0, North America), which is not correct, but the first two levels aren't rendered anyway, or are very quickly replaced by the upper levels.

In short, it's to avoid a request error when querying level 0.

At this point, it would actually be good to define a minimum zoom level for the image source.

src/three/plugins/images/ImageOverlayPlugin.js
Brakebein: But I have removed it in the constructor. All imageSource initialization is now here. Or did I miss something?

src/three/plugins/images/sources/XYZImageSource.js
gkjohnson: Oh interesting thanks for the reference. This is very bing-API specific so let's add a comment with a link to that explanation.

src/three/plugins/images/ImageOverlayPlugin.js
gkjohnson: Okay lets remove it in that case.

src/three/plugins/images/ImageOverlayPlugin.js
gkjohnson: Great, lets hard code it with a link to that page.

src/three/plugins/images/ImageOverlayPlugin.js
gkjohnson: I see now - it's initialized in all other constructors but replaced in the "ImageOverlayPlugin" section. In that case we should be assigning `imageSource.fetchData` for all image source creation cases, which we should just be able to do after the switch statement.

src/three/plugins/images/sources/XYZImageSource.js
gkjohnson: The "TilingScheme" defines the different tiling levels that are available to load (available on ImageSource.tiles). `XYZImageSource` implicitly fills in all levels of tiling scheme since XYZ tile sets contain 1x1 image the root. It would probably be best to _not_ extend XYZImageSource or at least define a custom custom "init" function that initializes the TilingScheme like so:

```js
// generate "levels" number of zoom levels with 2x2 images at the root.
tiling.generateLevels( levels, 2, 2, {
	tilePixelWidth: tileDimension,
	tilePixelHeight: tileDimension,
} );
```

src/three/plugins/images/sources/XYZImageSource.js
Brakebein: Actually, some OSM providers also have multiple subdomains, with similar statement:

> `{s}` should be replaced by a subdomain. The exact values are in the tile server's documentation, but typically it is a single letter `a, b, c` [...] Subdomains are used to help with browser parallel requests per domain limitation (not needed if the webserver uses http/2), so consider switching letters at runtime 

https://wiki.openstreetmap.org/wiki/Raster_tile_providers

src/three/plugins/images/sources/XYZImageSource.js
Brakebein: I actually can't reproduce the request error. The root level map tile is returned when quadkey is just an empty string `''`. So, I removed the 2 lines.

gkjohnson: From their [docs](https://learn.microsoft.com/en-us/bingmaps/articles/bing-maps-tile-system) it looks like "level 1" is considered the root level but I see that their id construction logic implies an empty string could be level zero. I don't love that this isn't explicitly documented in their spec but we can go with it until there's an issue 😅

src/three/plugins/images/sources/XYZImageSource.js
gkjohnson: Interesting - this may be a more basic feature to include at the "TiledImageSource" level so subdomains can be used if their available. Something for another day if it's of interest.

Brakebein:
I think I've resolved all change requests. Is there anything else to do, or can this be merged?

gkjohnson:
Just had a few final changes I wanted to make - thanks!

PR #1392, opened 2025-12-08
Add missing option to GeoJSONOverlay type

MarkoXR:
Added the missing `strokeWidth` property to the `GeoJSONOverlay` type definition. The implementation already supported this option, but it was absent from the type definition, causing TypeScript error.


gkjohnson:
Thanks!

PR #1397, opened 2025-12-11
Update babylon-support branch to use es6 imports, module-level GTLF loader function, auto-detect plugin extension, and add/remove observables

georginahalpern:
- Update dependency to es6 and more recent babylon version
- Update imports to avoid import * from BABYLON
- Update gltfLoader to use module-level gltfloader ImportFileAsync (to enable treeshaking)
- Update gltfLoader to detect plugin extension vs hardcode glb
- Integrate babylon observables into add/remove event listeners


example/babylonjs/googleMapsAerial.js
```diff
@@ -1,4 +1,4 @@
-import * as BABYLON from 'babylonjs';
+import { Scene, Engine, Vector3, ArcRotateCamera }  from '@babylonjs';
```
gkjohnson, line 1: Is this supposed to import from `@babylonjs/core`?

src/babylonjs/renderer/loaders/GLTFLoader.js
```diff
@@ -13,6 +16,57 @@ export class GLTFLoader extends LoaderBase {
...
+
+			return false;
+
+		}
+
+		const view = new DataView( arrayBuffer, byteOffset, byteLength );
+		const magic = view.getUint32( 0, true ); // little-endian
+		return magic === GLB_MAGIC;
+
+	}
```
gkjohnson, line 41: Just to be sure: there's no way for Babylon to determine whether a file is a binary or ascii glTF automatically?

package.json
```diff
@@ -91,8 +91,8 @@
     "@types/three": "^0.170.0",
     "@vitejs/plugin-react": "^4.3.2",
     "@vitest/eslint-plugin": "^1.5.1",
-    "babylonjs": "^7.0.0",
-    "babylonjs-loaders": "^7.0.0",
+    "@babylonjs/core": "^8.39.3",
+    "@babylonjs/loaders": "^8.39.3",
```
gkjohnson, line 95: Let's update the "peerDependencies" section, too. I'll let you decide which versions are needed for the current babylon support but `>=8.0.0` seem reasonable to me.

gkjohnson (commented):
This is great, thanks! Just a few small changes and questions. I'll fix some of the lint errors in the other branch - looks like the CI only runs on PR into master.

Issue #1342, opened 2025-10-20
3D tiles are not updated during VR session


afdmulder:
### Question

I start a VR session with a specific errorTarget value. As a result, I initially see only a few 3D tiles which is fine. 
When I move my avatar towards an adjacent tile, I expect that tile to be loaded and displayed. 
However, that does not happen. The adjacent tiles are only shown after I interrupt or pause the VR session.

Question:
Is this behavior expected or am I missing something?

### Supplemental Data

Steps to reproduce behaviour:

1. Start VR session
2. Navigate to the an adjacent tile
3 .Wait for a moment: the adjacent tile is not shown
4. interrupt the VR session
5. the tile(s) will show up 

Example video:
https://3dcitytour.nl/support/3dcitytour_support.mp4

Example environment:
https://3dcitytour.nl/support

### Library Version

0.4.17

### Three.js Version

0.160.1

gkjohnson:
Hi @afdmulder  - without simple demo code it's hard to say what may exactly be going on but I recommend taking a look at the vr example which demonstrates a [assigning a custom scheduling callback for tile loads](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/7270fc20729aa1550510bc32692e2ce9d2ca440b/example/vr.js#L102-L111) so they can be [handled in the relevant request animation from loop](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/7270fc20729aa1550510bc32692e2ce9d2ca440b/example/vr.js#L255-L264). WebXR uses a separate rAF loop from the one on window (which is used for scheduling) and the tiles renderer otherwise has no way of knowing which render loop to use without assigning a custom callback.

afdmulder:
Hi Garrett, thanks for your reply and tips.  In fact I based my code on the VR example. 
In mains.js/reinstantiateTiles() I add the tilesets and I apply schedulingCallback.
In the mains.js/animate()  I call handleTasks like you do in the VR example.

However, it looks like tasks are not added during the VR session.  As soon as I pause the VR session, the numbers of tasks increases and the tasks are processed.

My test repository and branch
https://github.com/afdmulder/3DCityTour/tree/feature/vr_rendering

afdmulder:
I've noticed that the 3D tile update process does work if I leave the web browser open (after pausing my session) as a kind of standby screen. I can still navigate in the VR world then.  But as soon as I resume the VR session the update proces holds

See screenshot.

[...]

PR #1407, opened 2025-12-13
AF Manager class for dynamic animation frame request/cancel

phoenixbf:
See #1342

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -67,6 +67,7 @@ export default [
 			'@typescript-eslint/no-unused-vars': [ 'error', { args: 'none' } ],
 			'@typescript-eslint/no-explicit-any': 'off',
 			'@typescript-eslint/no-empty-object-type': 'off',
+			'@typescript-eslint/no-unsafe-function-type': 'off',
 		},
 	},
 
--- a/src/core/renderer/utilities/Scheduler.d.ts
+++ b/src/core/renderer/utilities/Scheduler.d.ts
@@ -1,0 +1,8 @@
+export class Scheduler {
+
+	static setXRSession( session: XRSession ): void;
+	static requestAnimationFrame( cb: Function ): number;
+	static cancelAnimationFrame( handle: number ): void;
+	static flushPending(): void;
+
+}


src/core/renderer/utilities/AFManager.js
```diff
@@ -0,0 +1,43 @@
+class AFManager {
```
gkjohnson: Lets call this `FrameScheduler` for now.

src/core/renderer/utilities/FrameScheduler.js
```diff
@@ -0,0 +1,43 @@
...
+
+		this.xrsession = xrsession;
+
+	}
+
+	removeXRSession() {
+
+		this.xrsession = null;
+
+	}
```
gkjohnson: When changing the system being used for callbacks we will have to cancel all previous callbacks and reregister them with the new function. Eg if the "LRUCache" schedules an unload event it will be waiting until that event fires before scheduling another one. So if the last event is scheduled using the "xrSession" callback and then the session closes, the scheduled callback will never be fired.

So we'll need to keep track of all scheduled callbacks and rAF handles to swap over the scheduled events.

gkjohnson (commented):
Great, thanks! A couple comments but I think this is the right direction

gkjohnson:
@phoenixbf - just wanted to check in to see if you had any more questions on the changes needed here.

phoenixbf:
> @phoenixbf - just wanted to check in to see if you had any more questions on the changes needed here.

Hi, performing some local testing in order to have a clean rAF tracking system and handling swaps

phoenixbf:
I've added management of pending AF.
It basically keeps track of requested AFs via handle as suggested, providing methods to cancel or flush all pending AFs. When entering XR session, it internally flushes and completes all pending, switching to `this.xrsession.requestAnimationFrame` for new requests.

So far I tried the above logic in both local and online setups without any evident issues, also with mutiple tsets. If this is the right direction, it should be probably instantiated in `TilesRendererBase` and then referenced by `LRUCache`, `PriorityQueue`, etc. during their initializations

gkjohnson:
Thanks the direction of this looks great. I think we can cut down on some code redundancy in the implementation but we can deal with that once everything is working.

> If this is the right direction, it should be probably instantiated in TilesRendererBase and then referenced by LRUCache, PriorityQueue, etc. during their initializations

This sounds good to me. Looks like the "LRUCache", "PriorityQueue", the "throttle" function, and "QueryManager" are all places where rAF is used (though the throttle function can maybe be changed to "queueMicrotask")

src/core/renderer/tiles/TilesRendererBase.js
```diff
@@ -270,6 +278,22 @@ export class TilesRendererBase {
 
 	}
 
+	setXRSession( xrsession ) {
+
+		if ( ! this.framescheduler ) return;
```
gkjohnson: this should never be "null" so we should be able to remove this check.

src/core/renderer/tiles/TilesRendererBase.js
```diff
@@ -189,19 +190,26 @@ export class TilesRendererBase {
 		this.cachedSinceLoadComplete = new Set();
 		this.isLoading = false;
 
+		// FrameScheduler referenced by LRUCache, PriorityQueue
+		this.framescheduler = new FrameScheduler();
```
gkjohnson: lets use camel casing to align with the class naming:

```js
this.framescheduler -> this.frameScheduler
```

src/core/renderer/utilities/LRUCache.js
```diff
@@ -56,6 +56,8 @@ class LRUCache {
 
 		this._unloadPriorityCallback = null;
 
+		this.framescheduler = null;
```
gkjohnson: Lets initialize these to `new FrameScheduler()` so the classes are in a usable state when constructed. Then we can remove the null checks, as well. This should help the tests pass, too.

gkjohnson:
I've just made some of the remaining changes to fix tests, add new tests for FrameScheduler, and address some code styling and other issues. I think this should be ready to merge but if you could test it in a real XR environment that would be great.

phoenixbf:
> I've just made some of the remaining changes to fix tests, add new tests for FrameScheduler, and address some code styling and other issues. I think this should be ready to merge but if you could test it in a real XR environment that would be great.

Yep, sure. Today I'll carry out additional online tests with a few HMDs

gkjohnson:
@phoenixbf - I just want to make sure you're not waiting on me for anything with this PR. Is this tested and ready to merge?

phoenixbf:
> @phoenixbf - I just want to make sure you're not waiting on me for anything with this PR. Is this tested and ready to merge?

Hi @gkjohnson  I tested it with different tsets and conditions (local and online scenarios), so far I did not encounter issues.
In the future maybe we can further improve a few setups (e.g. automatically setting the XR session) but so far it is working fine

gkjohnson:
Thanks - would you mind adding docs for the "setXRSession" function on TilesRendererBase? Then we can get this merged.

At some point it will probably make sense to implicitly share all of these caches across all instances by default (see #1418) so there's not as much need to manage all of these. But this can happen later.

h0ldemslav:
I've been following this pull request (issue) since January and I'm very happy that this problem with rAF in VR is about to be fixed! @phoenixbf @gkjohnson  thank you so much! 

I've got a question though. I know you have tested it in VR headsets. I also tried your changes, and tiles are indeed properly rendered in VR (Meta Quest Pro). However, have you tried rendering QuantizedMesh in VR on demand, with the framescheduler? For example, you open a webpage, enter the VR mode, and *then* render QuantizedMesh? It does work in browser on PC but not in the VR headset (I tried it in Meta Quest Pro). I'm not sure what I'm missing. I'm currently working on an example that you can also run and see the problem (providing I didn't miss something obvious and the problem really exists).

This problem doesn't seem to appear with the Dingo Gap VR sample.

h0ldemslav:
Looks like the problem still persists on Meta Quest Pro in Meta's Browser (at least for me). Here is the link to the repo: https://github.com/h0ldemslav/vr-example-with-frame-scheduler/tree/main

It's basically a merge of VR and quantized mesh examples that @gkjohnson had provided. I include the build of the fork of the repo with FrameScheduler

It has one file `src/main.js`, and there (it's quite messy but I think it should be sufficient) I instantiate quantized mesh and tiles (just a large red box near the Golden Gate Bridge), once I enter the VR mode in the quest headset. 
```js
    renderer.xr.addEventListener("sessionstart", () => {
           ....
            createGlobeControls();     <- ignore this
            addQuantizedMesh();
            addBoxTiles();
        }
    });
    
    ...
        renderer.setAnimationLoop(animate);
   ```
   
   In the animate function I call `handleCamera`, whose body sets VR camera (only once):
   ```js
               tiles.cameras.forEach((c) => tiles.deleteCamera(c));
            tiles.setCamera(xrCamera);
            goldenBridgeTiles.cameras.forEach((c) =>
                goldenBridgeTiles.deleteCamera(c),
            );
            goldenBridgeTiles.setCamera(xrCamera);

            const leftCam = xrCamera.cameras[0];
            if (leftCam) {
                tiles.setResolution(
                    xrCamera,
                    leftCam.viewport.z,
                    leftCam.viewport.w,
                );
                goldenBridgeTiles.setResolution(
                    xrCamera,
                    leftCam.viewport.z,
                    leftCam.viewport.w,
                );
            }

            xrSession = renderer.xr.getSession();
            tiles.setXRSession(xrSession);    <- new api FrameScheduler
            goldenBridgeTiles.setXRSession(xrSession);    <- new api FrameScheduler
```

Note that this example will work if you run it on the desktop. Just click on 'Enter VR'. You will see the red box with axes helper and the quantized mesh (cesium overlay). But if you try it in the headset, you'll see only the box and the axes helper. The 'workaround' is to exit the VR mode and then enter the mode again. In that case that terrain would load (not sure if it keeps updating though).
 
 Would you mind looking at it? I would be very grateful for any help/advice!

gkjohnson:
Thanks for testing further. One note is that with this change the special `tilesSchedulingCB` logic in the vr example should no longer be needed, so those lines could be removed and tested again, as well.

That aside there's nothing in the QuantizedMeshPlugin that that relies on rAF, which seems to be the primary behavioral different between native Meta Quest devices and other VR headsets & browser emulation, unless there's something else going on. My best advice is to add logs throughout some of these critical paths to make sure that things like tile updates, priority queue iteration, tile loading, etc and find the inconsistencies.

I would also recommend making an issue for [Meta's web emulator](https://github.com/meta-quest/immersive-web-emulator) to request that the rAF callback behavior be made consistent with the real hardware (or better add a toggle so different device / browser behaviors can be emulated).

phoenixbf:
> Looks like the problem still persists on Meta Quest Pro in Meta's Browser (at least for me). Here is the link to the repo: https://github.com/h0ldemslav/vr-example-with-frame-scheduler/tree/main
> 
> It's basically a merge of VR and quantized mesh examples that @gkjohnson had provided. I include the build of the fork of the repo with FrameScheduler
> 
> It has one file `src/main.js`, and there (it's quite messy but I think it should be sufficient) I instantiate quantized mesh and tiles (just a large red box near the Golden Gate Bridge), once I enter the VR mode in the quest headset.
> 
> ```js
>     renderer.xr.addEventListener("sessionstart", () => {
>            ....
>             createGlobeControls();     <- ignore this
>             addQuantizedMesh();
>             addBoxTiles();
>         }
>     });
>     
>     ...
>         renderer.setAnimationLoop(animate);
> ```
> 
> In the animate function I call `handleCamera`, whose body sets VR camera (only once):
> 
> ```js
>             tiles.cameras.forEach((c) => tiles.deleteCamera(c));
>          tiles.setCamera(xrCamera);
>          goldenBridgeTiles.cameras.forEach((c) =>
>              goldenBridgeTiles.deleteCamera(c),
>          );
>          goldenBridgeTiles.setCamera(xrCamera);
> 
>          const leftCam = xrCamera.cameras[0];
>          if (leftCam) {
>              tiles.setResolution(
>                  xrCamera,
>                  leftCam.viewport.z,
>                  leftCam.viewport.w,
>              );
>              goldenBridgeTiles.setResolution(
>                  xrCamera,
>                  leftCam.viewport.z,
>                  leftCam.viewport.w,
>              );
>          }
> 
>          xrSession = renderer.xr.getSession();
>          tiles.setXRSession(xrSession);    <- new api FrameScheduler
>          goldenBridgeTiles.setXRSession(xrSession);    <- new api FrameScheduler
> ```
> 
> Note that this example will work if you run it on the desktop. Just click on 'Enter VR'. You will see the red box with axes helper and the quantized mesh (cesium overlay). But if you try it in the headset, you'll see only the box and the axes helper. The 'workaround' is to exit the VR mode and then enter the mode again. In that case that terrain would load (not sure if it keeps updating though).
> 
> Would you mind looking at it? I would be very grateful for any help/advice!

So far tested several tilesets (using both v1.0 and v1.1 specs), thus including b3dm and glb tiles (with draco + ktx compressions). In my setup I have also fade and v1.1 plugins enabled, without issues, in prolonged VR sessions.

I'll test your sample ASAP: maybe something specific with your setup? or something happening with quantized meshes?

Here testing on MetaQuest Pro with build v2.1.1034, using built-in browser.

h0ldemslav:
> So far tested several tilesets (using both v1.0 and v1.1 specs), thus including b3dm and glb tiles (with draco + ktx compressions). In my setup I have also fade and v1.1 plugins enabled, without issues, in prolonged VR sessions.
> 
> I'll test your sample ASAP: maybe something specific with your setup? or something happening with quantized meshes?
> 
> Here testing on MetaQuest Pro with build v2.1.1034, using built-in browser.

Thanks!! I don't really have any specific setup. I know that in my app quantized mesh didn't work as intended, so I was hoping that the fix with rAF would help, or maybe I really missed something. That's why I created a very minimal example using VR example and quantized mesh overlays example to see if I mess something up or not, and if the fix solves my issue. But it didn't work either. Anyway, by the end of this week I'll try to get to my example and add some logs as gkjohnson suggests to see if quantized mesh does something.

h0ldemslav:
Ok, so what I found so far:
- PriorityQueue works, it has some items and it does run jobs
- tiles that have quantized mesh fire events

What I discovered, however, is that when you press meta button on the controller to sort of 'quit' the webapp, you don't really quit, you just open the browser window that overlaps the scene. If you attach an event listener to the xr session (say after sessionstart) for the event 'visibilitychange', then you will find out that the session is in 'visible-blurred' visibilityState. Technically I'm still in VR session. And if I continue staying in that state (opened browser window), the imagery starts loading, as well as the quantized mesh is being rendered. If I go back to the scene (press the meta button again), then the imagery stops loading.

So I think the problem might be in the ImageOverlayPlugin/CesiumIonOverlay. Because when I commented out the addition of CesiumIonOverlay (or the whole ImageOverlayPlugin), then my quantized mesh rendered, without me pressing the meta button. Any suggestions? Thanks!!

gkjohnson:
The ImageOverlayPlugin creates a "PriorityQueue" internally so that would need to share a FrameScheduler, as well. The more of these changes that need to be made, though, the more brittle this solution seems. I'm wondering if we should just provide a global FrameScheduler that everything shares by default so these issues aren't encountered.

One other potential case that comes to mind: Is there ever a case where the XR Session is "paused" (as in not rendering or updating) when a separate window is open? It seems like there could be a case where the XR Session has initialized but the session "rAF" has stopped firing and the window "rAF" starts firing again, instead. Is this something you're able to reproduce? I'm wondering how to be resilient to this.

phoenixbf:
> The ImageOverlayPlugin creates a "PriorityQueue" internally so that would need to share a FrameScheduler, as well. The more of these changes that need to be made, though, the more brittle this solution seems. I'm wondering if we should just provide a global FrameScheduler that everything shares by default so these issues aren't encountered.
> 

I agree that a more centralized approach would be better (it was also my initial workaround when this issue appeared). A `FrameScheduler` is indeed needed at the moment to handle and delegate to the correct rAF routine. Where do you imagine this happening and how? A singleton?

> One other potential case that comes to mind: Is there ever a case where the XR Session is "paused" (as in not rendering or updating) when a separate window is open? It seems like there could be a case where the XR Session has initialized but the session "rAF" has stopped firing and the window "rAF" starts firing again, instead. Is this something you're able to reproduce? I'm wondering how to be resilient to this.

I dont know exactly how for instance Meta browser handles this.
Are they using a separate instance when pausing the session? So far it resumed fine during various tests

h0ldemslav:
> The ImageOverlayPlugin creates a "PriorityQueue" internally so that would need to share a FrameScheduler, as well. The more of these changes that need to be made, though, the more brittle this solution seems. I'm wondering if we should just provide a global FrameScheduler that everything shares by default so these issues aren't encountered.
> 
> One other potential case that comes to mind: Is there ever a case where the XR Session is "paused" (as in not rendering or updating) when a separate window is open? It seems like there could be a case where the XR Session has initialized but the session "rAF" has stopped firing and the window "rAF" starts firing again, instead. Is this something you're able to reproduce? I'm wondering how to be resilient to this.

Not sure if I understand the last paragraph. I mean rAF is fired always, even if you have other windows in Meta quest open. What's interesting is that, both window.requestAnimationFrame and xrsession.requestAnimationFrame are fired by the frame scheduler, whether your session has visibilityState "visible" or "visible-blurred". 

I made a video to illustrate the problem. Have a look at it: https://drive.google.com/file/d/12ZfhMkAjkJPxaYLlh30wFM-1iiOrX0Fw/view?usp=sharing (cannot attach to GitHub, size exceeds GitHub limits). I'm not good at it but I can make more videos, if you need, feel free to give me concrete instructions.

In the video, you can see how the imagery is not applied at all and how there are very few calls to both rAF, when I was in VR, or better say visibilityState was "visible". Once I press the meta button on the controller, the session visibilityState changes to "visible-blurred" and you can see how many rAF are fired (both window and xrsession). Once I toggle the state (from "visible" to "visible-again"), there are very few rAF calls.

gkjohnson:
@phoenixbf 
> A FrameScheduler is indeed needed at the moment to handle and delegate to the correct rAF routine. Where do you imagine this happening and how? A singleton?

Yeah lets just make this a singleton for now. I'm open to suggestions but we can manage it like so:

```js
import { FrameScheduler } from '3d-tiles-renderer';

FrameScheduler.requestAnimationFrame( ... );
FrameScehduler.setXRSession( ... )
```

@h0ldemslav 
> Not sure if I understand the last paragraph. I mean rAF is fired always, even if you have other windows in Meta quest open. What's interesting is that, both window.requestAnimationFrame and xrsession.requestAnimationFrame are fired by the frame scheduler, whether your session has visibilityState "visible" or "visible-blurred".

The window's "requestAnimationFrame" will have to fire when the "window" is visible so content in just the window, like DOM, can be updated. And the xr session rAF will have to be fired when the XR session is presenting so content can be updated for the XR view. If the XR rAF didn't fire while the background was rendering then it would just freeze and the camera view would not be able to update, either. So what we're seeing are the following situations when the presenting === true (the xr session is running):

| | xr rAF fires | window rAF fires |
|---|---|---|
| immersive view | true | false |
| window is up with immersive background | true | true |

What I'm wondering is if there is every a situation we can get into where the "presenting" is true (eg sessionstart has fired & sessionend has not) and the window rAF _is_ firing while the xr session rAF is _not_ (this would be a line with "false | true" on the above table). If we think there _is_ or there may be in the future then we should make sure we account for this in our FrameSchduler implementation. Otherwise this will happen again and it's quite a hard issue to understand the track down.

Relatedly, I'd appreciate it if you could make any issue at the meta emulator and link it here so we can more easily understand and test this behavior:

> I would also recommend making an issue for [Meta's web emulator](https://github.com/meta-quest/immersive-web-emulator) to request that the rAF callback behavior be made consistent with the real hardware (or better add a toggle so different device / browser behaviors can be emulated).

--

> I made a video to illustrate the problem. Have a look at it: https://drive.google.com/file/d/12ZfhMkAjkJPxaYLlh30wFM-1iiOrX0Fw/view?usp=sharing (cannot attach to GitHub, size exceeds GitHub limits)

I can't read any of the text in this video with the compression, unfortunately, so I can't fully understand what its showing.

h0ldemslav:
> The window's "requestAnimationFrame" will have to fire when the "window" is visible so content in just the window, like DOM, can be updated. And the xr session rAF will have to be fired when the XR session is presenting so content can be updated for the XR view. If the XR rAF didn't fire while the background was rendering then it would just freeze and the camera view would not be able to update, either. So what we're seeing are the following situations when the presenting === true (the xr session is running):

I see, now that makes much more sense. Thanks a lot!! Your table basically summarized what I did in the video. Though some window.rAF are called even in the immersive view (perhaps this is some other part of the library). I don't know why Google drive still hasn't provided a better quality video.

Anyway, I'm not sure if one can reproduce a situation where the renderer.xr is presenting, the window.rAF *is* fired, while xrsession.rAF *is not* fired. (I can't really think of any good example) If you look at the [WebXR specification, 4.3](https://immersive-web.github.io/webxr/#animation-frames:~:text=The%20behavior%20of%20the%20Window,animation%20loops%20is%20demonstrated%20below%3A) (links to the highlighted part), they discourage to use window.rAF in the context of xrsession because "the timing of callbacks of window.rAF may not coincide with xrsession.rAF". Though there is some sort of the pattern they suggest using for "transition between these two types of animation loops ".

So, I would rule out the use case where you solely use window.rAF while renderer.xr presenting. I think referring to the WebXR specification is a good justification. However, I'm thinking that maybe the frame scheduler could also use a third option, `setTimeout(cb, 16)` as a wilcard option that would suffice anyway, though someone has already mentioned in the issues that tiles start flickering if settimeout is used. What do you think?

gkjohnson:
> However, I'm thinking that maybe the frame scheduler could also use a third option, setTimeout(cb, 16) as a wilcard option that would suffice anyway

I'm not sure what you mean by this? The classes using frame scheduler need to run every frame and setTimeout is not guaranteed to do so. It may run multiple times per frame or not at all depending on the platform and the framerate.

phoenixbf:
Hi there,
I've committed a revised FrameScheduler that how behaves as singleton. If the current proposal sounds good, it should be now possible to use this anywhere:

```
import { FrameScheduler } from '../utilities/FrameScheduler.js';

let fs = new FrameScheduler(); // this now behave as singleton

// do stuff with fs
fs.setXRSession(...);
fs.requestAnimationFrame(...);
```

gkjohnson:
Thanks @phoenixbf - I'll take a deeper look this weekend and make a few more changes I've had in the back of my mind.

It looks like one of the tests is failing, though. Would you be able to take a look?

gkjohnson:
I've made a few changes - primarily changing the class to use all static functions and using it as a singleton everywhere. The class has also been renamed to "Scheduler" and I've removed the "setXRSession" function from "TilesRenderer" so you set it like so:

```js
import { Scheduler } from '3d-tiles-renderer';
Scheduler.setXRSession( session );
```

I'll merge this but if you could test it and follow up if there are any issues that would be great! Thanks again for your help in tracking down the issue and coming up with a solution.

phoenixbf:
> I've made a few changes - primarily changing the class to use all static functions and using it as a singleton everywhere. The class has also been renamed to "Scheduler" and I've removed the "setXRSession" function from "TilesRenderer" so you set it like so:
> 
> ```js
> import { Scheduler } from '3d-tiles-renderer';
> Scheduler.setXRSession( session );
> ```
> 
> I'll merge this but if you could test it and follow up if there are any issues that would be great! Thanks again for your help in tracking down the issue and coming up with a solution.

Thanks, I'll test these ASAP on my XR equipment

phoenixbf:
just to confirm I tested several online tilesets on different equipment (Meta Quest 2, Meta Quest Pro and Hololens 2) without issues

Issue #1327, opened 2025-09-19
GeoJSONOverlay: Possible improvements


gkjohnson:
**Features**
- Add support for a form of "styling" the GeoJSON.
- Add support for multiple GeoJSON shapes (array of URLs, shapes).
- Add support for regenerating images based on color / shape / styling changes.

**Quality
- Handle GeoJSON shapes across the +-180 degree longitude boundary.
- Add padding based on pixel size, stroke size, and point size to ensure shapes are included in bounding box calculations to prevent shape clipping.
- Add support for world space sizing so line width etc does not scale based on zoom / pixel size.
- Consider scaling canvas dimensions based on ellipsoid-projected size to retain roughly square pixels, prevent distortion.

**Memory**
- Add support for limiting the bounds in which images are produced to only where shapes are present.

**Performance**
- See if converting to SVG is performant.
- Use web workers and offscreen canvas.

Ylannl:
Very nice to see this effort!

What do you think about a picking feature? This would be very useful, ie. to query attributes, and for the ability to 'select' features (which would then be reflected with a color/styling change of those features).

As a reference, there is some sort of picking feature for draped vectors in [Giro3D](https://giro3d.org/latest/examples/planar_vector.html?view=1837049.6%2C5179620.1%2C3012.9%2C1841174.4%2C5175167.1%2C0).

gkjohnson:
> What do you think about a picking feature? This would be very useful, ie. to query attributes, and for the ability to 'select' features (which would then be reflected with a color/styling change of those features).

Raycasting seems like a good feature to support. I have expected this to be implemented as a separate utility in an end-user application but maybe it makes sense to include support in the plugin. The most simple way to do this is provide a function that takes a latitude and longitude value (or tile-space value to transform into the relevant projection) and perform polygon intersection checks with the geojson shapes.

Figuring out how to efficiently support dynamically colored shapes with a tiled system like this would still take some thinking, though.

> As a reference, there is some sort of picking feature for draped vectors in [Giro3D](https://giro3d.org/latest/examples/planar_vector.html?view=1837049.6%2C5179620.1%2C3012.9%2C1841174.4%2C5175167.1%2C0).

It looks like Giro3D _doesn't_ support recoloring of the shapes, is that right?

Ylannl:
[...]

PR #1419, opened 2025-12-22
GeoJSONOverlay: Improve performance for complex data sets

gkjohnson:
Related to #1327 

Improves the performance with complex GeoJSON data sets by caching the features and bounds ahead of time rather than regenerating them every draw. Tested with the existing complex data set url:

image

When testing with the initial camera view a full "redraw" of all tiles goes down from ~2300ms to ~500ms, or an almost 80% improvement. This smooths out navigation quite a bit, too. The remaining time on the redraw method seems to be dominated by the actual "draw" portion of the function rather than the forceful cache update (the draw method could probably be improved, as well):

image

cc @SoftwareMechanic


SoftwareMechanic:
This is much appreciated!
we will get in touch again soon for other features 🚀

Issue #1375, opened 2025-11-24
Migrate more logic to core


gkjohnson:
Related to #1374, #1373, [initial Babylon js implementation](https://github.com/gkjohnson/babylon-3dtiles-demo)

In order to facilitate adding the loader in to other engines more logic from the three.js TilesRenderer can be migrated to the core implementation. Some ideas:

**High Priority**
- "calculateTileViewError" is probably the most complicated function to implement and could be broken up into multiple more simple functions to make this more simple.
  - Move plugin call to superclass - what to call getter on base class?
  - ~Is it sensible to calculate error other than Infinity for tiles out of view?~ Yes for sorting
- Move more events to core.
  - "dispose-model" (possibly need dispose-content, dispose-tileset sibling events)
  - "tile-visibility-change"

**Defer**
- "calculateTileViewError" simplification
  - Per-camera callback?
- Relatedly, an abstract "camera" list could be maintained so the base class can iterate over them.
- simplify "engineData" initialization if possible. Just define a "getter"?
- If "boundingVolume" is required and shaped consistently then raycasting logic could be moved to core.
- Abstract handling of the "gltfUpAxis" behavior somehow.
- Use "scene" for now
  - Consider renaming "scene" to "content" "group" "object" "node" "sceneNode" or "model"
  - The above should allow for events to implicitly fire a "scene" object that is currently used in TilesRenderer.

**Complete**
- Move more update() function logic to core.
  - "update-before" / "update-after"
  - Probably requires some more life cycle functions for "beforeUpdate" or "prepareForUpdate"
- Remove underscore prefixes from implicit subtree nodes
- Move traversal variables to a "traversal" & "internal" subobjects
- Codify a dedicate field for necessary and per-engine pieces of data per tile (see the "cached" subfield in three.js version). This could require fields such as "scene" and "boundingVolume".
  - Consider renaming "cached" to "data" or "cache"

Issue #1344, opened 2025-10-21
QuantizedMesh: skip loading low resolution tiles when camera initialised close to mesh


Ylannl:
### Description

I'm working on a project that uses the WMTS and GeoJSON overlays on a quantised mesh. This works well, except for the initialisation phase in case the camera is initialised close to the terrain:

https://github.com/user-attachments/assets/456ed008-e730-4f2c-8e66-7040da3bf0b8

It takes quite long to load all the tiles all the way up to the desired state for the current camera postion (with the exception of the 3D Tiles with building models). This is especially the case with a slow connection/slow server, eg. I was able to improve this situation somewhat by moving the quantised mesh to a more responsive server. However, it is still very noticeable, seeing the low resolution tiles gradually being replaced by the higher resolution ones. It is also noticeable when you zoom in very fast (eg. from globe level to city level).

### Solution

I get the impression that it always starts loading from the root (coarsest/lowest resolution tile) and then gradually refines without skipping any tree levels. Maybe even also for the overlays? Is this correct? And is there a way skip those intermediate levels with the existing codebase and immediately load the high resolutions tiles? 
If not, I think this would be a nice optimisation to implement. It would give the user a more responsive experience + save bandwidth. Do you agree? If yes, how could this implemented best? I'm willing to contribute code.

### Alternatives

Moving all the tile sources to faster servers, using CDN etc? This will, however, not tackle the root cause of the problem, only make it less noticeable.

gkjohnson:
Hi @Ylannl! Yes currently the tiles renderer loads all tiles from the root of the tile set down to the target ones. This has become more noticeable specifically as tile overlays are added and servers become slow to respond. I've been thinking this through for a bit now and looking at how Cesium is handling it and will hopefully be able to work on this in the coming months. Handling this correctly and adjusting the traversal algorithm is not so straight forward. Hopefully we'll be able to avoid things like sudden pops and gaps in tile geometry loading but we'll see.

gkjohnson:
A "optimizedLoadStrategy" that supports skipping parent tiles has now been introduced into the project with the ability to disable ancestor loading (see #1421, #1533). In theory this lets us skip straight to loading the tiles necessary for rendering at the cost of some initial load artifacts. But due to the way QuantizedMeshPlugin (and some others) work at the moment even when these settings are set correctly the parent tiles are still displaying.

I've created #1582 to track how to address this so I'll close this for now.

[...]

Issue #1304, opened 2025-09-03
no visible tiles for certain zoom levels


dbuck:
### Description

Seeing a visual blackout when transitioning to a  `"content": { "uri": "4_0_1.json" }` child tile.

from: https://raw.githubusercontent.com/matterport/3DTilesSampleData/refs/heads/main/uskTEa4gEb5/mesh_tiles/4.json

to: https://raw.githubusercontent.com/matterport/3DTilesSampleData/refs/heads/main/uskTEa4gEb5/mesh_tiles/4_0_1.json

### Reproduction steps

1. https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/index.html?unlit=1#https://raw.githubusercontent.com/matterport/3DTilesSampleData/refs/heads/main/uskTEa4gEb5/mesh_tiles/4.json 
2. zoom in on the scan
3. Note at the point where it's transitions from the lowest to next lod level, there is a certain zoom level where it's blank? continue zooming in and it'll load the child tiles.

video:

https://github.com/user-attachments/assets/6e42ef79-6ede-45ac-bf93-18920c7b9bda

### Code

```js
// code goes here
```

### Live example

Live example on the current master/bundle builds:

https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/index.html?unlit=1#https://raw.githubusercontent.com/matterport/3DTilesSampleData/refs/heads/main/uskTEa4gEb5/mesh_tiles/4.json 

### Library Version

v0.4.15

### Three.js Version

r176

dbuck:
fwiw the canTraverse is negative on that empty parent tile. A helpful breakpoint that gets triggered right as the parent tile is dropped. That's definitely not the right spot to 'fix' though because the child tiles haven't been processed/loaded by the time it's being stopped.

<img width="1020" height="309" alt="Image" src="https://github.com/user-attachments/assets/90b91bc3-0403-4aa9-a556-86e7bc6b1768" />

gkjohnson:
[...]

Issue #857, opened 2024-12-07
TilesRenderer: Add support for "virtual" child traversal to avoid unnecessary memory usage


gkjohnson:
Specifically this would be useful for implicit tiling. Can use a pool of child tile objects and only "generate" children that are traversed. This may cause issues for cached data, though.

gkjohnson:
Added support for a "virtual" flag in #1498 to indicate that a child has been generated based on parent tile geometry / state.

PR #1421, opened 2025-12-25
Tiles Traversal: Only load necessary tiles

gkjohnson:
Related to #1344

- Add support for an optimized tile loading traversal that will only load tiles that need to be loaded. Based on on the [Cesium Unreal traversal approach](https://cesium.com/learn/cesium-native/ref-doc/selection-algorithm-details.html).
- Add "loaded" field to stats.

**Algorithm TODO**
- ~Add support for "kicking" to support zoom in without flashing~
- ~Delete unused queued tiles~ (see [#1422](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1422))
- ~Add support for "ancestor meets sse" logic for support zoom out without flashing~
- ~Test with google tiles data set~
- ~Correct handling of "ancestor meets sse" logic~
- ~Ensure raycasting works as expected~
- ~Allow for "trickle in" of tiles on initial load~
- ~Investigate high number of active tiles vs typical traversal~
- ~Add a "loadSiblings" option~
- ~Understand why a low LoD tile loads initially~
- ~Re-add matterport fix [#1304](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1304)~
- ~Investigate weird raycasting / low lod tile here:~
  - http://localhost:5173/three/googleMapsExample.html#lat=1.4056&lon=35.7235&height=1296.49&az=-3.31&el=-23.53&roll=-0.00
  - Happens because all children are outside the frustum so this gets hidden but is still active due to load siblings. Not much to be done I don't think
  - The flash occurs due to the child tileset being loaded
- ~Document limitations~
  - shared caches and queues need to be shared by tiles using the same strategy
  - fade tiles does not work
  - any "virtual" or generated geometry will not work (quantized mesh splitting, overlay splitting)
- ~Validate that intermediate tiles do not load~
- ~Understand the camera "jump" on reload~
- ~Ensure additive refinement works as expected~
- ~Clean up~

**Future**
- Review dynamically generated / expanded tiles to ensure it works
  - These do not work since they only expand the children once the tile itself is loaded and disposed of once the internal parents are disposed.
  - That means we can get into an internal oscillation state when the parents are disposed, subsequently disposing of the children, causing that parent to render again.
- Add traversal utilities to improve readability of traversal algorithm
- Separate traversal variables into a dedicated object (related to #1375)
- Ensure compatibility with broader plugins including fading & tile splitting (quantized mesh, overlays, related to [#857](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/857))
- Make this the default and only traversal strategy once known to be stable
- Add delay to downloads / tile updates to prevent unnecessary downloads & data used ([#1423](https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1423))

### DIFF
--- a/src/core/renderer/tiles/optimizedTraverseFunctions.js
+++ b/src/core/renderer/tiles/optimizedTraverseFunctions.js
@@ -19,12 +19,14 @@ function isUsedThisFrame( tile, frameCount ) {
 
 }
 
+// Checks whether all children have been processed and are ready to traverse
 function areChildrenProcessed( tile ) {
 
 	return tile.__childrenProcessed === tile.children.length && ( ! tile.__hasUnrenderableContent || isDownloadFinished( tile.__loadingState ) );
 
 }
 
+// Checks whether we can stop at this tile for rendering or not
 function canUnconditionallyRefine( tile ) {
 
 	return tile.__hasUnrenderableContent || ( tile.parent && tile.parent.geometricError < tile.geometricError );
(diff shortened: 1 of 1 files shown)


gkjohnson:
A comparison of this view:

image

| | Geometry Tiles Loaded | Internal Tilesets Loaded | Total | Delta |
|---|---|---|---|---|
| master | 340 | 77 | 417 | 0 |
| optimized | 254 | 77 | 331 | 86 |
| no siblings | 164 | 50 | 214 | 203 |

gkjohnson:
cc @dbuck - with this change you will no longer be able to override the download and parse queue priority callbacks. I seem to recall you mentioning this is something you were doing previously but I'm not sure if you're still using that.

PR #1422, opened 2025-12-26
TilesRenderer: Remove loading tiles that are no longer used

gkjohnson:
Related to #1421 

Removes tiles that are currently pending but no longer used to prioritize other downloads. It's not clear whether this is actually beneficial and how much time it may really save or even whether it could cause an issue with local camera movements as the user moves the camera. There's value in loading tiles that may come into view soon and unused, pending tiles are already the first to be discarded. The best this does is clear the download / parse queue for newly needed tiles more quickly after the camera move.

An alternative would be to allow for the user to specify that unused pending tiles should be removes or give them a function to call to do so.

### DIFF
--- a/example/three/googleMapsExample.js
+++ b/example/three/googleMapsExample.js
@@ -351,7 +351,7 @@ function updateHtml() {
 
 		const lruCache = tiles.lruCache;
 		const cacheFullness = lruCache.cachedBytes / lruCache.maxBytesSize;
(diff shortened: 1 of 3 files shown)


gkjohnson:
Modified to only remove "QUEUED" tiles which are no longer needed, ensuring download bandwidth and parsing time have not been wasted.

PR #1429, opened 2025-12-31
TilesRenderer: Change "cached" to "engineData"

gkjohnson:
Related to #1375 

- Change "cached" to "engineData"
- Define the "engineData" object in "TilesRendererBase" for reusability

**Future PR TODO**
- Adjust naming of "scene" to a more common field name with backwards compatibility (possibly "model" and "modelMetadata")
- Move traversal fields to a "traversal" field


gkjohnson:
cc @dbuck - Similar to #1421, the internal "tile.cached" field is changing, which I know you'd discussed using in the past. #1431 will also be reorganizing some other internal tile fields on tile, as well.

PR #1432, opened 2026-01-04
Ellipsoid context

### LINKED ISSUE #1257: r3f components: Add support for ellipsoid context
Rather than relying on "tiles" to be present via a tiles context some components only really need the ellipsoid and frame (compass, controls, etc).

pizza3:
Fix #1257

Introduces a new EllipsoidContext to provide ellipsoid and frame data separately from the full TilesRenderer context
Refactors ControlsBaseComponent and CompassGizmo to consume the new context instead of directly accessing TilesRendererContext

How to test:
Can be tested on r3f/globe.html


gkjohnson:
Thanks!

PR #1435, opened 2026-01-07
Fix mouseToCoords in controls (avoid subtracting offset twice)

pmi-gbs:
The offset is subtracted already in the PointerTracker, so when adding e.g. a margin at the canvas the coordinates are wrong.


gkjohnson:
Thanks for the catch and the fix!

PR #1438, opened 2026-01-09
Fix TilesRenderer options types

ebrann94:
Change type of `optimizedLoadStrategy` and `loadSiblings` from number to boolean


gkjohnson:
Great, thanks!

Issue #973, opened 2025-02-13
EnvironmentControls: Allow for double-tap to zoom


gkjohnson:
Ala Google Maps and other maps platforms. Need an option to disable it or cancel the behavior in case users are including this interaction in their apps.

pizza3:
https://github.com/user-attachments/assets/a4dcdcd3-8629-4804-af40-d05f386a4f1d

wanted to confirm the implementation before making pr @gkjohnson 

1. **Tap Detection**: When user lifts finger, checks if second tap occurred within 300ms at same location (±10px) as first tap
2. **Target Acquisition**: Raycasts from tap position to find 3D point in scene; if hit, zooms toward that point, otherwise zooms toward camera direction
3. **Animation Setup**: Calculates start/target values (zoom property for ortho camera, or distance for perspective camera) and clamps to min/max limits
4. **Frame-by-Frame Update**: Each frame interpolates between start→target using ease-out cubic easing (starts fast, ends slow)
5. **Camera Movement**: Updates camera.zoom (orthographic) or moves camera.position along direction vector (perspective) until animation completes after 300ms.

gkjohnson:
Thanks @pizza3! This sounds great - my only question is about the animation and how that's handled. But I think seeing a PR will make that more clear.

PR #1439, opened 2026-01-13
Add double-tap zoom for Environment Controls

pizza3:
#973

Implements double-tap to zoom functionality for EnvironmentControls, inspired by Google Maps and other modern mapping platforms. This adds an intuitive touch gesture for zooming in on mobile and tablet devices.

@gkjohnson the animation is in `_updateDoubleTapAnimation`

**Complete Implementation**
1. Configuration Options ([EnvironmentControls.js:107-109](src/three/renderer/controls/EnvironmentControls.js))

```
this.enableDoubleTapZoom = true;           // Enable/disable feature
this.doubleTapZoomScale = 2.0;             // 2x zoom multiplier
this.doubleTapAnimationDuration = 0.3;     // 300ms animation
```
2. State Management ([EnvironmentControls.js:133-136](src/three/renderer/controls/EnvironmentControls.js))

```
this._lastTapTime = 0;
this._lastTapPoint = new Vector2();
this._doubleTapAnim = null;  // Single object for animation state
```
3. Tap Detection ([EnvironmentControls.js:442-450](src/three/renderer/controls/EnvironmentControls.js))
```
Added in pointerupCallback
Detects single-finger touch taps
Calls _detectDoubleTap() before pointer cleanup

```
4. Three Private Methods ([EnvironmentControls.js:1141-1244](src/three/renderer/controls/EnvironmentControls.js))
```
_detectDoubleTap() - 300ms window, 10px tolerance
_startDoubleTapZoom() - Raycasts to find target, sets up animation
_updateDoubleTapAnimation() - Ease-out cubic interpolation
```
5. Update Loop Integration ([EnvironmentControls.js:713-747](src/three/renderer/controls/EnvironmentControls.js))
```
Calls _updateDoubleTapAnimation() every frame
Blocks normal zoom/drag/rotate during animation
Dispatches change events
```

### DIFF
--- a/src/three/renderer/controls/EnvironmentControls.js
+++ b/src/three/renderer/controls/EnvironmentControls.js
@@ -104,6 +104,9 @@ export class EnvironmentControls extends EventDispatcher {
 		this.adjustHeight = true;
 		this.enableDamping = false;
 		this.dampingFactor = 0.15;
+		this.enableDoubleTapZoom = true;
+		this.doubleTapZoomScale = 2.0;
+		this.doubleTapAnimationDuration = 0.3;
 
 		this.fallbackPlane = new Plane( new Vector3( 0, 1, 0 ), 0 );
 		this.useFallbackPlane = true;
@@ -126,6 +129,11 @@ export class EnvironmentControls extends EventDispatcher {
 		this.zoomDirection = new Vector3();
 		this.zoomPoint = new Vector3();
 		this.zoomDelta = 0;
+
+		// double-tap state
+		this._lastTapTime = 0;
+		this._lastTapPoint = new Vector2();
+		this._doubleTapAnim = null; // stores { startTime, startValue, targetValue, zoomPoint }
 
 		// fields used for inertia
 		this.rotationInertiaPivot = new Vector3();
@@ -431,6 +439,16 @@ export class EnvironmentControls extends EventDispatcher {
 
 			}
 
+			// Check for double-tap before deleting pointer
+			const isTouch = pointerTracker.getPointerType() === 'touch';
+			const isSingleTouch = pointerTracker.getPointerCount() === 1;
+
+			if ( this.enableDoubleTapZoom && isTouch && isSingleTouch ) {
+
+				this._detectDoubleTap( pointerTracker );
+
+			}
+
 			pointerTracker.deletePointer( e );
 
 			if (
@@ -442,7 +460,13 @@ export class EnvironmentControls extends EventDispatcher {
 
 			}
 
-			this.resetState();
+			// Don't reset state if we're animating a double-tap zoom
+			if ( ! this._doubleTapAnim ) {
+
+				this.resetState();
+
+			}
+
 			this.needsUpdate = true;
 
 		};
@@ -687,27 +711,33 @@ export class EnvironmentControls extends EventDispatcher {
 
 		// update the actions
 		const inertiaNeedsUpdate = this._inertiaNeedsUpdate();
-		const adjustCameraRotation = this.needsUpdate || inertiaNeedsUpdate;
-		if ( this.needsUpdate || inertiaNeedsUpdate ) {
+		const doubleTapNeedsUpdate = this._updateDoubleTapAnimation();
+		const adjustCameraRotation = this.needsUpdate || inertiaNeedsUpdate || doubleTapNeedsUpdate;
+		if ( this.needsUpdate || inertiaNeedsUpdate || doubleTapNeedsUpdate ) {
 
 			const zoomDelta = this.zoomDelta;
 
-			this._updateZoom();
-			this._updatePosition( deltaTime );
-			this._updateRotation( deltaTime );
+			// Don't process normal zoom/position/rotation while animating double-tap
+			if ( ! this._doubleTapAnim ) {
+
+				this._updateZoom();
+				this._updatePosition( deltaTime );
+				this._updateRotation( deltaTime );
+
+			}
 
 			if ( state === DRAG || state === ROTATE ) {
 
 				_forward.set( 0, 0, - 1 ).transformDirection( camera.matrixWorld );
 				this.inertiaTargetDistance = _vec.copy( pivotPoint ).sub( camera.position ).dot( _forward );
 
-			} else if ( state === NONE ) {
+			} else if ( state === NONE && ! this._doubleTapAnim ) {
 
 				this._updateInertia( deltaTime );
 
 			}
 
-			if ( state !== NONE || zoomDelta !== 0 || inertiaNeedsUpdate ) {
+			if ( state !== NONE || zoomDelta !== 0 || inertiaNeedsUpdate || doubleTapNeedsUpdate ) {
 
 				this.dispatchEvent( _changeEvent );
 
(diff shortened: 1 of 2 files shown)


src/three/renderer/controls/EnvironmentControls.js
```diff
@@ -127,6 +130,11 @@ export class EnvironmentControls extends EventDispatcher {
 		this.zoomPoint = new Vector3();
 		this.zoomDelta = 0;
 
+		// double-tap state
+		this._lastTapTime = 0;
+		this._lastTapPoint = new Vector2();
+		this._doubleTapAnim = null; // stores { startTime, startValue, targetValue, zoomPoint }
```
gkjohnson: Lets initialize this during initialization rather than setting it to null.

src/three/renderer/controls/EnvironmentControls.js
```diff
@@ -1114,6 +1144,112 @@ export class EnvironmentControls extends EventDispatcher {
 
 	}
 
+	_detectDoubleTap( pointerTracker ) {
+
+		const currentPoint = new Vector2();
```
gkjohnson: Lets use an existing scratch variable for this.

src/three/renderer/controls/EnvironmentControls.js
```diff
@@ -1114,6 +1144,112 @@ export class EnvironmentControls extends EventDispatcher {
...
+		pointerTracker.getLatestPoint( currentPoint );
+
+		const now = performance.now();
+		const timeSinceLastTap = now - this._lastTapTime;
+		const moveDistance = pointerTracker.getMoveDistance();
```
gkjohnson: "getMoveDistance" seems to always be "0" because the "pointerup" event is fired without any "pointerMove" events on the same frame. And the "pointerTracker.updateFrame()" call in "update()" resets the distances. Have you checked if this is returning what you expect it to?

src/three/renderer/controls/EnvironmentControls.js
```diff
@@ -1114,6 +1144,112 @@ export class EnvironmentControls extends EventDispatcher {
...
+	_startDoubleTapZoom( tapPoint ) {
+
+		const { camera, domElement, raycaster } = this;
+
+		// Find zoom target point
+		_pointer.copy( tapPoint );
+		adjustedPointerToCoords( _pointer, domElement, _pointer );
+		setRaycasterFromCamera( raycaster, _pointer, camera );
+		const hit = this._raycast( raycaster );
+		const zoomPoint = hit ? hit.point.clone() : camera.position.clone().addScaledVector( raycaster.ray.direction, 100 );
```
gkjohnson, line 1189: Please use the scratch variables

src/three/renderer/controls/EnvironmentControls.js
```diff
@@ -1114,6 +1144,112 @@ export class EnvironmentControls extends EventDispatcher {
...
+
+		const { camera } = this;
+		const { startTime, startValue, targetValue, zoomPoint } = this._doubleTapAnim;
+		const elapsed = this.clock.getElapsedTime() - startTime;
+		const duration = this.doubleTapAnimationDuration;
+
+		// Ease-out cubic: t = 1 - (1-t)³
+		const progress = Math.min( elapsed / duration, 1 );
+		const t = 1 - Math.pow( 1 - progress, 3 );
+		const currentValue = startValue + ( targetValue - startValue ) * t;
```
gkjohnson: Can we use MathUtils.lerp here?

src/three/renderer/controls/EnvironmentControls.js
```diff
@@ -1114,6 +1144,112 @@ export class EnvironmentControls extends EventDispatcher {
...
+
+	}
+
+	_updateDoubleTapAnimation() {
+
+		if ( ! this._doubleTapAnim ) return false;
+
+		const { camera } = this;
+		const { startTime, startValue, targetValue, zoomPoint } = this._doubleTapAnim;
+		const elapsed = this.clock.getElapsedTime() - startTime;
```
gkjohnson: We should take a "deltaTime" passed into the function and thread it through the "update" function, as is done elsewhere in the class. Then the animation time can be counted down on each update.

src/three/renderer/controls/EnvironmentControls.js
```diff
@@ -442,7 +460,13 @@ export class EnvironmentControls extends EventDispatcher {
 
 			}
 
-			this.resetState();
+			// Don't reset state if we're animating a double-tap zoom
+			if ( ! this._doubleTapAnim ) {
+
+				this.resetState();
+
```
gkjohnson: Why is this necessary? It doesn't seem to impact anything relating to the animation

src/three/renderer/controls/EnvironmentControls.js
```diff
@@ -687,27 +711,33 @@ export class EnvironmentControls extends EventDispatcher {
...
-			this._updatePosition( deltaTime );
-			this._updateRotation( deltaTime );
+			// Don't process normal zoom/position/rotation while animating double-tap
+			if ( ! this._doubleTapAnim ) {
+
+				this._updateZoom();
+				this._updatePosition( deltaTime );
+				this._updateRotation( deltaTime );
+
+			}
```
gkjohnson: This guard shouldn't be needed, either. If the zoom animation has begun / is animating then the mouse / touch should be up and there should be no way for these functions to do anything

src/three/renderer/controls/EnvironmentControls.js
```diff
@@ -687,27 +711,33 @@ export class EnvironmentControls extends EventDispatcher {
...
 
 			if ( state === DRAG || state === ROTATE ) {
 
 				_forward.set( 0, 0, - 1 ).transformDirection( camera.matrixWorld );
 				this.inertiaTargetDistance = _vec.copy( pivotPoint ).sub( camera.position ).dot( _forward );
 
-			} else if ( state === NONE ) {
+			} else if ( state === NONE && ! this._doubleTapAnim ) {
```
gkjohnson: Same here - this check should not be needed. if the double tap has begun then all the residual inertia field should have been reset. Otherwise we'll run into a case where the "inertia" just begins again once the animation completes.

gkjohnson (commented):
Thanks! I've left a few comments about the code structure - I think a few things can be simplified. Regarding that actual usability, here are two things I've noticed:
- It's currently not possible to "cancel" the animation. As in if you place your finger on the screen to draw while the "zoom" animation is occurring then the animation should cancel so the user can interact.
- If you drag your finger a long distance, then release and tap again in the same spot this counts as a "double tap".

gkjohnson:
@pizza3 is this ready to review again? I saw you've made a commit but it's not clear if it's ready with no comment or review request.

pizza3:
hey @gkjohnson sorry I was ill 🤒 for past 2 weeks wasnt online, yes the changes you requested are done, and it is ready to be reviewed again

src/three/renderer/controls/EnvironmentControls.js
gkjohnson: This doesn't look like it's been addressed, yet. We should avoid cloning vectors as much as possible.

src/three/renderer/controls/EnvironmentControls.js
```diff
@@ -687,15 +720,19 @@ export class EnvironmentControls extends EventDispatcher {
...
 		const inertiaNeedsUpdate = this._inertiaNeedsUpdate();
-		const adjustCameraRotation = this.needsUpdate || inertiaNeedsUpdate;
-		if ( this.needsUpdate || inertiaNeedsUpdate ) {
+		const doubleTapNeedsUpdate = this._updateDoubleTapAnimation( deltaTime );
+		const adjustCameraRotation = this.needsUpdate || inertiaNeedsUpdate || doubleTapNeedsUpdate;
+		if ( this.needsUpdate || inertiaNeedsUpdate || doubleTapNeedsUpdate ) {
 
 			const zoomDelta = this.zoomDelta;
 
+			// Don't process normal zoom/position/rotation while animating double-tap
```
gkjohnson, line 729: What is this comment for? It doesn't seem to be correct - the double tap animation is not preventing any of these updates.

gkjohnson (commented):
Thanks - it looks like there are still a few things that haven't been updated. I'm also seeing that when double-tapping to zoom from afar in the "googleMapsExample", then checking again on the the globe near the edge the screen the camera will "jump" to a new position.

GlobeControls automatically orients the camera after movement to conform to the orientation of the ellipsoid and prevent the orientation from becoming inconsistent, which isn't happening in the tap-to-zoom logic. The "update zoom" function could probably be refactored to allow for zooming in a certain direction to a certain point, allowing for this orientation correct logic to operate consistently

gkjohnson:
Added in #1699

PR #1440, opened 2026-01-13
feat: WIP add MVT and PMTiles support

AlaricBaraou:
This PR is a first attempt at introducing support for Mapbox Vector Tiles (MVT).
I explored two approaches, rendering either as native Three.js geometry or as rasterized textures. 
This addresses the goals outlined in #982 and expands upon the initial work in #990.

This can be used as a start to build on top of / refine.

I have very little knowledge of MVT, most of this have been achieved by imitating the existing plugins and loaders and through LLM assistance.

While it seems to mostly work ( especially the rasterized  approach )
It would be better to validate / review everything.

I'm more than happy to assist with that and will commit my own future improvements as I start to use this in my app. 

Here is a video of the two added examples.

http://localhost:5173/three/mvt.html to debug loading a single tile and comparing it with the expected result.
http://localhost:5173/three/mvt_globe.html to test the projection of the tiles on a globe.

https://github.com/user-attachments/assets/de51f29a-428a-48d5-9fc7-4db050f7524a


gkjohnson:
Closing in favor of #1575

Issue #1441, opened 2026-01-15
TilesRendererBase.ensureChildrenArePreprocessed does not evaluate all children


Jordan-Lane:
As always, thank you to all the contributors to this amazing library! First PR here, so please let me know if there is anything I am missing or can add to help fix this bug.

### What is the issue?

The `TileRendererBase` method `ensureChildrenArePreprocessed` checks if a child has the `__depth` attribute initialized. If it is, then it breaks out of the for loop and assumes that all later children are initialized. This can result in undefined children which can cause a crash in the ThreeJS render or when trying to access `Tile` attributes. 

### How was it discovered?

I discovered this issue when trying to enable / disable the `DebugTilePlugin` on the fly with 3D Tiles. Disabling the plugin while nodes were loading could result in this error being thrown:

![Image](https://github.com/user-attachments/assets/539ade32-9a8c-47c2-889f-6ad1056e9790)

Upon being disabled, the `DebugTilePlugin` calls `_onDisposeModel` for every tile, however the app crashes when trying to access `tile.cached` on an undefined tile. 

## What is the fix?

The fix is very trivial, in the `ensureChildrenArePreprocessed` we just continue the loop instead of breaking, ensuring that all children are preprocessed and defined./

### Reproduction steps

1. Add multiple large 3D tiles to the ThreeJS scene
2. While 3D tiles are loading, register and enable a DebugTilePlugin to each 3d tile. 
3. Disable each of the DebugTilePlugins
4. May take a couple tries, but occasionally the app will crash with an undefined `tile`

### Code

```js
// code goes here
```

### Live example

Currently do not have a minimal reproduction, however I can take the time to try and recreate the bug in a minimal environment if required

### Screenshots & Tile Set

![Image](https://github.com/user-attachments/assets/539ade32-9a8c-47c2-889f-6ad1056e9790)

### Library Version

v0.4.18

### Three.js Version

r0.170.0

PR #1442, opened 2026-01-15
fix: ensure all children are preprocessed

Jordan-Lane:
Related issue found here: https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1441

### DIFF
--- a/src/core/renderer/tiles/TilesRendererBase.js
+++ b/src/core/renderer/tiles/TilesRendererBase.js
@@ -792,7 +792,7 @@ export class TilesRendererBase {
 			if ( '__depth' in child ) {
 
 				// the child has already been processed
-				break;
+				continue;
 
 			} else if ( immediate ) {
 


src/core/renderer/tiles/TilesRendererBase.js
```diff
@@ -792,7 +792,7 @@ export class TilesRendererBase {
 			if ( '__depth' in child ) {
 
 				// the child has already been processed
-				break;
+				continue;
```
gkjohnson, line 795: Thanks! I'm not sure if this fixes the issue, though - this loop breaks here because if __depth is defined here it means that all children have been processed or at least queued already since at least 1 is already processed. Continuing in the loop won't change that. There will also still be tiles elsewhere in the hierarchy that have not been processed.

Instead we should change the "disposeTile" function to check whether "cached" is present before trying to access any of the fields on it.

gkjohnson: I actually see what's happening now - this function is getting called with "immediate = true" after having been called with "immediate=false" previously so it's possible that it exits early when immediate is set to true even though some of the other children are still processing.

I'll merge this for now but it is removing an optimization (all children are now checked every time even though they may all be processed). I'll look into addressing that, though. Thanks again -

PR #1446, opened 2026-01-16
TilesRendererBase: Early out of the child processing

### LINKED ISSUE #1441: TilesRendererBase.ensureChildrenArePreprocessed does not evaluate all children

gkjohnson:
Related to #1442
Fix #1441 

- Early out of child processing if they are already all processed
- Remove immediate processing of children in DebugTilesPlugin disposal / enable toggle
- Make DebugTilesPlugin model disposal robust to unprocessed tiles

cc @Jordan-Lane

### DIFF
--- a/src/core/renderer/tiles/TilesRendererBase.js
+++ b/src/core/renderer/tiles/TilesRendererBase.js
@@ -810,6 +810,12 @@ export class TilesRendererBase {
 	ensureChildrenArePreprocessed( tile, immediate = false ) {
 
 		const children = tile.children;
+		if ( tile.__childrenProcessed === children.length ) {
+
+			return;
+
+		}
+
 		for ( let i = 0, l = children.length; i < l; i ++ ) {
 
 			const child = children[ i ];
--- a/src/three/plugins/DebugTilesPlugin.js
+++ b/src/three/plugins/DebugTilesPlugin.js
@@ -911,21 +911,21 @@ export class DebugTilesPlugin {
 	_onDisposeModel( tile ) {
 
 		const cached = tile.cached;
-		if ( cached.boxHelperGroup ) {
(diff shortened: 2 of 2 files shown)


Jordan-Lane:
Awesome, the use of `__childrenProcessed` is much better. Thanks for fixing this so fast @gkjohnson

PR #1454, opened 2026-01-19
Move implicit tiling fields to an internal field

gkjohnson:
Related to #1375 

Move "implicit tiling" fields to an "implicitTilingDat" tile sub object.


gkjohnson:
cc @AnthonyGlt FYI

PR #1467, opened 2026-01-29
Update babylon-support branch to use GeospatialCamera, es6 imports, module-level GTLF loader function, auto-detect extension, and observables

georginahalpern:
- Update to ES6 Imports
- Support Version >=8.47.2 (where geospatial camera features are stable)
- Swap googlemapsaerial to using geospatialCamera centered at tokyo tower, plus a flyTo animation to zoom in
- Module-level GLTF loader + auto-detect extension
- Observables

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,10 +121,10 @@
     "@react-three/fiber": {
       "optional": true
     },
-    "babylonjs": {
+    "@babylonjs/core": {
       "optional": true
     },
-    "babylonjs-loaders": {
+    "@babylonjs/loaders": {
       "optional": true
     },
     "react": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,13 +48,13 @@
         "three": ">=0.167.0"
       },
       "peerDependenciesMeta": {
+        "@babylonjs/core": {
+          "optional": true
+        },
+        "@babylonjs/loaders": {
+          "optional": true
+        },
         "@react-three/fiber": {
-          "optional": true
-        },
-        "babylonjs": {
-          "optional": true
-        },
-        "babylonjs-loaders": {
           "optional": true
         },
         "react": {


src/babylonjs/renderer/loaders/GLTFLoader.js
```diff
@@ -13,6 +16,58 @@ export class GLTFLoader extends LoaderBase {
...
+
+			return false;
+
+		}
+
+		const view = new DataView( arrayBuffer, byteOffset, byteLength );
+		const magic = view.getUint32( 0, true ); // little-endian
+		return magic === GLB_MAGIC;
+
+	}
```
gkjohnson: From the [other PR](https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/1397):

> Just to be sure: here's no way for Babylon to determine whether a file is a binary or ascii glTF automatically?

Do all apps in Babylon manually check the header of every file to ensure the right loader is used? Or is the file extension used considered enough?

src/babylonjs/renderer/tiles/TilesRenderer.js
```diff
@@ -20,14 +20,61 @@ export class TilesRenderer extends TilesRendererBase {
...
+
+		if ( ! this._observables.has( type ) ) {
+
+			this._observables.set( type, new Observable() );
+
+		}
+
+		return this._observables.get( type );
+
+	}
```
gkjohnson: Small question: is there a reason this isn't done in the "addEventListener" callback? Is this intended to be a public function?

package.json
```diff
@@ -111,8 +111,8 @@
   },
   "peerDependencies": {
     "@react-three/fiber": "^8.17.9 || ^9.0.0",
-    "babylonjs": "^7.0.0",
-    "babylonjs-loaders": "^7.0.0",
+    "@babylonjs/core": ">=8.47.2",
+    "@babylonjs/loaders": ">=8.47.2",
```
gkjohnson: Is there any reason these can't be set to `>=8.0.0`? I know the GeospatialCamera is new but that shouldn't be a requirement for a lot of 3d tiles sets.

src/babylonjs/renderer/loaders/GLTFLoader.js
georginahalpern: in theory the plugin extension should be enough, but i was running into all sorts of issues when testing. turns out my problem was caused by npm linking to my local babylon/loaders version  -- if i remove my link and just use the loader as-is, none of this is necessary :) pushing an update

src/babylonjs/renderer/tiles/TilesRenderer.js
georginahalpern: no good reason, updating!

src/babylonjs/renderer/loaders/GLTFLoader.js
georginahalpern: ah wait i stand corrected - i dont think we can send gltf extension and send data as array buffer. need to modify approach

package.json
georginahalpern: since we're using the camera in the example files (along with useLargeWorldRendering, a newer feature since 8.0) i figure better to just support all geospatial features together. if community pushes back we can revisit ?

gkjohnson: If there are no features required by the library from >8.0.0 I don't think there's a reason to require a higher version minor / patch version. We can use the latest versions in the examples but I'll change the peer dependency requirement to >= 8.0.0 since it seems to work just fine the mars example.

src/babylonjs/renderer/loaders/GLTFLoader.js
gkjohnson: Got this fixed by passing the "rootUrl" into the gltf load function so the loader knows where to load dependencies relative to:

```js
const pluginExtension = extension === 'gltf' ? '.gltf' : '.glb';
const container = await ImportMeshAsync(
	new File( [ buffer ], uri ),
	scene,
	{ pluginExtension, rootUrl }
);
```

PR #1468, opened 2026-01-30
Babylon Support: Store parsed JSON in metadata field so we can access asset.copyright

georginahalpern:
Store the parsed JSON so that tile renderer has access to.metadata.asset.copyright so that we can properly attribute tiles


gkjohnson:
Looks great, thanks!

PR #1469, opened 2026-02-02
Add PMTiles plugin with MVT support

AlaricBaraou:
Minimal PR extracted from this draft https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/990 

Adds support for rendering PMTiles archives on the globe using HTTP range requests.

  ## New files

  **Core:**
  - `MVTLoaderBase.js` - Parses MVT format
  - `PMTilesLoaderBase.js` - PMTiles archive access

  **Three.js:**
  - `PMTilesPlugin.js` - Main plugin
  - `PMTilesImageSource.js` / `MVTImageSource.js` - Image sources
  - `VectorTileCanvasRenderer.js` / `VectorTileStyler.js` - Canvas rendering
  - `layerColors.js` - Default layer styles

  **Demo:**
  - `pmtiles.html` / `pmtiles.js` - Globe example with layer controls

  ## Dependencies

  Adds optional peer dependencies:
  - `pmtiles` - Read tiles from PMTiles archives via range requests
  - `@mapbox/vector-tile` - Parse MVT (Mapbox Vector Tile) format
  - `pbf` - Protocol buffer decoding (required by @mapbox/vector-tile)

### DIFF
--- a/example/three/pmtiles.js
+++ b/example/three/pmtiles.js
@@ -1,0 +1,176 @@
+import {
+	Scene,
+	WebGLRenderer,
+	PerspectiveCamera,
+	AmbientLight,
+	DirectionalLight,
+} from 'three';
+import {
+	TilesRenderer,
+	GlobeControls,
+} from '3d-tiles-renderer';
+import {
+	UpdateOnChangePlugin,
+	PMTilesPlugin,
+} from '3d-tiles-renderer/plugins';
+import { GUI } from 'three/examples/jsm/libs/lil-gui.module.min.js';
+
+let scene, renderer, camera, controls, tiles, gui;
+
+// Layer configuration for Protomaps v4 basemap
+const LAYERS = {
+	water: { enabled: true, color: '#4a90d9' },
+	earth: { enabled: true, color: '#f2efe9' },
+	landuse: { enabled: false, color: '#e8e4d8' },
+	landcover: { enabled: false, color: '#d4e8c2' },
+	natural: { enabled: false, color: '#c8d9af' },
+	roads: { enabled: false, color: '#ffffff' },
+	buildings: { enabled: false, color: '#d9d0c9' },
+	transit: { enabled: false, color: '#888888' },
+	boundaries: { enabled: true, color: '#ff6b6b' },
+	places: { enabled: true, color: '#333333' },
+	pois: { enabled: false, color: '#7d4e24' },
+};
+
+// Application state
+const state = {
+	layers: {},
+	colors: {},
+};
+
+// Initialize state from layer config
+for ( const key in LAYERS ) {
+
+	state.layers[ key ] = LAYERS[ key ].enabled;
+	state.colors[ key ] = LAYERS[ key ].color;
+
+}
+
+state.colors.default = '#cccccc';
+
+init();
+setupGUI();
+createTiles();
+
+function init() {
+
+	renderer = new WebGLRenderer( { antialias: true } );
+	renderer.setAnimationLoop( render );
+	renderer.setPixelRatio( window.devicePixelRatio );
+	renderer.setSize( window.innerWidth, window.innerHeight );
+	renderer.setClearColor( 0x111111 );
+	document.body.appendChild( renderer.domElement );
+
+	scene = new Scene();
+	camera = new PerspectiveCamera( 60, window.innerWidth / window.innerHeight, 100, 1e8 );
+
+	const dirLight = new DirectionalLight( 0xffffff );
+	dirLight.position.set( 1, 1, 1 );
+	scene.add( dirLight );
+	scene.add( new AmbientLight( 0x444444 ) );
+
+	controls = new GlobeControls( scene, camera, renderer.domElement );
+	controls.enableDamping = true;
+	controls.camera.position.set( 0, 0, 1.5 * 1e7 );
+
+	window.addEventListener( 'resize', onWindowResize, false );
+
+}
+
+function createFilter() {
+
+	return function ( feature, layerName ) {
+
+		if ( layerName in state.layers ) {
+
+			return state.layers[ layerName ] === true;
+
+		}
+
+		// Unknown layers: hide by default
+		return false;
+
+	};
+
+}
+
+function createTiles() {
+
+	if ( tiles ) {
+
+		scene.remove( tiles.group );
+		tiles.dispose();
+
+	}
+
+	tiles = new TilesRenderer();
+	tiles.registerPlugin( new UpdateOnChangePlugin() );
+	tiles.registerPlugin( new PMTilesPlugin( {
+		url: 'https://demo-bucket.protomaps.com/v4.pmtiles',
+		center: true,
+		shape: 'ellipsoid',
+		levels: 15,
+		tileDimension: 512,
+		styles: state.colors,
+		filter: createFilter()
+	} ) );
+
+	tiles.group.rotation.x = - Math.PI / 2;
+	tiles.setCamera( camera );
+	scene.add( tiles.group );
+
+	if ( controls ) controls.setEllipsoid( tiles.ellipsoid, tiles.group );
+
+}
+
+function setupGUI() {
+
+	gui = new GUI();
+
+	// Layers folder
+	const layersFolder = gui.addFolder( 'Layers' );
+	for ( const key in LAYERS ) {
+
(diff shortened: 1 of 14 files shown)


src/core/renderer/loaders/PMTilesLoaderBase.js
```diff
@@ -0,0 +1,70 @@
+// PMTiles Archive Format
+// https://github.com/protomaps/PMTiles
+
+import { PMTiles } from 'pmtiles';
+
+export class PMTilesLoaderBase {
```
gkjohnson, line 6: I think we can move anything from these "LoaderBase" classes to "ImageSource". They should only really be used in that one location and they're pretty small.

src/core/renderer/loaders/PMTilesLoaderBase.js
```diff
@@ -0,0 +1,70 @@
...
+		const i1 = url.lastIndexOf( '/', i2 - 1 );
+		const i0 = url.lastIndexOf( '/', i1 - 1 );
+
+		return {
+			z: parseInt( url.slice( i0 + 1, i1 ) ),
+			x: parseInt( url.slice( i1 + 1, i2 ) ),
+			y: parseInt( url.slice( i2 + 1 ) ),
+		};
+
+	}
```
gkjohnson, line 68: This logic I think should be in the plugin since it's specific to how the plugin is interpreting and handling tile urls, which the loaders and image sources shouldn't have to know about (they're tiles agnostic).

src/three/renderer/utils/VectorTileCanvasRenderer.js
```diff
@@ -0,0 +1,176 @@
...
+				if ( k === 0 ) ctx.moveTo( p.x * scale, p.y * scale );
+				else ctx.lineTo( p.x * scale, p.y * scale );
+
+			}
+
+			ctx.closePath();
+
+		}
+
+		ctx.fill();
```
gkjohnson, line 172: In order to support holes in the polygons you need to specify the "evenodd" rule - otherwise I think you'll just get solid shapes. See the [geojson implementation](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/87dbf597e1d1dd83819361c1c24d92b84e087992/src/three/plugins/images/sources/GeoJSONImageSource.js#L445-L470).

An aside but I'm wondering if we can reuse this in the GeoJSON overlay since these shapes are all about the same 🤔

src/three/renderer/utils/VectorTileCanvasRenderer.js
```diff
@@ -0,0 +1,176 @@
...
+		tex.colorSpace = SRGBColorSpace;
+		tex.generateMipmaps = false;
+		tex.needsUpdate = true;
+		return tex;
+
+	}
+
+	_renderPoints( ctx, geometry, layerName, scale ) {
+
+		const isLabelLayer = ( layerName === 'place_label' );
```
gkjohnson, line 108: What is place_label here? And why is it preventing drawing points specifically?

src/three/renderer/utils/VectorTileCanvasRenderer.js
```diff
@@ -0,0 +1,176 @@
...
+
+		for ( const multiPoint of geometry ) {
+
+			for ( const p of multiPoint ) {
+
+				const x = p.x * scale;
+				const y = p.y * scale;
+
+				if ( ! isLabelLayer ) {
```
gkjohnson, line 117: If this is blocking all point drawing then I think we can remove the label layer check from this class and just skip calling the "_renderPoints" function from the call site.

src/three/renderer/utils/VectorTileCanvasRenderer.js
```diff
@@ -0,0 +1,176 @@
...
+
+					const radius = ( layerName === 'poi' ) ? 3 : 2;
+
+					ctx.beginPath();
+					ctx.moveTo( x + radius, y );
+					ctx.arc( x, y, radius, 0, Math.PI * 2 );
+					ctx.fill();
```
gkjohnson, line 124: Since the pixel ratios are not consistent over the surface of the globe just drawing a circle to the canvas will result in something that looks like an ellipsoid rather than a circle on the surface. The geojson layer [scales the points by the aspect ratio at that point](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/87dbf597e1d1dd83819361c1c24d92b84e087992/src/three/plugins/images/sources/GeoJSONImageSource.js#L363-L373) derived from the derivative of the lat / lon the given point.

You can see from the comments there, though, that it could still be improved a bit. So we can skip this for now but FYI.

See #1325 for an illustration of the issue and fix.

src/three/renderer/utils/VectorTileCanvasRenderer.js
```diff
@@ -0,0 +1,176 @@
...
+		this.styler = styler;
+		this.tileDimension = options.tileDimension || 512;
+
+	}
+
+	render( vectorTile ) {
+
+		const canvas = this._createCanvas( this.tileDimension, this.tileDimension );
+		const ctx = canvas.getContext( '2d' );
+		const scale = this.tileDimension / MVT_EXTENT;
```
gkjohnson, line 18: Can we add some comments for some of these things - what is MVT_EXTENT? Why is it hardcoded to a 4k size? I'm wondering how much of the vector tiles-specific stuff we can move to the plugin and make this a more reusable class for use with GeoJSON plugin, as well

src/three/renderer/utils/VectorTileStyler.js
```diff
@@ -0,0 +1,54 @@
...
+
+			_color.set( colorsToSet[ key ] );
+			this._styles[ key ] = {
+				hex: _color.getHex(),
+				css: _color.getStyle()
+			};
+
+		}
+
+	}
```
gkjohnson, line 25: Can you give me a quick overview of how "features" and "styles" are handled in MVT and PMTiles? Does every shape etc only have a single type or class? Or is it possible for there to be multiple attributes associated with a single shape? As in `"building", "city"` and `"building", "rural"`.

src/three/plugins/images/PMTilesPlugin.js
```diff
@@ -0,0 +1,32 @@
...
+			const { z, x, y } = PMTilesLoaderBase.parseUrl( url );
+
+			return this.imageSource.pmtilesLoader.getTile( z, x, y, options?.signal )
+				.then( buffer => buffer || new ArrayBuffer( 0 ) );
+
+		}
+
+		return null;
+
+	}
```
gkjohnson, line 30: We should just be able to implement "fetchItem" here, as we do for "GeoJSONImageSource", which will get called with `x, y, z, level` - then we shouldn't have to deal with custom urls or anything.

example/three/pmtiles.js
```diff
@@ -0,0 +1,176 @@
...
+
+	gui = new GUI();
+
+	// Layers folder
+	const layersFolder = gui.addFolder( 'Layers' );
+	for ( const key in LAYERS ) {
+
+		layersFolder.add( state.layers, key )
+			.name( key.charAt( 0 ).toUpperCase() + key.slice( 1 ) )
+			.onChange( createTiles );
```
gkjohnson, line 136: The GeoJSONImageSource includes a "redraw" function to rerender every texture in place in case something changes, allowing for styling updates, animations, etc. I'm thinking we should allow the same thing here rather than rebuilding the full set of tiles.

gkjohnson (commented):
This looks really great. It's feeling like a lot of this code could be shared with the GeoJSONImageSource for a general "vector overlay" system that can allow for drawing these kinds of shapes. We can look into that refactor after.

One architectural difference I would recommend is sticking more closely with the GeoJSONImageSource approach (using RegionImageSource) rather than XYZImageSource (which uses TiledImageSource). The difference in the way that these are being used is that the XYZ image plugin will draw / load images for every individual tile and then these tiles will be "composed" into a large texture covering the necessary region for overlays. This means we have texture memory for the original tiles and for the composed ones.

What GeoJSONImageSource is doing is drawing directly to those final region textures which allows for quick updates and redraws for styling changes. Explaining the architecture and how to implement this isn't so straight forward so maybe I can help with this if needed one things are closer to being merged.

gkjohnson:
Closing in favor of #1575

PR #1483, opened 2026-02-24
Add search functionality to babylon demo site

georginahalpern:
image

- Adds a collapsible 'navigation' panel which includes instructions for camera navigation as well as the ability to jumpto a lat/long/alt or ECEF XYZradius point 
- Also parses latlonalt from URL
- using nominatim.openstreetmap.org and corsproxy.io to avoid initial cors error with first search

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - LoadRegionPlugin: Add typings for new function, fix bug related to "distance" calculations.
 - GlobeContorls: Account for ellipsoid frame scale when calculating max zoom out distance.
 - Babylon TilesRenderer: Adjusted imports to support treeshaking.
+- ImageOverlayPlugin: Fix case where removing all overlays does not clear overlay textures correctly.
 
 ## [0.4.21] - 2026-02-04
 ### Fixed
--- a/src/three/plugins/images/ImageOverlayPlugin.js
+++ b/src/three/plugins/images/ImageOverlayPlugin.js
@@ -1135,7 +1135,7 @@ export class ImageOverlayPlugin {
 
 	_updateLayers( tile ) {
 
-		const { overlayInfo, overlays, tileControllers } = this;
+		const { overlayInfo, overlays, tileControllers, meshParams } = this;
 		const tileController = tileControllers.get( tile );
 
 		// by this point all targets should be present and we can force the memory to update
@@ -1143,6 +1143,34 @@ export class ImageOverlayPlugin {
 
 		// if the tile has been disposed before this function is called then exit early
 		if ( ! tileController || tileController.signal.aborted ) {
+
+			return;
+
+		}
+
+		// handle the case where all overlays have been removed - we need to reset
+		// the materials to have no layers
+		if ( overlays.length === 0 ) {
+
+			const scene = tile.engineData && tile.engineData.scene;
+			if ( scene ) {
+
+				scene.traverse( c => {
+
+					if ( c.material && meshParams.has( c ) ) {
+
+						const params = meshParams.get( c );
+						params.layerMaps.length = 0;
+						params.layerInfo.length = 0;
+
+						c.material.defines.LAYER_COUNT = 0;
+						c.material.needsUpdate = true;
+
+					}
+
+				} );
+
+			}
 
 			return;
 
@@ -1156,7 +1184,7 @@ export class ImageOverlayPlugin {
 			meshInfo.forEach( ( { attribute }, mesh ) => {
 
 				const { geometry, material } = mesh;
-				const params = this.meshParams.get( mesh );
+				const params = meshParams.get( mesh );
 
 				// assign the new uvs
 				const key = `layer_uv_${ i }`;


example/babylonjs/googleMapsAerial.js
```diff
@@ -109,3 +176,191 @@ window.addEventListener( 'resize', () => {
...
 
 } );
+
+// --- Place search functionality ---
+const placeSearch = document.getElementById( 'place-search' );
+const searchBtn = document.getElementById( 'search-btn' );
+const searchResult = document.getElementById( 'search-result' );
+const searchResultText = document.getElementById( 'search-result-text' );
+const togglePanelBtn = document.getElementById( 'toggle-panel' );
+const panelContent = document.getElementById( 'panel-content' );
```
gkjohnson: Can we include these in the existing gui dropdown rather than adding new elements and styles?

example/babylonjs/googleMapsAerial.js
```diff
@@ -8,6 +8,51 @@ const GOOGLE_TILES_ASSET_ID = 2275207;
...
+		const N = WGS84_A / Math.sqrt( 1 - WGS84_E2 * sinLat * sinLat );
+		lat = Math.atan2( z + WGS84_E2 * N * sinLat, p );
+
+	}
+
+	const sinLat = Math.sin( lat );
+	const N = WGS84_A / Math.sqrt( 1 - WGS84_E2 * sinLat * sinLat );
+	const alt = p / Math.cos( lat ) - N;
+
+	return [ ( lat * 180 ) / Math.PI, ( lon * 180 ) / Math.PI, alt ];
```
gkjohnson: Are there plans to include this functionality in Babylon? The functionality for this is included in the three.js "Ellipsoid" class within the project already.

gkjohnson (commented):
Thanks - I think my initial feedback here is I'm wondering if this can be simplified down to just a single "search" field that jumps the camera to the new location. I think there's value to keeping these example simple where possible so users can understand how to set up the basics of the tiles renderer.

I'm also seeing that the search positions the camera under the terrain pretty frequently so I'm feeling like that should be addressed, as well.

example/babylonjs/googleMapsAerial.js
georginahalpern: yes! see comment about upstreaming to babylon. plan is to introduce these conversion functions and ellipsoidal concept to our library, it just might have to happen after we lift code freeze for 9.0 release. but i will take point on updating here to reference upstream functions once available

example/babylonjs/googleMapsAerial.js
georginahalpern: updated!

georginahalpern:
> Thanks - I think my initial feedback here is I'm wondering if this can be simplified down to just a single "search" field that jumps the camera to the new location. I think there's value to keeping these example simple where possible so users can understand how to set up the basics of the tiles renderer.
> 
> I'm also seeing that the search positions the camera under the terrain pretty frequently so I'm feeling like that should be addressed, as well.

enabled collision detection to avoid navigating underneath terrain. for search, used a new API to grab terain elevation at a given lat/lon. i imagine this can also be grabbed from the tiles thesmselves, could happen in a later change

src/babylonjs/renderer/index.d.ts
```diff
@@ -4,6 +4,12 @@ import { Scene, TransformNode } from '@babylonjs/core';
 export class TilesRenderer extends TilesRendererBase {
 
 	group: TransformNode;
+	readonly visibleTiles: Set<object>;
+	readonly activeTiles: Set<object>;
+
```
gkjohnson: All the type changes here can be added to the "TilesRendererBase" d.ts file.

src/babylonjs/renderer/tiles/TilesRenderer.js
```diff
@@ -206,7 +206,7 @@ export class TilesRenderer extends TilesRendererBase {
 		super.disposeTile( tile );
 
 		const engineData = tile.engineData;
-		if ( engineData.container ) {
+		if ( engineData.container && typeof engineData.container.dispose === 'function' ) {
```
gkjohnson: This looks left over from the other PR

example/babylonjs/googleMapsAerial.js
```diff
@@ -67,6 +125,24 @@ tiles.registerPlugin( new CesiumIonAuthPlugin( {
...
 tiles.errorTarget = params.errorTarget;
 
+camera.checkCollisions = true;
+// Enable collisions on tile meshes as they load
+tiles.addEventListener( 'load-model', ( event ) => {
+
+	const tileScene = event?.scene;
+	if ( tileScene ) {
+
+		const meshes = tileScene.getChildMeshes?.() ?? [];
```
gkjohnson: Is this overly defensive? It looks like this function will [always be available](https://doc.babylonjs.com/typedoc/classes/BABYLON.Node#getchildmeshes) if the scene is present. Also if "load-model" has been called then the scene object should always be available. Are you seeing otherwise?

example/babylonjs/googleMapsAerial.js
```diff
@@ -67,6 +125,24 @@ tiles.registerPlugin( new CesiumIonAuthPlugin( {
 } ) );
 tiles.errorTarget = params.errorTarget;
 
+camera.checkCollisions = true;
+// Enable collisions on tile meshes as they load
+tiles.addEventListener( 'load-model', ( event ) => {
+
+	const tileScene = event?.scene;
```
gkjohnson: Likewise the "event" object will never be null or undefined here.

example/babylonjs/googleMapsAerial.html
```diff
@@ -3,19 +3,31 @@
...
     <link rel="stylesheet" href="../styles.css">
+    <style>
+        .lil-gui.root {
+            --width: 350px;
+            position: absolute;
+            top: 10px;
+            left: 10px;
+            right: auto;
+        }
+    </style>
```
gkjohnson: What's the purpose of these added styles? Why offset the position of the lil-gui root?

If we want to adjust the width we should use one of the approaches recommended in the [lil-gui docs](https://lil-gui.georgealways.com/#API) - which is either styling for the ".lil-gui" class (no ".root") or pasing it into the constructor. I'd prefer the constructor so everything relating to the gui is in one place.

example/babylonjs/googleMapsAerial.js
georginahalpern: yeah i wonder if instead it makes more sense to do the setting of collision on the mesh at load time, vs having to iterate over the meshes in the scene to set them as collidable? perhaps a param on the loader itself ?

gkjohnson: This makes sense - I think there's value in effectively being able to treat all meshes in the "TilesRenderer" instance a collective whole. The three.js version of the tiles renderer includes a "autoDisableRendererCulling" flag to toggle per-mesh frustum culling for all tiles since the tiles renderer is already doing that, for example. Maybe the same pattern makes sense here.

src/babylonjs/renderer/index.d.ts
georginahalpern: updated

example/babylonjs/googleMapsAerial.js
georginahalpern: updated!

example/babylonjs/googleMapsAerial.html
georginahalpern: removed!

src/babylonjs/renderer/tiles/TilesRenderer.js
```diff
@@ -199,6 +200,16 @@ export class TilesRenderer extends TilesRendererBase {
...
+		if ( this.checkCollisions ) {
+
+			for ( const mesh of scene.getChildMeshes() ) {
+
+				mesh.checkCollisions = true;
+
+			}
+
+		}
+
```
gkjohnson, line 212: I'll merge this as-is for now but one thing to note is that if a user changes this flag after models have loaded then the model will be in a state where some models have "checkCollisions" set to true and others do not. The "autoDisableRendererCulling" will iterate over all loaded models on change to ensure the state is consistent so it may be worth considering doing the same here:

https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/f830d08895c4eb24ca76c9860da0e98205975c91/src/three/renderer/tiles/TilesRenderer.js#L47-L66

Also not sure if this flag should be added to the docs.

Issue #1479, opened 2026-02-11
ImageOverlayPlugin: Deleting the last overlay does not reset texture layers


TQtong:
### Question

When I attempt to change the overlay’s visibility using the visible property, I notice it has no effect.

### Supplemental Data

https://jsfiddle.net/tangqi/fhzuvnL1/9/

### Library Version

0.4.21

### Three.js Version

0.170.0

gkjohnson:
> imageryOverlay.visible = value

"visible" is not a documented or used property for image overlays anywhere the project. Please make sure you are using valid properties.

You can hide an overlay by setting the opacity to 0.

TQtong:
<img width="1992" height="437" alt="Image" src="https://github.com/user-attachments/assets/e5a68dcf-23ab-4ba4-a3be-02491f7822d4" />

I noticed that even when the opacity is set to 0, network requests still occur.

gkjohnson:
If you'd like the overlay to stop loading then you need to remove it from the ImageOverlayPlugin.

TQtong:
Thank you very much, now I understand.

TQtong:
https://jsfiddle.net/tangqi/fhzuvnL1/23/

I noticed that when I used the same method to remove the GeoJSON, the GeoJSON-rendered content still persisted in the scene.

<img width="383" height="321" alt="Image" src="https://github.com/user-attachments/assets/010b1609-65fd-4efd-af49-fe9c0ff9fb53" />

When I zoom in or out with the mouse, the GeoJSON toggles its visibility accordingly.

gkjohnson:
Thanks for the repro - it looks like a bug related to deleting an overlay when there is only one remaining in the overlay plugin. I'll look into addressing it for the next release.

TQtong:
Thank you so much, this is the solution I used to fix this bug. I tested it locally and it works fine.

# deleteOverlay Rendering Issue When Removing Last Overlay

## Problem Description

When using the `deleteOverlay` method of `ImageOverlayPlugin` to remove the last overlay, the deleted overlay continues to render in the scene, even though it has been removed from the `overlays` array.

## Steps to Reproduce

1. Add multiple overlays (e.g., `CesiumIonOverlay` and `GeoJSONOverlay`)
2. Delete overlays one by one
3. When deleting the last overlay, its texture still displays in the scene

```javascript
// Example code
[...]

PR #1485, opened 2026-02-26
Fix the issue where it still renders after deleting the last overlay

TQtong:
Fix the issue where it still renders after deleting the last overlay #1479


gkjohnson:
Looks good, thanks!

TQtong:
You are welcome😁

PR #1486, opened 2026-02-26
Babylon: Import from subdirectory vs babylon/core. Fix container.dispose error

georginahalpern:
Enables proper es6 treeshaking

### DIFF
--- a/src/babylonjs/renderer/loaders/GLTFLoader.js
+++ b/src/babylonjs/renderer/loaders/GLTFLoader.js
@@ -1,6 +1,6 @@
 import { LoaderBase } from '3d-tiles-renderer/core';
 import { Matrix, Quaternion } from '@babylonjs/core/Maths/math.vector';
-import { ImportMeshAsync } from '@babylonjs/core/Loading/sceneLoader';
+import { LoadAssetContainerAsync } from '@babylonjs/core/Loading/sceneLoader';
 import '@babylonjs/loaders/glTF/2.0';
 
 const _worldMatrix = /* @__PURE__ */ Matrix.Identity();
@@ -31,7 +31,7 @@ export class GLTFLoader extends LoaderBase {
 		// load the file
 		const pluginExtension = extension === 'gltf' ? '.gltf' : '.glb';
 		let metadata = null;
-		const container = await ImportMeshAsync(
+		const container = await LoadAssetContainerAsync(
 			new File( [ buffer ], uri ),
 			scene,
 			{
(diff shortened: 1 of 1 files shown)


gkjohnson:
Can you explain the necessity of this? [Parcel](https://parceljs.org/features/code-splitting/#tree-shaking), [Webpack](https://webpack.js.org/guides/tree-shaking/), and [Rollup](https://rollupjs.org/configuration-options/#treeshake) (and by extension Vite) all support culling unused exports from files via tree-shaking assuming the projects are set up correctly. Is Babylon doing something that's breaking the ability to tree shake?

src/babylonjs/renderer/tiles/TilesRenderer.js
```diff
@@ -188,7 +192,7 @@ export class TilesRenderer extends TilesRendererBase {
 			.decompose( scene.scaling, scene.rotationQuaternion, scene.position );
 
 		// exit early if a new request has already started
-		if ( abortSignal.aborted ) {
+		if ( abortSignal.aborted && typeof result.container.dispose === 'function' ) {
```
gkjohnson: Why is this necessary? Assuming this is a Babylon AssetContainer, the class seems to [always have a dispose function](https://doc.babylonjs.com/typedoc/classes/BABYLON.AssetContainer#dispose). And even if it doesn't and this condition fails then it will cause the function to incorrectly continue and save the scene, container, and metadata to the engineData object even though the request was aborted.

georginahalpern: oops this was leftover from a (bad) copilot suggestion

if i navigate around the demo a bunch i run into console error

```
Uncaught TypeError: t.container.dispose is not a function
    at aCe.disposeTile (index-DaIn0mBc.js:28587:229324)
    at index-DaIn0mBc.js:463:213764
    at aCe.invokeAllPlugins (index-DaIn0mBc.js:463:205230)
    at index-DaIn0mBc.js:463:213726
    at index-DaIn0mBc.js:463:188826
    at Array.forEach (<anonymous>)
    at G8.unloadUnusedContent (index-DaIn0mBc.js:463:188776)
    at index-DaIn0mBc.js:463:189130

```


Image

but thats because the dispose fucntion doesnt exist on container , working on a fix !

georginahalpern:
> Can you explain the necessity of this? [Parcel](https://parceljs.org/features/code-splitting/#tree-shaking), [Webpack](https://webpack.js.org/guides/tree-shaking/), and [Rollup](https://rollupjs.org/configuration-options/#treeshake) (and by extension Vite) all support culling unused exports from files via tree-shaking assuming the projects are set up correctly. Is Babylon doing something that's breaking the ability to tree shake?

Unfortunately today Babylon is not tree-shaking friendly -- when u import from core you import the whole framework because of how index.ts is implemented :( i believe it had a lot to do with keeping backwards compatability --
 -- more details here https://doc.babylonjs.com/setup/frameworkPackages/es6Support/
 
 the work to fix this was supposed to be in 9.0 but is now slotted for 10. but for now the solution is to import from the subdirectory

src/babylonjs/renderer/tiles/TilesRenderer.js
georginahalpern: okay pushed a fix!

PR #1489, opened 2026-02-28
DebugTilesPlugin: add boundsColorMode for independent bounds coloring

### LINKED ISSUE #1488: DebugTilesPlugin: color leaf tile bounding boxes in white
## Problem

The `DebugTilesPlugin` renders bounding box wireframes using a random color based on tile depth (`getIndexedRandomColor(tile.internal.depth)`). There is no visual distinction between leaf and non-leaf tile bounding boxes, making it harder to debug tile hierarchy visually.

## Proposed Solution

Color structural leaf tile bounding boxes (`!tile.children || tile.children.length === 0`) in white, while non-leaf tiles keep their existing depth-based random color. This provides a clear visual code for identifying leaves during debug visualization.

## Implementation

Extend `_updateHelperMaterials()` in `DebugTilesPlugin.js` to set bounding box helper colors based on structural leaf status. This is the natural location since the method already runs every frame on visible helpers and handles material opacity.

### Changes (single file: `src/three/plugins/DebugTilesPlugin.js`)

1. Add a `_white` color constant at module scope
2. Add color update logic in `_updateHelperMaterials()` — white for leaves, depth-based color for non-leaves

No new public API, no `.d.ts` changes, no new configuration options.

## Verification

- `npm run lint` passes
- `npm test` passes
- Visual check: leaf bounding boxes render white, non-leaf retain depth-based colors

diegomazala:
## Summary

Adds a `boundsColorMode` property to `DebugTilesPlugin` that accepts all the same
  options as `colorMode`, but applies them to bounding volume helpers (box, sphere,
  region) instead of tile materials.

  - Shared `applyColor()` function handles color logic for both `colorMode` and
    `boundsColorMode`, avoiding redundancy
  - Defaults to `NONE` — existing depth-based random colors preserved
  - No breaking changes, no new dependencies

  Closes #1488. 

  ## Test plan

  - [x] `npm test` passes (1177 tests)
  - [ ] Visual: set `boundsColorMode` to various modes with `displayBoxBounds: true`. Bounds reflect chosen color mode, tile materials remain untouched

### DIFF
--- a/src/three/plugins/DebugTilesPlugin.js
+++ b/src/three/plugins/DebugTilesPlugin.js
@@ -43,6 +43,7 @@ const RANDOM_COLOR = 7;
 const RANDOM_NODE_COLOR = 8;
 const CUSTOM_COLOR = 9;
 const LOAD_ORDER = 10;
+const INDEXED_COLOR = 11;
 
 const ColorModes = Object.freeze( {
 	NONE,
@@ -111,12 +112,7 @@ export class DebugTilesPlugin {
 		if ( v !== this._boundsColorMode ) {
 
 			this._boundsColorMode = v;
-
-			if ( v === NONE ) {
-
-				this._resetBoundsColors();
-
-			}
+			this.materialsNeedUpdate = true;
 
 		}
 
@@ -598,6 +594,15 @@ export class DebugTilesPlugin {
 
 				}
 
+				case INDEXED_COLOR: {
+
+					child.material.color.copy( getIndexedRandomColor( tile.internal.depth ) );
+					delete child.material[ HAS_RANDOM_COLOR ];
+					delete child.material[ HAS_RANDOM_NODE_COLOR ];
+					break;
+
+				}
+
 			}
 
 		};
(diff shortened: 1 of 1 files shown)


src/three/plugins/DebugTilesPlugin.js
```diff
@@ -100,6 +100,28 @@ export class DebugTilesPlugin {
...
+
+		if ( v !== this._boundsColorMode ) {
+
+			this._boundsColorMode = v;
+
+			if ( v === NONE ) {
+
+				this._resetBoundsColors();
+
+			}
```
gkjohnson: Why not just use the same pattern as above and set "materialsNeedUpdate" to true?

src/three/plugins/DebugTilesPlugin.js
```diff
@@ -439,172 +464,257 @@ export class DebugTilesPlugin {
...
+				}
 
-							break;
+			} );
 
-						}
+		} );
+
+		// update bounds helper colors
+		if ( boundsColorMode !== NONE ) {
```
gkjohnson: Can we just create a special INDEXED_COLOR flag that we can pass into the "applyColor" function when NONE is set rather than relying on on "_resetBoundsColors"? Then we can get rid of that function.

src/three/plugins/DebugTilesPlugin.js
```diff
@@ -439,172 +464,257 @@ export class DebugTilesPlugin {
...
+
+			const groups = [ this.boxGroup, this.sphereGroup, this.regionGroup ];
+			for ( const group of groups ) {
 
-						case CUSTOM_COLOR: {
+				for ( const helper of group.children ) {
 
-							if ( this.customColorCallback ) {
+					const tile = helper.userData.tile;
+					if ( ! tile ) continue;
```
gkjohnson: What's the situation where this will not be present?

src/three/plugins/DebugTilesPlugin.js
diegomazala: Done. The setter now follows the same pattern as `colorMode` and `unlit`. No special-case logic needed since the bounds color loop in `update()` runs unconditionally now.

src/three/plugins/DebugTilesPlugin.js
diegomazala: Added an `INDEXED_COLOR` case to `applyColor` and mapped `boundsColorMode === NONE` to it. The bounds color loop now runs unconditionally through the same path. Removed `_resetBoundsColors`.

src/three/plugins/DebugTilesPlugin.js
diegomazala: It shouldn't happen. `userData.tile` is always set in `_createBoundHelper`. Removed the guard

src/three/plugins/DebugTilesPlugin.js
gkjohnson: Are these commits pushed? I'm not seeing it in the diff

diegomazala: Sorry. My bad. Pushed now

gkjohnson:
Looks great, thanks!

PR #1497, opened 2026-03-05
Add spark.js support to three.js 3DTilesRenderer

castano:
This is just a proof of concept to demonstrate spark.js integration in the 3DTilesRenderer. For more details see this blog post:

https://www.ludicon.com/castano/blog/2026/03/announcing-spark-js-0-1/


gkjohnson:
Hello! Can you provide some more context for this PR?

castano:
This is just a proof of concept to demonstrate spark.js integration in the 3DTilesRenderer. You can find more details in this blog post: 

https://www.ludicon.com/castano/blog/2026/03/announcing-spark-js-0-1/

If there's interest in considering this for integration I'd be happy to to add more details and submit it for review.

src/three/renderer/utils/MemoryUtils.js
```diff
@@ -9,6 +9,13 @@ export function getTextureByteLength( tex ) {
 
 	}
 
+	// IC: The code below assumes tex was created from an ImageBitmap
+	if ( tex instanceof ExternalTexture && tex.userData?.byteLength ) {
+
+		return tex.userData.byteLength;
+
+	}
```
gkjohnson, line 17: What happens if an "ExternalTexture" is provided _without_ userData.byteLength? It looks like "image" will be "null" on the Texture, then and the follow code will crash, right? Assuming we can't get the actual texture size from the ExternalTexture handle we should figure out a reasonable default behavior when byteLength isn't provided.

src/three/renderer/utils/MemoryUtils.js
```diff
@@ -9,6 +9,13 @@ export function getTextureByteLength( tex ) {
 
 	}
 
+	// IC: The code below assumes tex was created from an ImageBitmap
```
gkjohnson, line 12: Small question: what does "IC" mean here?

src/three/plugins/GLTFExtensionsPlugin.js
```diff
@@ -17,6 +18,8 @@ export class GLTFExtensionsPlugin {
 			ktxLoader: null,
 			meshoptDecoder: null,
 			autoDispose: true,
+			spark: null,
+			sparkOptions: null,
```
gkjohnson: I would prefer to not add library specific hooks to the plugin, I think. Is the existing "plugins" array that will get initialized on the GLTFLoader not sufficient for integrating Spark?

gkjohnson (commented):
> You can find more details in this blog post

Thanks, this is helpful. Just to make sure I understand: it sounds like spark allows for textures to be transcoded at run time to memory-efficient, GPU-compatible formats, allowing for formats like webp (small on disk footprint) to be used for download while gpu-optimized formats are used at run time (small in-memory footprint), is that right?

I see there are screenshots showing increased geometric detail with spark, implying reduced memory usage so more tiles can fit in the cache, but it would be helpful to see a before / after table detailing comparisons of on-disk size, in-memory size, and additional parse time required due to transcoding from the library so the tradeoffs are clear.

src/three/renderer/utils/MemoryUtils.js
castano: My initials: Ignacio Castano.

src/three/renderer/utils/MemoryUtils.js
castano: The code will crash with ExternalTextures generated by other tools. It also won't compute the size of CompressedTextures correctly, as the generateMipmaps parameter is forced to false, but textures may still have mipmaps. I'd be happy to robustify this code. I was aiming for minimal changes here.

src/three/plugins/GLTFExtensionsPlugin.js
castano: The problem is that in order to handle AVIF and WebP images, we needs to be able to not only add new plugins, but remove or replace existing ones. 

See the implementation of `registerSparkLoader` here:

https://github.com/Ludicon/spark.js/blob/0be8d264f0b09667cec6ce84f014e5575ae41b7a/src/three-gltf.js#L219

Another option would be to have the caller explicitly register the spark loader:

```
	if ( params.enableSpark ) {
		registerSparkLoader( gltfPlugin._loader, spark, { preferLowQuality: params.preferLowQuality, generateMipmaps: params.generateMipmaps });
	}
```

but that means directly accessing `_loader` attribute and I'm assuming the `_` signals that it's intended to be a private.

This is arguably a problem in three.js. My expectation was also that registering a loader plugin would replace the existing one, but instead it's registered and never used, because the first one takes precedence.

castano:
> Thanks, this is helpful. Just to make sure I understand: it sounds like spark allows for textures to be transcoded at run time to memory-efficient, GPU-compatible formats at run time, allowing for formats like webp (small on disk footprint) to be used for download while gpu-optimized formats are used at run time (small in-memory footprint), is that right?

That's exactly right.

> I see there are screenshots showing increased geometric detail with spark, implying reduced memory usage so more tiles can fit in the cache, but it would be helpful to see a before / after table detailing comparisons of on-disk size, in-memory size, and additional parse time required due to transcoding from the library so the tradeoffs are clear.

The overhead of spark.js is fairly low, because transcoding happens on the GPU and the codecs are extremely fast. The bottleneck is usually in the image decoding on the CPU, which is orders of magnitude slower. 

I've provided some numbers in previous blog posts, for example, here's a size comparison of the sponza scene:

https://www.ludicon.com/castano/blog/2026/02/an-updated-sponza-gltf/

On that test, enabling or disabling spark did not affect the loading time in a perceptible way.

Here are some more numbers from an earlier release:

https://www.ludicon.com/castano/blog/2025/09/three-js-spark-js/

I should note that spark compression will increase loading time, because the renderer will load many more tiles, so it will increase bandwidth use, but if that's a concern it's possible to control that with the `errorTarget` parameter.

src/three/renderer/utils/MemoryUtils.js
gkjohnson: Making it more robust would be great. I expect it's not possible to determine the actual in-memory size of one of these textures from just the WebGLTexture (or WebGPU) handle itself, right? At least not without the gl contenxt. In lieu of any other information being available for calculating the size I was thinking we could check for a custom field (as you are here) and otherwise fallback to a fixed memory size of a 64x64 texture or something just so the value isn't "0" in the cache.

src/three/plugins/GLTFExtensionsPlugin.js
gkjohnson: > This is arguably a problem in three.js. My expectation was also that registering a loader plugin would replace the existing one, but instead it's registered and never used, because the first one takes precedence.

I see - having to delete plugins in gltfLoader (and needing to instantiate them to detect the name no less) doesn't seem to so clean. Is there an existing three.js issue for this? I agree that ideally any user-registered plugins should take priority over the built-in ones.

We could make the "loader" field public but that doesn't feel great, either. I think improving the GLTFLoader feels like the best option.

gkjohnson:
> I should note that spark compression will increase loading time, because the renderer will load many more tiles, so it will increase bandwidth use, but if that's a concern it's possible to control that with the errorTarget parameter.

The max memory limit in the LRUCache shouldn't really typically be limiting the tiles that the renderer is determining to be needed unless "errorTarget" is specifically being picked to create the issue. I assume you were reducing the "errorTarget" value (or lru cache memory cap) to a point where there amount of tiles loaded was being limited by memory. This shouldn't be a typical problem, though - noting that it's still a benefit to have more memory overhead where possible. Am I misunderstanding?

src/three/renderer/utils/MemoryUtils.js
castano: I'll propose that change in a separate PR. Here I just wanted to keep it simple to demonstrate that the changes required for integration are minimal.

src/three/plugins/GLTFExtensionsPlugin.js
castano: I should certainly have done that. Thanks for calling me out. I'll go ahead, open a three.js issue, and propose a fix. It'd be great if this change wasn't necessary in the future.

castano:
> The max memory limit in the LRUCache shouldn't really typically be limiting the tiles that the renderer is determining to be needed unless "errorTarget" is specifically being picked to create the issue. I assume you were reducing the "errorTarget" value (or lru cache memory cap) to a point where there amount of tiles loaded was being limited by memory. This shouldn't be a typical problem, though - noting that it's still a benefit to have more memory overhead where possible. Am I misunderstanding?

That's not what I see on my end. At errorTarget=16 I often see loading limited by the memory cap. With spark.js the cap is usually not reached, which is why the results have greater detail. Lowering the errorTarget certainly makes the difference more stark.

castano:
For example, in the following screenshot memory caps at 275 MB:
Screenshot 2026-03-10 at 1 27 19 PM

With spark enabled, memory use goes down to 113 MB, even though detail is noticeably higher:
Screenshot 2026-03-10 at 1 27 02 PM

gkjohnson:
> That's not what I see on my end. At errorTarget=16 I often see loading limited by the memory cap.

You're right, I was misremembering. I'm not reaching the cap with the default 20 error target in the demo but it does reach it with 16. With the updated load strategy ("TilesRenderer.optimizedLoadStrategy = true", which will be changed to the default at some point) the number of tiles is reduced by ~20% in the Google tiles case but I agree more overhead is always better.

src/three/renderer/utils/MemoryUtils.js
gkjohnson: That would be great. A PR to improve handling of just ExternalTextures is something we could merge immediately. A couple other things that come to mind that need to be handled:

- How is disposal handled for these external textures? Does calling ExternalTexture.dispose actually dispose of the WebGLTexture / WebGPUTexture handle? Or does disposal need to happen separately, as with an imageBitmap?
- The `UnloadTilesPlugin` is designed to delete any GPU memory for tiles that aren't actively visible or being rendered. When the tiles are made visible again the textures and geometry are re-initialized on the GPU. I assume this model wouldn't work for the ExternalTexture since re-initializing it isn't as simple as just reuploading the existing data? The easy thing to do in that case is to just ignore and not dispose ExternalTextures encountered in the plugin.

src/three/plugins/GLTFExtensionsPlugin.js
castano: I've been looking into this and I'm unable to reproduce the problem that the unregister calls were intended to solve. The relevant code in three.js has not changed in years, so I'm a bit puzzled. The good news is that plugin registration can now be simpler and won't require changes to to the `GLTFExtensionsPlugin` as proposed here.

I'm keeping the existing `registerSparkLoader` function, but I've added a new `createSparkPlugins` that you can use as follows:

```js
plugins: createSparkPlugins( spark, { preferLowQuality, generateMipmaps } )
```

package.json
```diff
@@ -120,7 +121,7 @@
     "@react-three/fiber": "^8.17.9 || ^9.0.0",
     "react": "^18.3.1 || ^19.0.0",
     "react-dom": "^18.3.1 || ^19.0.0",
-    "three": ">=0.167.0"
+    "three": ">=0.182.0"
```
gkjohnson, line 124: We shouldn't change the peer dependency requirements for the project

example/three/googleMapsAerial.js
```diff
@@ -9,14 +11,27 @@ import {
...
+let spark;
+let stats;
+
+const params = {
+	enableSpark: true,
+	preferLowQuality: true,
+	generateMipmaps: false,
+	errorTarget: 16.0,
+	textureVRAM: 0,
+};
```
gkjohnson, line 29: I think my preference would be to not expose all these spark-specific parameters. I think we can add spark to some of these demos that really benefit (googleMapsExample, googleMapsAerial) and add a toggle for spark to the googleMapsExample that already has a gui panel. Then the spark documentation can expose and explain all the parameters. And / or show a more complete 3d tiles integration demo.

Basically I want to keep these demos simple where I can, so lets make minimal changes to the demo if possible. I'm happy to also include something like a "FAQ" on the main README that includes some recommendations for improving memory and draw performance that includes a reference to Spark.

example/three/googleMapsAerial.js
```diff
@@ -154,4 +202,8 @@ function render() {
 
 	}
 
+	// Update GPU memory info
+	const approximateVRAM = estimateBytesUsed( scene );
+	params.textureVRAM = ( approximateVRAM / ( 1024 * 1024 ) ).toFixed( 1 );
```
gkjohnson, line 207: The "googleMapsExample" demo displays the cached bytes value by accessing the `tiles.lruCache.cachedBytes` field, which should be adequate for demonstrating the improvements.

src/three/plugins/GLTFExtensionsPlugin.js
gkjohnson: This looks good to me assuming the GLTFLoader isn't now processing each image twice, right? Once of the original plugin, once for the new one?

gkjohnson:
@castano somewhat offtopic for this PR but is it possible for spark.js to transcode a "Canvas" element or "SVG" element used for a texture to a more memory efficient GPU format? Some of the plugins are using Canvas to draw vector graphics (like for GeoJSON or other formats) or compose multiple tiled image textures so we don't have a traditional image handle to process in these cases but it affording memory improvements would be nice.

src/three/plugins/GLTFExtensionsPlugin.js
castano: No, I confirmed the new plugin is indeed replacing the old one.

package.json
castano: I'll need to double check whether the required three.js features are all present in that older version, and if they don't gracefully handle the error. It may just work, since most of the three.js changes were in the WebGPU backend, and to add better support for normal maps, which these demos don't need.

example/three/googleMapsAerial.js
castano: I currently have a spark-specific example that is probably better suited to showcase the individual options. Would you be OK adding a separate example/demo?

example/three/googleMapsAerial.js
castano: Thanks, I'll look into using that instead.

castano:
> somewhat offtopic for this PR but is it possible for spark.js to transcode a "Canvas" element or "SVG" element used for a texture to a more memory efficient GPU format? Some of the plugins are using Canvas to draw vector graphics (like for GeoJSON or other formats) or compose multiple tiled image textures so we don't have a traditional image handle to process in these cases but it affording memory improvements would be nice.

Canvas and SVG elements should work fine, but I have not tested that code path under WebGL. You can find a 
WebGPU SVG example in the spark.js SDK:

https://github.com/Ludicon/spark.js/blob/b01b03fd4cd19d70a0c3ea977e348c416e6700c8/examples/svg.html

Examples are automatically published at:

https://ludicon.github.io/spark.js/

I'll look into testing the canvas and SVG code paths under WebGL, but even if it doesn't work now, it shouldn't be hard to support that.

If you can point me to the examples that could benefit from that support I can take a look and ensure it works for those use cases.

package.json
gkjohnson: Spark.js is not a dependency or requirement of this project so the peer dependency should not change - users can use and install 3d-tiles-renderer with r167+ just fine. Spark.js will need to specify it's own three.js dependency limit so if users install Spark that peer dependency version will need to be respected.

example/three/googleMapsAerial.js
gkjohnson: The examples are already proving to be a burden to maintain so I'd prefer to not add more unless they're necessary for new features within the project. We can add a "use spark" option to some of the existing demos (and link to it) but otherwise I think the Spark.js demos and documentation should be responsible for fully demonstrating Spark.js features.

gkjohnson:
> Canvas and SVG elements should work fine, but I have not tested that code path under WebGL. You can find a
WebGPU SVG example in the spark.js SDK:
>
> https://github.com/Ludicon/spark.js/blob/b01b03fd4cd19d70a0c3ea977e348c416e6700c8/examples/svg.html

It looks like this is passing the SVG path as a string but I expect it will work just fine if you pass the SVG as an image element, as well:

```js
const svg = new Image();
svg.src = './assets/Tiger.svg';

const texture = await spark.encodeTexture( svg, { format: "rgba" } );
```

--

Regarding canvas, though - it looks like the "encodeTexture" function does not take an `HTMLCanvasElement` (or `OffscreenCanvas`) as an argument, which is why I ask:

> source (string | HTMLImageElement | ImageBitmap | GPUtexture)
> The image to encode. Can be a GPUTexture, URL, DOM image or ImageBitmap.

The "ImageOverlayPlugins", which is used in "quantMeshOverlays" demo, composes tiled textures for overlays. I'll have to think through how something like spark can be integrated, though.

src/three/renderer/utils/MemoryUtils.js
castano: OK, I just submitted https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/1560

My understanding is that external textures are disposed just like standard textures do. No difference there.

The external texture does not keep track of the source data, so it's not possible to re-initialize it easily. It should be possible to have the external texture track the source image and creation options to allow that, but that would require the plugin to be aware of spark.js external textures.

example/three/googleMapsAerial.js
castano: Sounds good. Here's my plan: once the changes to `getTextureByteLength` are integrated, I'll merge:

https://github.com/Ludicon/spark.js/pull/29

This is a minimal example showing how to use three.js 3d tiles renderer with spark.js

I also have a more complex demo ready to go, but I'll publish that in a standalone repository.

I'd be happy to add spark support to the 3DTilesRenderer examples that you think would benefit the most. It should not require more than a handful lines of code per example.

castano:
> Regarding canvas, though - it looks like the "encodeTexture" function does not take an HTMLCanvasElement (or OffscreenCanvas) as an argument

Right. HTMLCanvasElement and OffscreenCanvas did actually work already, but did not want to document that since I had never tested it. I've updated the docs and added an example to validate that code path:

https://github.com/Ludicon/spark.js/pull/30

src/three/renderer/utils/MemoryUtils.js
gkjohnson: > My understanding is that external textures are disposed just like standard textures do. No difference there.

I think this should be validated. Looking through the code it looks like the "dispose" callback which ultimately disposes of the internal texture handle is [registered in "initTexture"](https://github.com/mrdoob/three.js/blob/5a3de5c7324d4cf1834d369050f0f8e8d86b2068/src/renderers/webgl/WebGLTextures.js#L715-L721) (called from "updloadTexture"), which [never gets called if using "ExternalTexture"](https://github.com/mrdoob/three.js/blob/5a3de5c7324d4cf1834d369050f0f8e8d86b2068/src/renderers/webgl/WebGLTextures.js#L559-L578). It's likely that the application is responsible for disposing of the WebGLTexture handle, similar to "image bitmap". Note that closing image bitmaps is explicitly handled [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/d26e924a8e262c7f2bf79260af5b4b6b8b653f70/src/three/renderer/tiles/TilesRenderer.js#L922-L926) and [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/d26e924a8e262c7f2bf79260af5b4b6b8b653f70/src/three/renderer/tiles/TilesRenderer.js#L851-L855) in the project, so we may have to do something similar here.

> The external texture does not keep track of the source data, so it's not possible to re-initialize it easily. It should be possible to have the external texture track the source image and creation options to allow that, but that would require the plugin to be aware of spark.js external textures.

I think if we handle the textures like image bitmap this should be okay.

castano: You are right. I think this should take care of it: https://github.com/Ludicon/spark.js/pull/31

I'll be traveling for the next few days, but I'll test it more thoroughly and merge when I return.

gkjohnson: This may work for use within this 3d tiles project but I don't think it exactly lines up with expectations around texture disposal within three.js.

In three multiple textures can share the same "source" (the external webgl texture in this case), which three.js will use reference counting to determine whether to actually remove the content from the GPU. And in common cases users may dispose a texture to remove the content from the GPU only to reuse it later (this project includes a plugin that does exactly that) but given the nature of the texture handle isn't something that can be simply reuploaded.

This is a similar issue to what we have to ImageBitmap because once that data handle has been closed it can't be reuploaded. This is why it's the users responsibility to call "close" when the data is definitely finished with even when the image is loaded via something like GLTFLoader.

It seems like one of the issues in this case is that in order to release the WebGLTexture handle we need access to the associated WebGL context, which may not always be easily available (as is the case here). If the context were stored on the ExternalTexture itself this would help, though. I'm wondering if storing the associated context on the texture is something three.js would support? I'm trying to think through other clean solutions, but this is a bit complicated. WebGPU seems to make this easier since you can just call "destroy" on the handle itself.

castano: My (limited) understanding is that ExternalTextures are not managed by three.js and thus it's up to the application to control their lifetime. This can potentially result in duplicate allocations, for example, if you have multiple glTF models sharing a texture, the glTF loader with the spark plugin will create multiple instances of that same texture, while the standard three.js loading path may only create one.

I believe that for 3DTiles that's usually not a problem, because textures are unique and not shared, (unless you load the same model multiple times). For other applications this is worth taking into consideration. I'll start opening an issue to track this: https://github.com/Ludicon/spark.js/issues/32

I think the solution I proposed here works in this context because there's always a one to one mapping between ExternalTexture and associated WebGL or WebGPU texture. For WebGL, the gl context is captured by the dispose lambda, so it's associated to the ExternalTexture transparently.

gkjohnson: > My (limited) understanding is that ExternalTextures are not managed by three.js and thus it's up to the application to control their lifetime. This can potentially result in duplicate allocations, for example, if you have multiple glTF models sharing a texture, the glTF loader with the spark plugin will create multiple instances of that same texture, while the standard three.js loading path may only create one.

If the same texture is embedded in two different loaded glTF files then three.js will load the data twice. The same instance of the image data can be shared between multiple meshes loaded within the same glTF, though.

In three.js there is a concept of an "Image Source" (confusingly called [Source](https://threejs.org/docs/?q=source#Source)) and a [Texture](https://threejs.org/docs/?q=texture#Texture). The image source represents the actual color values for pixels while the "Texture" represents UV transforms and references a shared "image source" instance.

> I believe that for 3DTiles that's usually not a problem, because textures are unique and not shared, (unless you load the same model multiple times). For other applications this is worth taking into consideration.

In a basic sense, possibly. But users will be expecting these textures to behave as normal textures do elsewhere so I'm somewhat thinking forward to "other applications". This project contains an optional [UnloadTilesPlugin](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/plugins/API.md#unloadtilesplugin), for example, which disposes of all geometry and texture data from the GPU specifically to keep VRAM usage low at the cost of having to reupload texture data. This plugin calls "dispose" on textures because in all cases, including when using image bitmap, this will still enable the texture to be reuploaded and reused. With the changes suggested using spark along with the above plugin will likely result in errors since suddenly "dispose" is behaving differently in this specific case.

This case is no different than what's going on with "ImageBitmap" textures. The "ImageBitmap" textures are specifically not closed when calling "dispose" to avoid the above kind of issue. It's left to the users to determine when that texture data is _absolutely and completely done with_, which is not always synonymous with calling "dispose".

As a simple illustration, here's a very basic case in which I believe spark will break with that change:

```js
const gpuTexture = await spark.encodeTexture( textureUrl, { srgb: true, flipY: true } );
const texA = new THREE.ExternalTexture( gpuTexture );
texA.center.x = 0.5;

const texB = new THREE.ExternalTexture( gpuTexture );
texA.center.y = 0.5;

const texC = texA.clone();
texC.rotation = Math.PI / 2;

// calling "texA.dispose" will cause texB and texC to stop working since
// they share the "gpuTexture" handle
texA.dispose();
```

---

All to say this isn't as simple as it might seem on the surface and disposing of the handle this way really limits the usability of the texture, imo. My general preference would be for this to be disposed in a similar way as "imageBitmap":

```js
if ( texture.image instanceof ImageBitmap ) {
  texture.image.close();
}
```

In WebGPU it's simple and comparable to ImageBitmap:

```js
if ( texture.isExternalTexture ) {
  texture.image.destroy();
}
```

But in WebGL we need the handle to context to dispose of it. One option would be keeping the context on the "userData" object so it's available for use during disposal. Then if the context isn't available we can log a warning and let the user know that their external texture isn't set up properly rather than silently accumulating memory:

```js
if ( texture.isExternalTexture ) {
  if ( texture.image instanceof WebGLTexture ) {
    // webgl
    if ( ! texture.userData.gl ) {
      console.warn( 'TilesRenderer: Cannot dispose ExternalTexture that does not include WebGL context.' )
    } else {
      texture.userData.gl.deleteTexture( texture.image );
    }
  } else {
    // webgpu
    texture.image.destroy();
  }
}
```

The down sides are that this property is not automatically copied with something like a "clone" call. Really, though, it would be best if three.js made this easier and at least stored the relevant context on the external texture by default so this were more guaranteed.

---

Sorry for the long response - I think my primary concern is that as it is we can't notify a user that their texture isn't being cleaned up if they're using a non-Spark ExternalTexture instance, which feels like a pretty big design issue, though I'm not sure how prominent the class usage is.

donmccurdy: > ... I think my primary concern is that as it is we can't notify a user that their texture isn't being cleaned up if they're using a non-Spark ExternalTexture instance...

Hm, my original intention for THREE.ExternalTexture had been that if the user (or a library the user delegates to) creates a GPU resource, they must also manage the GPU resource throughout its lifecycle. From that perspective: 

1. It feels unexpected that the user must store the WebGL context on the texture
2. It should be up to the user (or delegate library) to decide whether the GPU resource has a 1:1 or 1:many relationship with ExternalTexture container(s), and to manage GPU resources accordingly

If the user is relying on Spark to create these ExternalTextures, it becomes Spark's responsibility to either manage the resource lifecycle, or communicate that responsibility to the user. If the user has created an ExternalTexture manually or with some other dependency, it would be the user and/or dependency's responsibility to do the same.

The proposal https://github.com/Ludicon/spark.js/pull/31 is consistent, I think, in defining a clear contract for the lifecycle of the ExternalTexture and its GPU resource. Arguably, GLTFLoader is not doing so good of a job there, by constructing ImageBitmap textures in [unspecified conditions](https://github.com/mrdoob/three.js/blob/1d9c50e33ed61e52845330468ca2a174cd3ff8af/examples/jsm/loaders/GLTFLoader.js#L2600-L2604), and leaving it to the user to dispose those resources manually... which I'd guess is often overlooked. I don't have an easy fix for that, but it's not a pattern I'd want to promote.

***

I think it's not practical for an ExternalTexture to store enough metadata for its GPU texture to be recreated after disposal, would you agree? The original image should no longer exist in memory. Assuming so... maybe a `spark.destroyTexture(gpuTexture)` utility would be helpful rather than passing around the GL context? The typical pattern is that three.js textures can be reused after disposal, so 3DTilesRendererJS could dispose the texture eagerly, and the GPU resource only after the tile is finally unloaded.

gkjohnson: @donmccurdy I guess fundamentally I'm confused as to why one would advocate for his behavior in the spark library but not with ImageBitmap in GLTFLoader - it's just as possible for GLTFLoader to register the same kind of disposal event but I think it's known this causes massive usability issues, which is why it is the way it is.

> If the user is relying on Spark to create these ExternalTextures, it becomes Spark's responsibility to either manage the resource lifecycle, or communicate that responsibility to the user. 

To be clear I completely agree that symmetry in the life cycle of the texture is both a nicer and safer pattern. The problem is the practical issues it causes. The suggested pattern betrays the expectations by every other Texture instance in the project. Off the top of my head my own real-world use cases this would have broken if it behaved this way:

- Disposing of a texture to temporarily replace it with another, then swapping the original texture back in later.
- Disposing of resources to store them in a CPU-side cache to avoid having to reload them again when needed.
- Cloning a texture to adjust offsets, then disposing the original.
- Disposing of a cloned instance _will not_ actually dispose of the external texture handle

All of these _very reasonable_ use cases would wind up breaking if textures were managed in the way that is being suggested.

> It should be up to the user (or delegate library) to decide whether the GPU resource has a 1:1 or 1:many relationship with ExternalTexture container(s), and to manage GPU resources accordingly

This isn't how three.js' API presents itself though. Perhaps it's just becoming clear that three.js' API surface was not designed for these kinds of use cases if even just because the original source image textures used did not have these issues. Why should "dispose" behave inconsistently for image bitmap or external texture? Perhaps it's just the way I'm thinking about but as I mentioned before, "dispose" has _never meant_ "dispose content forever" in three.js. Perhaps there's are argument to be made for adding a "disposeForever" function to three.js textures to accommodate these types of use cases.

> I think it's not practical for an ExternalTexture to store enough metadata for its GPU texture to be recreated after disposal, would you agree

Of course agree with this. But I also think that I should also be able to call "dispose" and expect consistent behavior. These classes are supposed to be reliable abstractions with consistent, usable behavior so I'm not in agreement that the "dispose" function should suddenly be overloaded to completely discard resources and make a texture instance unusable, which again doesn't happen anywhere else.

> Assuming so... maybe a spark.destroyTexture(gpuTexture) utility would be helpful rather than passing around the GL context?

My preference is to not add library-specific code throughout the project to accommodate every special case someone may come to the project with.

--

Either way if this is the preference for how to manage the disposal of these textures that's fine. I only intended to point out what will be ergonomic issues down the line (and within this project).

castano: The more I learn about how ExternalTextures work, the more confused I am. The behavior of the WebGL and WebGPU backends are entirely different.

In WebGPU, it appears that the behavior depends on whether the texture has been used for rendering or not. During rendering, three calls `Textures.updateTexture` and this registers the textureData in the Textures DataMap and enables the shared ownership behavior that you have come to expect. 

This is quite confusing. If you call dispose() prior to rendering, nothing happens, if you call it after, the texture is destroyed. If you have multiple textures referencing the same GPUTexture, but some have been used for rendering and others haven't, then the behavior is completely broken.

The WebGL backend doesn't appear to behave like that, and it never takes ownership of the textures objects. This is at least more consistent.

I'm working on a solution that involves the creation of a map to manage the lifetime of spark external textures. Instead of using THREE.ExternalTexture to create a texture, you use a derived class that overrides the copy constructor and the dispose method and tracks the gl object associated with spark and the number of textures that reference it, so that it can dispose the texture object when no texture references it anymore. I'll post an update once I have something solid.

castano: OK, here's what I have:

https://github.com/Ludicon/spark.js/pull/31/changes/f2915cdc2939e5fed0e0297a8c39cd7b479e868a

Usage is as follows:
```
        const gl = this.spark.gl
        const onRelease = isWebGL ? tex => gl.deleteTexture(tex) : tex => tex.destroy()
        const texture = new SparkThreeExternalTexture(gpuTexture, onRelease)
```

in the case of the 3DTilesRenderer no changes are necessary. The GLTF loader plugin takes care of creating the external texturs using the SparkThreeExternalTexture class.

It works as expected in my examples, but it needs more testing. Let me know what you think of this approach.

gkjohnson: > During rendering, three calls Textures.updateTexture and this registers the textureData in the Textures DataMap and enables the shared ownership behavior that you have come to expect.

This difference in behavior between WebGLRenderer and WebGPURenderer may not be intended or desired. WebGPURenderer is still very much in development and there are still quite a few oversights like this throughout.

> OK, here's what I have:
>
> https://github.com/Ludicon/spark.js/commit/f2915cdc2939e5fed0e0297a8c39cd7b479e868a

This doesn't completely address everything but I think this is fundamentally something that three.js needs to promote a more consistent pattern for if it's going to become more commonly used. I think we can wait for more problems to arise before trying to address this further.

PR #1503, opened 2026-03-10
Update wms image source wmts image source and add url template image source

TQtong:
Update wms image source wmts image source and add url template image source

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4274,11 +4274,14 @@
       }
     },
     "node_modules/dompurify": {
-      "version": "3.3.1",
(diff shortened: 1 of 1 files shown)


gkjohnson:
Can we keep these changes well-scoped? There are far more changes here than are needed for updating the WMTS image source. Unless I'm missing something there shouldn't be any changes needed to WMS image source and I don't think we need to add a URLTemplateImageSource or a new example.

Please keep the changes minimal. If you want to update the API for specifying WMTS Image Source to support just literals for matrix sets and layers then please only make the changes necessary to do so.

PR #1504, opened 2026-03-10
Improve doc

TQtong:
I referenced the three.js documentation to build it, and used the simplest content to test whether it can achieve the expected effect.


gkjohnson:
I've aligned the changes here what I was imagining in #1508, which generates markdown directly from the JSDoc json rather than using "turndown" to convert the HTML, so I'll go ahead and close this.

gkjohnson:
Thanks for the start here - I've merged #1508. The next steps will be moving documentation over to JSDoc and ensuring we can convert it to markdown well

Issue #1505, opened 2026-03-12
Title: `quantMeshOverlays` demo intermittently crashes when moving the camera


WilliamLiu-1997:
### Description

The official `quantMeshOverlays` example intermittently crashes when the camera is moved around for a while. This is reproducible on the hosted demo, without any local modifications.

Demo:
https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/three/quantMeshOverlays.html

## Reproduction

1. Open the `quantMeshOverlays` example.
2. Leave the default settings as-is (`enableTileSplitting` is enabled).
3. Orbit / pan / zoom the camera continuously and somewhat aggressively for a bit.
4. After some camera movement, the example occasionally throws runtime errors and rendering breaks.

Possible cause: `QuantizedMeshPlugin` and `ImageOverlayPlugin` may get out of sync when tiles are split, unloaded, or disposed while the camera is moving. It looks like one part of the system still expects a tile and its children / overlay state to exist, while another part has already cleared them. That would explain the `null` / `undefined` errors and the follow-up traversal failures.

<img width="1904" height="968" alt="Image" src="https://github.com/user-attachments/assets/7717f29f-44ec-4e29-be4f-f1ebf7d04d9e" />

### Reproduction steps

1. Open the `quantMeshOverlays` example.
2. Leave the default settings as-is (`enableTileSplitting` is enabled).
3. Orbit / pan / zoom the camera continuously and somewhat aggressively for a bit.
4. After some camera movement, the example occasionally throws runtime errors and rendering breaks.

### Code

The official `quantMeshOverlays` example intermittently crashes when the camera is moved around for a while. This is reproducible on the hosted demo, without any local modifications.

### Live example

https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/three/quantMeshOverlays.htm

### Screenshots & Tile Set

<img width="3747" height="1906" alt="Image" src="https://github.com/user-attachments/assets/28abecc0-37af-4db3-8f3a-6a8e974eb2e1" />

### Library Version

latest

### Three.js Version

latest

gkjohnson:
Thanks for the report - thinking about the changes in #1498, I think there are a couple cases that are unaccounted for:

[...]

PR #1506, opened 2026-03-12
fix: handle dynamically appended quantized mesh children safely

WilliamLiu-1997:
Preprocess missing appended child tiles before traversal instead of assuming the first child represents the whole children array.

Also initialize quantized mesh engine resource arrays when the plugin assigns engineData.scene so early disposal does not see partially initialized tile state.

This fixes the #1505

### DIFF
--- a/src/three/plugins/QuantizedMeshPlugin.js
+++ b/src/three/plugins/QuantizedMeshPlugin.js
@@ -254,6 +254,9 @@ export class QuantizedMeshPlugin {
 			clipper.minLon = west;
 			clipper.maxLon = east;
 
+			// There is a short async gap between the parent tile finishing parse and TilesRenderer
+			// assigning engineData.scene. Split children can be processed during that gap, so keep a
+			// temporary reference to the parsed parent scene for clipping.
 			const scene = tile.parent.engineData.scene || tile.parent[ TILE_SPLIT_SOURCE_SCENE ];
 			result = clipper.clipToQuadrant( scene, left, bottom );
 
@@ -312,6 +315,9 @@ export class QuantizedMeshPlugin {
 
 		}
 
+		// Store the parsed scene separately until TilesRenderer finishes registering the tile model.
+		// Because parseTile is async, split children may be processed before the parent scene has
+		// been assigned on engineData and still need access to the parsed parent scene for clipping.
 		tile[ TILE_SPLIT_SOURCE_SCENE ] = result;
 		this.expandChildren( tile );
 


src/core/renderer/tiles/optimizedTraverseFunctions.js
```diff
@@ -30,8 +30,19 @@ function isProcessed( tile ) {
...
+	for ( let i = 0, l = children.length; i < l; i ++ ) {
+
+		if ( ! isProcessed( children[ i ] ) ) {
+
+			childrenReady = false;
+			break;
+
+		}
+
+	}
```
gkjohnson: One of the reasons only a single child was checked is to avoid iterating over and checking every single child in cases where there may be thousands depending on the tile set. Rather than the first child I'm thinking this can be changed to check that _last_ child, which will work under the assumption that new children are always added to the end of the children array.

src/three/plugins/QuantizedMeshPlugin.js
```diff
@@ -315,6 +348,10 @@ export class QuantizedMeshPlugin {
 		// first, otherwise, while the parent is still downloading.
 		// Ideally we would be able to guarantee parents are loaded first but this is an odd case.
 		// Assign the scene value preemptively to ensure it's available for splitting.
+		tile.engineData.geometry = [];
+		tile.engineData.materials = [];
+		tile.engineData.textures = [];
+		collectSceneResources( result, tile.engineData );
```
gkjohnson: Looks like this is intended to fix point 1 in https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1505#issuecomment-4052098390. I don't love the redundancy since TilesRenderer is already doing this and now means we have to maintain the logic in two places in case new fields wind up being added later.

It would be best if we could remove the race condition necessitating this early scene assignment if possible.

WilliamLiu-1997: I removed the early `engineData.scene` assignment entirely and stopped duplicating the resource collection logic in `QuantizedMeshPlugin`. Split children now use a plugin-private temporary scene reference until `TilesRenderer` finishes registering the tile model, so the original split timing is preserved without exposing a partially initialized `engineData` state.
Does this match what you were suggesting?

gkjohnson: Keeping the scene on a separate private variable as you've done here works and is one solution I was considering, as well. I think we should still keep a comment or modify the existing one so it's clear why this is necessary and that there may be cases where the child nodes are being processed but the parent scene hasn't been assigned yet due to the async nature of the "parseTile" function.

gkjohnson:
Thanks for the fixes on this - looks great

gkjohnson:
I'll try to have this published this week

Issue #1492, opened 2026-03-04
How can load wmts service?


TQtong:
### Question

This is my reference address: https://cesium.com/learn/ion-sdk/ref-doc/WebMapTileServiceImageryProvider.html #943

But the request keeps returning errors.

https://jsfiddle.net/tangqi/fhzuvnL1/44/

### Supplemental Data

	const wmtsOverlay = new WMTSTilesOverlay( {
		url: 'https://basemap.nationalmap.gov/arcgis/rest/services/USGSShadedReliefOnly/MapServer/WMTS',
		layer: 'USGSShadedReliefOnly',
		style: 'default',
		format: 'image/jpeg',
		tileMatrixSet: 'default028mm',
		// tileMatrixLabels : ['default028mm:0', 'default028mm:1', 'default028mm:2' ...],
		maximumLevel: 19,
		color: '#ffffff',
		opacity: 1.0,
	} );

### Library Version

0.4.21

### Three.js Version

r170.0

TQtong:
I found an issue: In WMTSTilesOverlay, the `capabilities` are set as optional, but if I only pass the URL (without providing `capabilities`), it fails to load properly.

TQtong:
# WMTSImageSource Enhancement: Auto-load Capabilities Support

## Problem Description

In the original implementation, the `capabilities` parameter in `WMTSImageSource` and `WMTSTilesOverlay` was marked as optional in the type definitions (`capabilities?: WMTSCapabilitiesResult | null`), but the actual code logic **required** `capabilities` to function properly.

### Original Code Issue

```javascript
init() {
    const { tiling, dimensions, capabilities } = this;
    
    // Directly accesses capabilities.layers, crashes if capabilities is null
    if ( ! layer ) {
        layer = capabilities.layers[ 0 ];  // ❌ Crashes when capabilities is null
    }
    
    // Even with url provided, capabilities is still needed for:
    // - layer info (boundingBox, dimensions, styles)
    // - tileMatrixSet info (supportedCRS, tileMatrices)
}
```

This prevented users from using `WMTSTilesOverlay` by simply passing a WMTS service URL.

## Solution

Modified `WMTSImageSource` to support two usage patterns:

### Pattern 1: Pass capabilities object (original approach)

```javascript
const capabilities = await new WMTSCapabilitiesLoader().loadAsync(wmtsUrl);
const wmtsOverlay = new WMTSTilesOverlay({
    capabilities: capabilities,
[...]

PR #1507, opened 2026-03-13
Chage wmts params(delete capabilities parameter)

TQtong:
These are the changes I made to WMTSImageSource.js after removing the capabilities parameter. I would appreciate it if the author could review whether there are any unreasonable parts, so I can understand and make targeted modifications accordingly. Thank you. #1492

### DIFF
--- a/src/three/plugins/images/sources/WMTSImageSource.js
+++ b/src/three/plugins/images/sources/WMTSImageSource.js
@@ -297,7 +297,7 @@ export class WMTSImageSource extends TiledImageSource {
 
 				for ( const key in dimensions ) {
 
-					url = url.replace( new RegExp( `{\\s*${key}\\s*}`, 'gi' ), dimensions[ key ] );
+					url = url.replace( new RegExp( `{\\s*${ key }\\s*}`, 'gi' ), dimensions[ key ] );
 
 				}
 


src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -1,143 +1,341 @@
...
-		this.dimensions = dimensions;
 		this.url = url;
+		this.format = format;
+		this.dimensions = dimensions;
+		this.tileMatrixLabels = tileMatrixLabels;
+		this.tileMatrices = tileMatrices;
+		this.projection = projection;
+		this.levels = levels;
+		this.tileDimension = tileDimension;
+		this.subdomains = Array.isArray( subdomains ) ? subdomains.slice() : subdomains.split( '' );
```
gkjohnson: Are you using subdomains? Unless there's a reason to include support for them I'd prefer to exclude this. With HTTP2 becoming more common place I understand that subdomains are less necessary.

src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -1,143 +1,341 @@
...
-			.replace( /{\s*TileCol\s*}/gi, x )
-			.replace( /{\s*TileRow\s*}/gi, y );
+		const bracketMatch = url.match( /\{/g );
+		if ( ! bracketMatch || ( bracketMatch.length === 1 && /\{s\}/.test( url ) ) ) {
+
+			return true; // KVP
+
+		}
+
+		return false; // RESTful
```
gkjohnson: With subdomains being removed we can just change this to `return /\{/.test( url )`.

src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -1,143 +1,341 @@
...
-		const { tiling, dimensions, capabilities } = this;
-		let { layer, tileMatrixSet, style, url } = this;
+		const {
+			tiling, tileDimension, levels, dimensions, contentBoundingBox,
+			tileMatrices,
+		} = this;
+		let { url, style, tileMatrixSet } = this;
+
+		const tileMatrixSetId = typeof tileMatrixSet === 'string' ? tileMatrixSet : 'default';
+		const resolvedStyle = style || 'default';
```
gkjohnson: The style is already defaulted to `'default'` in the  constructor so there's no reason to || with a default value here, again.

src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -1,143 +1,341 @@
...
 
-		const { tiling, dimensions, capabilities } = this;
-		let { layer, tileMatrixSet, style, url } = this;
+		const {
+			tiling, tileDimension, levels, dimensions, contentBoundingBox,
+			tileMatrices,
+		} = this;
+		let { url, style, tileMatrixSet } = this;
+
+		const tileMatrixSetId = typeof tileMatrixSet === 'string' ? tileMatrixSet : 'default';
```
gkjohnson: We should just set this to `'default'` in the construct the way we do "style" so we don't have to default it again here.

src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -1,143 +1,341 @@
...
+		if ( contentBoundingBox !== null ) {
 
-			layer = capabilities.layers.find( l => l.identifier === layer );
+			// contentBoundingBox is in degrees [west, south, east, north]; convert to radians
+			tiling.setContentBounds(
+				contentBoundingBox[ 0 ] * MathUtils.DEG2RAD,
+				contentBoundingBox[ 1 ] * MathUtils.DEG2RAD,
+				contentBoundingBox[ 2 ] * MathUtils.DEG2RAD,
+				contentBoundingBox[ 3 ] * MathUtils.DEG2RAD,
+			);
```
gkjohnson, line 244: The WMTSCopabilitiesLoader return value and all other references to content bounds is in radians throughout the project so we should take this argument in radians for consistency.

src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -1,143 +1,341 @@
...
 
 	init() {
 
-		const { tiling, dimensions, capabilities } = this;
-		let { layer, tileMatrixSet, style, url } = this;
+		const {
+			tiling, tileDimension, levels, dimensions, contentBoundingBox,
+			tileMatrices,
+		} = this;
+		let { url, style, tileMatrixSet } = this;
```
gkjohnson: "style" and "tileMatrixSet" look like they're never reassigned so they can use "const" field. I'm not sure why the linter isn't complaining about this? I'll have to take a look.

src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -1,143 +1,341 @@
...
+		} else {
+
+			tiling.setContentBounds( ...tiling.projection.getBounds() );
 
 		}
 
-		// extract the tile matrix set
-		if ( ! tileMatrixSet ) {
+		// Tiered initialization
+		if ( tileMatrices !== null && tileMatrices.length > 0 ) {
```
gkjohnson: We can just use `if ( Array.isArray( tileMatrices ) )` here

src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -1,143 +1,341 @@
...
-		// extract the style
-		if ( ! style ) {
+				let tileBounds = null;
+				if ( Math.abs( normW - 1 ) > 1e-6 || Math.abs( normH - 1 ) > 1e-6 ) {
 
-			style = layer.styles.find( style => style.isDefault ).identifier;
+					tileBounds = projection.toCartographicRange( [
+						0, 1 - normH,
+						normW, 1,
+					] );
```
gkjohnson: I'm not sure I'm understanding what's happening here

src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -1,143 +1,341 @@
...
+
+		}
+
+		const params = new URLSearchParams( {
+			SERVICE: 'WMTS',
+			VERSION: '1.0.0',
+			REQUEST: 'GetTile',
+			LAYER: this.layer,
+			STYLE: this.style || 'default',
+			TILEMATRIXSET: this.tileMatrixSet || 'default',
```
gkjohnson: We can remove these || value checks since the style is already defaulted in the constructor and we can do the same for "tileMatrixSet".

src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -1,143 +1,341 @@
...
+ */
 export class WMTSImageSource extends TiledImageSource {
 
+	/**
+	 * @param {WMTSImageSourceOptions} options - Configuration options.
+	 */
 	constructor( options = {} ) {
 
 		const {
-			capabilities = null,
```
gkjohnson, line 68: For the sake of backwards compatibility we should try to keep this parameter available. We can detect if it's present and fill in the other fields accordingly for now and add a deprecation warning using `console.warn` as other deprecated features are throughout the project. I plan to remove any deprecated features in the next minor version release.

src/three/plugins/images/sources/WMTSImageSource.js
```diff
@@ -1,143 +1,341 @@
-import { TiledImageSource } from './TiledImageSource.js';
+import { TiledImageSource } from '../sources/TiledImageSource.js';
```
gkjohnson: Why was this change made?

gkjohnson (commented):
Thanks - this is more followable for me. I have left some review comments with feedback.

src/three/plugins/images/sources/WMTSImageSource.js
TQtong: I previously encountered services that require subdomains in other projects at my company. That was the only time.

src/three/plugins/images/sources/WMTSImageSource.js
TQtong: During testing, I found that the service at https://gibs.earthdata.nasa.gov/wmts/epsg4326/best/wmts.cgi does not follow the conventional 2x grid approach. Loading it in the standard way would result in blank areas, so I applied normalization.

src/three/plugins/images/sources/WMTSImageSource.js
TQtong: My mistake — I copied directly from the test file and assumed the IDE would auto-replace it, but it didn't. Thanks for pointing that out.

src/three/plugins/images/sources/WMTSImageSource.js
gkjohnson: Can you explain what the issue was? And what this "normalization" step is actually doing and why? The GIBS tilesets were rendering fine in the demo before without this, is that not right?

src/three/plugins/images/sources/WMTSImageSource.js
gkjohnson: I would prefer to remove subdomains unless it can be shown they're really needed. As far as I've read these using subdomains should no longer be needed with HTTP2, which has been around for 10+ years.

src/three/plugins/images/sources/WMTSImageSource.js
TQtong: image

This is the rendering result when the tileMatrices parameter is not set.

image

This is the rendering result when the tileMatrices parameter is set.

		tileMatrices: [
			{ identifier: '0', matrixWidth: 2, matrixHeight: 1 },
			{ identifier: '1', matrixWidth: 3, matrixHeight: 2 },
			{ identifier: '2', matrixWidth: 5, matrixHeight: 3 },
			{ identifier: '3', matrixWidth: 10, matrixHeight: 5 },
			{ identifier: '4', matrixWidth: 20, matrixHeight: 10 },
			{ identifier: '5', matrixWidth: 40, matrixHeight: 20 },
			{ identifier: '6', matrixWidth: 80, matrixHeight: 40 },
			{ identifier: '7', matrixWidth: 160, matrixHeight: 80 },
			{ identifier: '8', matrixWidth: 320, matrixHeight: 160 },
		],
		

EPSG:4326 covers [-180°, 180°] × [-90°, 90°] — a 2:1 aspect ratio. At level 0, a 2×1 grid fits exactly. But at level 1, the grid becomes 3×2. Since tiles are square (512×512), this grid covers 3×512 = 1536 pixels wide and 2×512 = 1024 pixels tall — an aspect ratio of 1.5:1, not 2:1. The grid overshoots vertically: it implies coverage of 240° in latitude instead of the actual 180°.

Without `tileBounds`, the renderer would request tiles in this overflow region, receiving blank or error responses.

`normW` and `normH` are the normalized width and height of the current level's grid relative to the highest-resolution reference level. When they equal 1, the grid matches the content area perfectly and no `tileBounds` is needed.

When they don't equal 1, this code expresses the grid's actual coverage in normalized [0,1] coordinates (origin at top-left per WMTS convention), then converts it to cartographic (radian) bounds via `toCartographicRange`. The result tells the rendering system the true geographic extent of this level's tile grid, allowing it to skip tiles that fall outside the content area.

src/three/plugins/images/sources/WMTSImageSource.js
TQtong: Thanks for pointing that out — I hadn't noticed it either.

src/three/plugins/images/sources/WMTSImageSource.js
gkjohnson: I'm not sure I believe this is reliable. Is there a reference for this solution? The NASA GIBS data sets provide, and WMTS allows from what I understand, specifying a different bounds of coverage per layer - these bounds need to be provided by the user because they're otherwise arbitrary bounds that cannot be reliably calculated as far as I can see. This code now overwrites those provided by the parsed capabilities xml and may break other data sets.

Here's the tile matrix data provided by the GIBS capabilities XML:

```js
[ {
  "tileWidth": 512,
  "tileHeight": 512,
  "matrixWidth": 2,
  "matrixHeight": 1,
  "scaleDenominator": 223632905.6114871,
  "bounds": [
    -3.141592653589793,
    -3.4557519189487724,
    6.911503837897545,
    1.5707963267948966
  ]
},
{
  "tileWidth": 512,
  "tileHeight": 512,
  "matrixWidth": 3,
  "matrixHeight": 2,
  "scaleDenominator": 111816452.8057436,
  "bounds": [
    -3.141592653589793,
    -3.4557519189487746,
    4.398229715025713,
    1.5707963267948966
  ]
},
// ...
]
```

The test data you're using is failing because it does not reflect what is included in the capabilities file.

gkjohnson:
It would also be good to update the "wmtsTiles" example to avoid the use of "capabilities" since it's no longer recommended. Let me know once the conflicts are resolved and things are ready and I will take another look 🙏

src/three/plugins/images/sources/WMTSImageSource.js
TQtong: I apologize — I didn't think this through carefully enough. If the user can provide additional parameters like bounds, that would be the most accurate approach. Currently, I'm assuming a strict 2x scaling between levels and making a rough estimation of the bounds. For the most accurate bounds calculation, TopLeftCorner and ScaleDenominator would also be needed, but those are embedded in the XML. So I'd like to simply have the user pass in the bounds directly, instead of doing automatic calculation, and make the bounds parameter required. Does the author think this is acceptable?

TQtong: pixelSpan    = ScaleDenominator × 0.00028 / metersPerUnit(CRS)

tileSpanX    = TileWidth  × pixelSpan
tileSpanY    = TileHeight × pixelSpan

bottomRight  = [
    TopLeftCorner.x + tileSpanX × MatrixWidth,
    TopLeftCorner.y - tileSpanY × MatrixHeight,
]

bounds       = [ TopLeftCorner.x, bottomRight.y, bottomRight.x, TopLeftCorner.y ]

TQtong:
Author, the changes are done for now. Please point out any errors and I'll continue to optimize. Thank you.

example/three/wmtsTiles.js
```diff
@@ -4,22 +4,43 @@ import {
...
+	planar: false,
+	source: 'NASA GIBS',
+};
 
 init();
 render();
 
 function init() {
 
-	// renderer
```
gkjohnson, line 22: Please do not remove unrelated comments when adjusting files.

example/three/wmtsTiles.js
```diff
@@ -160,23 +84,22 @@ function rebuildTiles() {
...
+
 	tiles = new TilesRenderer();
 	tiles.registerPlugin( new TilesFadePlugin() );
 	tiles.registerPlugin( new UpdateOnChangePlugin() );
 	tiles.registerPlugin( new WMTSTilesPlugin( {
 		shape: params.planar ? 'planar' : 'ellipsoid',
 		center: true,
-		capabilities,
-		...params,
+		...sourceConfig,
```
gkjohnson: We should still use the WMTSCapabilitiesLoader so it can be tested and demonstrated. But we just need to pass the results directly into the plugin directly into the plugin. This should be a small change:

```js
const layer = capabilities.layers.find( l => l.identifier === params.layer );
const tileMatrices = layer.tileMatrixSets.find( tms => tms.identifier === params.tileMatrixSet )
const url = layer.resourceUrls[ 0 ].template;
tiles.registerPlugin( new WMTSTilesPlugin( {
  url,
  tileMatrices: tileMatrices,
  tileMatrixSet: params.tileMatrixSet,
  layer: params.layer,
  /* ...rest */
} );
```

TQtong:
I've resubmitted the example and resolved the conflicts.

example/three/wmtsTiles.js
gkjohnson: Far too much has changed in the wmtsTiles example. The only portion of the code that needs to be updated is the instantiation site of the WMTSTilesPlugin. Please review the suggestion I've made above showing how we can extract the fields from capabilities and pass them directly into the plugin arguments and ask questions if needed. I would push commits to the code but I am blocked from making changes on your branch.

gkjohnson:
There are lint errors in the code - you can review the issues in the CI log here:

image

Or by reviewing the issues output from running `npm run lint`.

gkjohnson:
I've cleaned up the example and fixed up the documentation and typescript type in #1550. Thanks for the help on this.

PR #1520, opened 2026-03-17
DebugTilesPlugin: add missing update() type definition

sguimmara:


gkjohnson:
What's the purpose of this addition? The function is not intended to be publicly called

sguimmara:
It appeared in the release notes, making me think it was public API. Closing.

gkjohnson:
Oh! Thanks for the clarification - I forgot I'd updated that function to be to reflect settings changes even when TilesRenderer update isn't being run but it never got documented. I added docs and update the types in 55fc9d954fa363ea98fbfb22ea140c59ba9a3e9d since I was in there.

PR #1521, opened 2026-03-17
Fix BatchedTilesPlugin instance leak

ErikSom:
- Fix instances never being freed when discardOriginalContent is true — dispose-model event was skipped because tile.engineData.scene was already set to null
- Fix capacity check using stale instanceCount instead of counting active instances
- Add debug option and getStats() for monitoring active/free instances

### DIFF
--- a/src/three/plugins/batched/BatchedTilesPlugin.js
+++ b/src/three/plugins/batched/BatchedTilesPlugin.js
@@ -46,8 +46,6 @@ export class BatchedTilesPlugin {
 			maxInstanceCount: Infinity,
 			discardOriginalContent: true,
 			textureSize: null,
-			debug: false,
-
 			material: null,
 			renderer: null,
 			...options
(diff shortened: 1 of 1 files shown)


src/three/plugins/batched/BatchedTilesPlugin.js
```diff
@@ -46,6 +46,7 @@ export class BatchedTilesPlugin {
 			maxInstanceCount: Infinity,
 			discardOriginalContent: true,
 			textureSize: null,
+			debug: false,
```
gkjohnson: Lets remove ny debug paths - if we'd like to add something like a div with debug data into the page I think doing so with the DebugTilesPlugin scraping other plugins would be the better spot.

src/three/plugins/batched/BatchedTilesPlugin.js
```diff
@@ -213,6 +235,12 @@ export class BatchedTilesPlugin {
 
 	}
 
+	disposeTile( tile ) {
+
+		this.removeSceneFromBatchedMesh( null, tile );
```
gkjohnson: It looks like "removeSceneFromBatchedMesh" no long really needs two arguments anymore. While we're making the changes here can we adjust the function signature, as well?

src/three/plugins/batched/BatchedTilesPlugin.js
```diff
@@ -347,7 +375,9 @@ export class BatchedTilesPlugin {
 
 		} );
 
-		const canAddMeshes = ! this.batchedMesh || this.batchedMesh.instanceCount + meshes.length <= this.maxInstanceCount;
+		let activeInstanceCount = 0;
+		this._tileToInstanceId.forEach( ids => activeInstanceCount += ids.length );
+		const canAddMeshes = ! this.batchedMesh || activeInstanceCount + meshes.length <= this.maxInstanceCount;
```
gkjohnson: Can you explain this? `instanceCount` should be the number of active instances in the BatchedMesh. I'm not seeing that these are ever different?

gkjohnson:
Thanks for looking into this!

src/three/plugins/batched/BatchedTilesPlugin.js
ErikSom: You are absolutely right. I was just trying multiple stuff and this was indeed not necessary. I have no experience with the BatchedMesh but the instanceCount is for -active- only:
https://github.com/mrdoob/three.js/blob/1939c35f2d92a4c870568da011aab54dabdfdd30/src/objects/BatchedMesh.js#L305

src/three/plugins/batched/BatchedTilesPlugin.js
ErikSom: Updated!

src/three/plugins/batched/BatchedTilesPlugin.js
ErikSom: Done, I thought so, but it would be an easy way for you to verify things for testing.

gkjohnson:
Looks good, thanks!

PR #1527, opened 2026-03-19
BatchedTilesPlugin copyTextureToTexture Adreno bug

ErikSom:
I created this PR to just address the issue and to provide a fix but I understand if you don't want hardware related fixes in your code. So what is this about..::

image

I noticed this issue on my Galaxy Z Fold 7. When using BatchedTIlesPlugin the moment the array expands all tiles start using the lowest level texture (earth view). After some debugging I figured it is in the threejs code and more specifically in the `renderer.copyTextureToTexture`, see my issue here:
https://github.com/mrdoob/three.js/issues/33211

In a nutshell there is a driver bug for many samsungs (potentially others) that causes copyTextureToTexture to duplicate layer 0 across all array texture layers.

ThreeJS closed the issue stating they don't think this fix should be in their code and that I should push further with ANGLE & Qualcomm so that is what Im doing see the issue here:
https://issues.angleproject.org/issues/42261453

But in the meantime this is an issue affecting real users and since there is no fix yet this is the best I could come up with. Feel free to close the PR but then this can stay as documentation or potential patch for people to apply themselves.

### DIFF
--- a/src/three/plugins/batched/BatchedTilesPlugin.js
+++ b/src/three/plugins/batched/BatchedTilesPlugin.js
@@ -2,7 +2,7 @@
 import { WebGLArrayRenderTarget, MeshBasicMaterial, DataTexture, REVISION } from 'three';
 import { FullScreenQuad } from 'three/examples/jsm/postprocessing/Pass.js';
 import { ExpandingBatchedMesh } from './ExpandingBatchedMesh.js';
-import { convertMapToArrayTexture, isColorWhite } from './utilities.js';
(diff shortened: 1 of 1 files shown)


gkjohnson:
Oh my favorite kind of bug 😅 I think I'm aligned with three.js in not wanting to integrate device-specific fixes in the project and hopefully the manufacturer can get this fixed sooner rather than later. But if you're running into issues in your local testing in the meantime you can override the "copyTextureToTexture" function like so:

```js
renderer.copyTextureToTexture = function( src, dst, srcBox, dstPos, ...args ) {
  // ...
};
```

PR #1529, opened 2026-03-20
EnvironmentControls: Support rendering pivotMesh in custom scene

shotamatsuda:
This PR adds a separate scene property for the pivot mesh in `EnvironmentControls`.

In fancy render pipelines with post-processing effects, there are many cases where the pivot mesh should be excluded from the main scene pass and blended after the effects are applied. For example, tone mapping exposure changes the brightness of the pivot circle, anti-aliasing blurs it, and depth-dependent effects reveal its rectangular bounds if it is rendered in the buffer where effects are applied. Having a property to keep the pivot mesh in a separate scene provides a way to address it on the user side (there _is_ [a way to do it](https://github.com/takram-design-engineering/three-geospatial/blob/6431ef4b245bfb21861bb42a58ca495fccef6924/storybook-webgpu/src/components/GlobeControls.tsx#L28) but it ends up being quite hacky).

An alternative approach would be to change its visibility instead of adding/removing it, then let users decide where it should live, similar to `TilesRenderer.group`.

(Or perhaps `Layers` could be used instead of this PR, though I have not tried yet.)

### DIFF
--- a/src/three/renderer/controls/EnvironmentControls.js
+++ b/src/three/renderer/controls/EnvironmentControls.js
@@ -390,23 +390,23 @@ export class EnvironmentControls extends EventDispatcher {
 					pointerTracker.isLeftClicked() && e.shiftKey
 				) {
 
-					this.setState( pointerTracker.isPointerTouch() ? WAITING : ROTATE );
-
 					pivotPoint.copy( hit.point );
 					pivotMesh.position.copy( hit.point );
 					pivotMesh.visible = pointerTracker.isPointerTouch() ? false : enabled;
 					pivotMesh.updateMatrixWorld();
 					scene.add( pivotMesh );
 
+					this.setState( pointerTracker.isPointerTouch() ? WAITING : ROTATE );
+
 				} else if ( pointerTracker.isLeftClicked() ) {
 
-					// if the clicked point is coming from below the plane then don't perform the drag
-					this.setState( DRAG );
 					pivotPoint.copy( hit.point );
-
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thanks! Would it be possible to adjust things so that you can perform this "fix" using the "start" event? This would require calling "setState" after the pivot mesh has been added to the scene in the "pointerdown" callback. It would mean you can handle it like so:

```js
controls.addEventListener( 'start', () => {
  const { pivotMesh } = controls;
  if ( pivotMesh.parent ) {
    customScene.add( pivotMesh );
  }
} );
```

Small aside but I wonder if this can be addressed in three.js at some point now that postprocessing is more embedded in the renderers. Interleaving object draws for things like UI between post processing effects is a common point of difficulty in post processing. Something like a draw priority or order for specifying where in the render pipeline something needs to be drawn may be helpful in the future:

```js
material.drawOrder = UI + 1;
```

shotamatsuda:
Thanks! The change is much cleaner. I confirmed that the pivot mesh can now be rerouted to a custom scene by the handler you suggested.

> Small aside but I wonder if this can be addressed in three.js at some point now that postprocessing is more embedded in the renderers. Interleaving object draws for things like UI between post processing effects is a common point of difficulty in post processing. Something like a draw priority or order for specifying where in the render pipeline something needs to be drawn may be helpful in the future:

Agreed. At least with WebGPURenderer, PostProcessing is now tightly integrated in the engine, so that objects and materials can be made aware of the order in the render pipeline.

Issue #1531, opened 2026-03-21
Ellipsoid: Error in getPositionToCartographic


mqrty:
### Description

Hi,

I found that the Cartesian to Cartographic conversion `Ellipsoid.getPositionToCartographic` is inaccurate in the resulting latitude. The error gets larger with increasing distance from the Ellipsoid.

This bug is due to the surface normal being calculated at the input position instead of its surface projection. When comparing to Cesium's [cartesianToCartographic](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Core/Ellipsoid.js#L512-L535) you can see they correctly do the latter.

https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/6090bdd295216f10acca1e3159ef20f8d0ad09a3/src/three/renderer/math/Ellipsoid.js#L351-L365
Suggested fix:
```diff
-	this.getPositionToNormal( pos, _norm );
+	this.getPositionToNormal( _vec, _norm );
```

I can submit a PR if you'd like.

### Reproduction steps

Plug a Cartesian with a height > 0 into `getPositionToCartographic`

### Code

See live example but i will paste its code here just in case:

```js
import * as Cesium from 'cesium';
import * as THREE from 'three';
import { WGS84_ELLIPSOID } from '3d-tiles-renderer';

const lon = -75;
const lat = 40;
const height = 5000;

const cartesian = Cesium.Cartesian3.fromDegrees(lon, lat, height);
const position = new THREE.Vector3(cartesian.x, cartesian.y, cartesian.z)

const cesiumCarto = Cesium.Ellipsoid.WGS84.cartesianToCartographic(cartesian, new Cesium.Cartographic());
const tilesRendererCarto = WGS84_ELLIPSOID.getPositionToCartographic(position, {});

console.log('cesium', {
  lon: Cesium.Math.toDegrees(cesiumCarto.longitude),
  lat: Cesium.Math.toDegrees(cesiumCarto.latitude),
  height: cesiumCarto.height
});

console.log('3dtilesrenderer', {
  lon: Cesium.Math.toDegrees(tilesRendererCarto.lon),
  lat: Cesium.Math.toDegrees(tilesRendererCarto.lat),
  height: tilesRendererCarto.height
});

console.log('delta lon:', Cesium.Math.toDegrees(tilesRendererCarto.lon - cesiumCarto.longitude));
[...]

PR #1532, opened 2026-03-23
Ellipsoid: fix normal not calculated at surface point in getPositionToCartographic

mqrty:
As explained in #1531, fixes the latitudinal inaccuracy of `getPositionToCartographic` by calculating the surface normal at the surface-projected position instead of the input position.

I also added a regression test that compares the results of `getPositionToCartographic` with the Cesium counterpart at a known coordinate.


gkjohnson:
Great, thank you!

Issue #1426, opened 2025-12-30
TilesRenderer: Make optimized traversal the default


gkjohnson:
Related to #1421 

**To Confirm / Fix**
- ~Test with a more complex additive tile set~
- ~Fix tile splitting for QuantizedMeshPlugin, ImageOverlaysPlugin~
  - See #857 
  - ~"Virtual" or "hasVirtualChildren" can be added to a separate fields which can be used to determine whether tiles should be marked as used until the necessary number of children have been disposed, at least.~ #1498 
- Expose variable indicating whether the tile set has full coverage or not
  - See #1452 
  - See #1445
  - _allChildrenReady may almost be acceptable but it seems to switch to false even when all tiles are rendering.
- Add option so the "optimized" path behaves like the original traversal approach. "loadParents" option, for example.

**Future**
- Fix auto-expanding children (QuantizedMeshPlugin, ImageFormatPlugins)
  - See #1344 
  - These do not work since they only expand the children once the tile itself is loaded and disposed of once the internal parents are disposed. That means we can get into an internal oscillation state when the parents are disposed, subsequently disposing of the children, causing that parent to render again.
  - Could implicitly add a tile with "unrenderable content" at some given number of layers so they re added / disposed in the same logic as internal tile sets. This won't work for tile splitting, though.
  - Could also adjust tiles to expand on "preprocess" or custom "expand virtual" callback and then wait for the tile to load before traversing to or loading children.
- Consider more flags
  - Add a flag indicating whether "full coverage" has been achieved during loading such that there will be no gaps even when moving the camera. A "loadProgress"-esque fields indicating a percentage of coverage filled may be better for initialization.
  - The priority download / parse queues could be adjusted to prioritize processing tiles that will fill gaps first over loading tiles that already have deeper children (or parents) loaded.

**Migration**
- Change "optimizedLoadStrategy" to "true" by default with a deprecation warning first
- Remove previous load strategy option

PR #1533, opened 2026-03-25
TilesRendererBase: Add "loadAncestors" flag for "optimizedLoadStrategy" traversal

gkjohnson:
Related to #1426 

Adds `TilesRenderer.loadParentTiles` flag to switch to loading parent tiles when `optimizedLoadStrategy` is true. This will allow for both types of load behavior with a single set of traversal functions.

**Next**
- Mark "optimizedLoadStrategy" as true by default, log deprecation warning when setting it to "false".
- Change "loadAncestors" flag to default to "true" (note to users that are setting optimized to true in log, as well)
- Handle the differences in the sort function.


gkjohnson:
Note that one current issue with "optimizedLoadStrategy" is that "on the fly"-expanded plugins will always wait for the parent tiles to be fully loaded (I think? Worth checking). It's possible that some "traverse" function would be best for this?

Issue #1280, opened 2025-08-18
Only root tile of XYZTilesOverlay is loaded


xawill:
### Description

I am opening this bug report following discussion about point 3 in #1264.

Trying to overlay swissBUILDINGS3D with aerial imagery using ImageOverlayPlugin (XYZTilesOverlay), only the root tile is loaded.

NB: Overlaying with exact same setup a quantized mesh works as expected (see [fiddle](https://jsfiddle.net/m19spob0/)).

I already investigated the bug and identified what I think is the issue:
- For swissBUILDINGS3D tileset (see link below), `tile.content?.uri` ([here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/bf91c00e9b1ae37004dcee4d36d195734e4db857/src/core/renderer/tiles/TilesRendererBase.js#L533)) is false. Even though there are children tiles. They are just not in the uri, but in a `refine: ADD` tileset.
- Therefore `tile.__hasRenderableContent` is set to false.
- Therefore `tile.__depthFromRenderedParent` is incorrectly set [here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/bf91c00e9b1ae37004dcee4d36d195734e4db857/src/core/renderer/tiles/TilesRendererBase.js#L580).
- Therefore [_calculateLevelFromOverlay](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/bf91c00e9b1ae37004dcee4d36d195734e4db857/src/three/plugins/images/ImageOverlayPlugin.js#L953) returns a wrong zoom level.
 
I don't know if this is the correct way to fix it, but currently I am using `return tile.__depth - 1;` instead of `return tile.__depthFromRenderedParent - 1;` ([here](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/bf91c00e9b1ae37004dcee4d36d195734e4db857/src/three/plugins/images/ImageOverlayPlugin.js#L953)) as a quick personal patch.

### Reproduction steps

1. Open network tab of Chrome dev tools (do this first in order to record network activity since the beginning)
2. Open fiddle below (in Chrome)
3. Filter by "jpeg"
6. Notice in the console that only the root tile (0/0/0.jpeg) is loaded.

### Code

See fiddle below

### Live example

~~[fiddle](https://jsfiddle.net/bdvc3wsp/)~~ [Updated fiddle](https://jsfiddle.net/m60p3bc5/)

[...]

Issue #1278, opened 2025-08-18
Allow higher zoom level overlay than "master" tileset tile's using ImageOverlayPlugin


xawill:
### Description

When overlaying aerial imagery on quantized mesh, for instance, I find the resolution of the texture to be lower than personal preference. I can reach the max zoom level of the overlay, but only when I am extremely zoomed in (almost not being able to see anything, being so close). `errorTarget=1`, if it helps; also tried with `errorTarget=0` with same results.

My understanding is that the plugin uses the same zoom level for the overlay as the "master" tileset's current tile, and only uses a higher zoom level for the overlay in case the "master" tileset's max zoom level is reached and there is still a higher zoom level available for the overlay (and `enableTileSplitting` is set to true).

I personally find it to be too low, at least with Swiss dataset:
[Quantized mesh](https://3d.geo.admin.ch/ch.swisstopo.terrain.3d/v1/layer.json)
[Aerial imagery](https://wmts.geo.admin.ch/1.0.0/ch.swisstopo.swissimage/default/current/3857/{z}/{x}/{y}.jpeg)

See fiddle [here](https://jsfiddle.net/m19spob0/).

### Solution

I would like to have a way to force using a higher zoom level for the overlay than the underlying tileset's tile.

I know 3D tiles work with setting `errorTarget`, but here I am not sure if that would make sense to have an errorTarget for the overlay as well. 

Maybe having a zoom level offset? If `zoomLevelOffset=2`, for instance, the overlay zoom level will be picked at "master" tileset tile's zoom level + 2 (if the 3D tile or quantized mesh is zoom level 18, it will overlay the tile with zoom level 20 imagery).

### Alternatives

None. Living with it :)

### Additional context

I have much better overlay resolution using `TextureOverlayPlugin`. Live demo available [here](https://earth-3d-viewer.xavier.willemin.swiss/). Look for "Vevey" in the search bar top right to see the same location as the fiddle. Screen capture:

<img width="2688" height="1828" alt="Image" src="https://github.com/user-attachments/assets/7075cc46-e5ff-41ec-8a14-1e1f263ce4a5" />

gkjohnson:
[...]

PR #1543, opened 2026-04-01
Unify overlay level calculation for planar and ellipsoid projections

xawill:
This is a proposal to compute the ImageOverlay level for ellipsoid projection using pixel-density-based level calculation. This is the same approach already used for planar projection.

This allows tilesets having ancestor tiles without renderable content to use the ImageOverlayPlugin properly. In particular, this would solve #1280.

I ran the test suite and all tests passed.

Thank you for your consideration.

### DIFF
--- a/src/three/plugins/images/ImageOverlayPlugin.js
+++ b/src/three/plugins/images/ImageOverlayPlugin.js
@@ -1480,7 +1480,6 @@ class TiledImageOverlay extends ImageOverlay {
 
 		}
 
-		// TODO: should this be one layer higher LoD?
 		return level;
 
 	}


src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1449,64 +1449,56 @@ class TiledImageOverlay extends ImageOverlay {
...
-			// TODO: should this be one layer higher LoD?
-			return level;
-
-		} else {
-
-			return tile.internal.depthFromRenderedParent - 1;
-
 		}
 
+		// TODO: should this be one layer higher LoD?
```
gkjohnson: I think this comment can be removed?

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -1449,64 +1449,56 @@ class TiledImageOverlay extends ImageOverlay {
...
-			const maxLevel = this.tiling.maxLevel;
-			for ( ; level < maxLevel; level ++ ) {
+			// the number of pixels per image on each axis
+			const wProj = resolution / w;
+			const hProj = resolution / h;
 
-				// the number of pixels per image on each axis
-				const wProj = resolution / w;
-				const hProj = resolution / h;
+			const { pixelWidth, pixelHeight } = this.tiling.getLevel( level );
```
gkjohnson: It's not guaranteed that every level is defined in a tiling scheme so we need to check if the layer is "null"

xawill: The diffs are a bit messed up, but notice that the code is exactly the same as before. Do you mean that for ellipsoid projection it might happen and now that the code branch are merged this is a case to account for?

src/three/plugins/images/ImageOverlayPlugin.js
xawill: I don't know, you wrote it so I kept it. The diffs are a bit messed up, but notice it was there already on line 1478. I guess with the callback proposed in #1278 the issue is less important anyway.

xawill:
Okey, I'll take a look at the demos.

xawill:
@gkjohnson It looks like the demos were already broken before this PR (tried master, v0.4.23 and v0.4.22: all failing). This is my first time building locally the examples, so I might miss something, but it looks like Cesium authentication plugin might suddenly be incompatible with current Cesium API?

Note that I can confirm I setup my access_token properly. Accessing the Cesium API with Postman using my token as a Bearer token works. So it looks like the `access_token` query string is not supported anymore?

src/three/plugins/images/ImageOverlayPlugin.js
gkjohnson: It means that the planar cases weren't tested with tiles that have the types of level gaps I'm referring to. But this is why changes need to be tested for completeness across the examples that use the code.

gkjohnson:
> It looks like the demos were already broken before this PR 

The demos are not broken. You can review the [CONTRIBUTING doc](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/CONTRIBUTING.md#running-demos-with-cesium-ion-tokens-locally) on how to run the cesium ion demos locally.

src/three/plugins/images/ImageOverlayPlugin.js
xawill: Understood. I just pushed the fix to handle this case.

I hesitated to init level at `this.tiling.minLevel`, but decided not to since this is not a precomputed attribute and anyway with the null/undefined check the case is already covered.

xawill:
You are right. The issue was that the required assets were not in my Cesium ION assets. By the way, it was painful to find the proper asset by ID in the Cesium ION asset depot (search doesn't work by assetID, but assetID can be set in the url, if this can help somebody in the future).

src/three/plugins/images/ImageOverlayPlugin.js
xawill: I am assuming you are considering this comment as outdated and therefore removed it. Let me know if you want it to be handled differently.

xawill:
I tested it with the two demos using `ImageOverlayPlugin` and my available datasets: it works fine. I also thought about other potential datasets and think all the cases are covered now. Let me know if your experience foresees any other potential issues.

.gitignore
```diff
@@ -66,6 +66,7 @@ typings/
 
 example/local-data
 example/dev-bundle
+example/bundle
```
gkjohnson, line 69: In the future please avoid making unrelated changes in PRs.

PR #1556, opened 2026-04-11
clear children after dispose in imageFormatPlugin

WilliamLiu-1997:
relate to https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1554

### DIFF
--- a/src/three/plugins/images/ImageFormatPlugin.js
+++ b/src/three/plugins/images/ImageFormatPlugin.js
@@ -173,6 +173,8 @@ export class ImageFormatPlugin {
 
 		}
 
+		tile.children.length = 0;
+
 	}
 
 	// Local functions
--- a/src/three/plugins/images/ImageFormatPlugin.js
+++ b/src/three/plugins/images/ImageFormatPlugin.js
@@ -142,21 +142,15 @@ export class ImageFormatPlugin {
 
 		}
 
-		return mesh;
-
-	}
-
-	preprocessNode( tile ) {
-
-		// generate children
-		const { tiling } = this;
-		const maxLevel = tiling.maxLevel;
-		const level = tile[ TILE_LEVEL ];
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thanks! Can you make sure you've tested with "TilesRenderer.optimizedLoadStrategy = true", as well, and test for regressions when scrolling around in the "mapTiles" and "wmtsTiles" demos, as well? They both use plugins that are based on ImageFormatsPlugin.

WilliamLiu-1997:
> optimizedLoadStrategy

optimizedLoadStrategy and mapTiles/wmtsTiles work, but I found an **issue** on this change:

If we set `children.length = 0` in `disposeTile()`, then when we access this node later, the `ensureChildrenArePreprocessed` function in `TilesRendererBase.js` will see `children.length === 0` and skip preprocessing the children. As a result, the level of detail gets stuck at a mid-level and cannot progress further.

I will investigate this and will let you know when I fix this issue

gkjohnson:
I think it's worth reviewing what QuatizedMeshPlugin is doing more completely. In that plugin children are expanded in the "parseToMesh" function which means that children will only be available once the parent tile has been loaded and then, symmetrically, those children are deleted when the tile is unloaded in "disposeTile". We should follow this same pattern in "ImageFormat" plugin.

WilliamLiu-1997:
> I think it's worth reviewing what QuatizedMeshPlugin is doing more completely. In that plugin children are expanded in the "parseToMesh" function which means that children will only be available once the parent tile has been loaded and then, symmetrically, those children are deleted when the tile is unloaded in "disposeTile". We should follow this same pattern in "ImageFormat" plugin.

Thanks, I have updated it according to what QuatizedMeshPlugin is doing. And I have tested it works fine with optimizedLoadStrategy and mapTiles/wmtsTiles

gkjohnson:
Excellent - thank you!

PR #1560, opened 2026-04-14
Update getTextureByteLength to handle compressed textures and external textures.

castano:
`getTextureByteLength` assumed the `tex` argument is a regular uncompressed texture backed by an image. When using compressed textures or external textures the code would throw an exception. 

This PR extends it to handle `CompressedTexture` (summing the attached mip-chain buffers) and `ExternalTexture` (reading `userData.byteLength`), and returns `UNKNOWN_TEXTURE_BYTE_LENGTH` (0) when the size can't be determined (missing image, missing dimensions, or unknown format) instead of throwing.

Originally discussed in #1497; splitting out as a standalone fix with no Spark dependencies.


src/three/renderer/utils/MemoryUtils.js
```diff
@@ -9,10 +27,47 @@ export function getTextureByteLength( tex ) {
...
+
+			if ( mip?.data?.byteLength ) {
+
+				bytes += mip.data.byteLength;
+
+			} else if ( mip ) {
+
+				bytes += getFormatByteLength( mip.width, mip.height, format, type );
+
+			}
```
gkjohnson: Small style nit: it shouldn't be the case that an mipmap entry can be "null", can it? I think we can just use an "else" block here.

src/three/renderer/utils/MemoryUtils.js
```diff
@@ -9,10 +27,47 @@ export function getTextureByteLength( tex ) {
...
+
+		return bytes;
+
+	}
+
+	if ( ! image || ! image.width || ! image.height ) {
+
+		return UNKNOWN_TEXTURE_BYTE_LENGTH;
+
+	}
```
gkjohnson: Are there real cases where we can run into this? What is "! image.width" really checking for?

gkjohnson (commented):
Thanks! I've made a couple comments - and we'll likely need to handle texture disposal correctly, too (see https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/1497#discussion_r3084316375)

castano:
Just trying to be extra careful to handle unexpected texture objects, but I can remove that if you think there's no need to. The `&& Array.isArray( mipmaps ) && mipmaps.length > 0` could also go away too.

src/three/renderer/utils/MemoryUtils.js
castano: done.

src/three/renderer/utils/MemoryUtils.js
castano: removed.

gkjohnson:
Thanks!

> see https://github.com/NASA-AMMOS/3DTilesRendererJS/pull/1497#discussion_r3084316375

Disposal handling is that last step, I think. Let me know if there's anything I can help clarify on that.

> Just trying to be extra careful to handle unexpected texture objects

I definitely understand the inclination but have found that in order to keep the code manageable, especially in javascript, it's best to assume that data is coming in well-shaped. Otherwise the amount of error handling and data checking will be a bit too much

castano:
Thanks. I think registering dispose event handlers from Spark's glTF loader might be the right approach, but I'm not sure whether that should live in the loader or be left to the application. What do you think? Either way, I think that's independent of this PR.

Here's a possible implementation: https://github.com/Ludicon/spark.js/pull/31

castano:
Do you have any more feedback regarding the changes in this PR? I think they are largely orthogonal to the handling of dispose events.

gkjohnson:
Thanks for the changes - we can use this for now and then revisit if there are any other issues related to disposal.

PR #1566, opened 2026-04-19
Allow user to override tile level calculation

xawill:
Following discussion in #1278.

In this proposal, `calculateLevel` can now return `null`. This user-override indicates that the current tile has no content and should not be rendered. Let me know what you think about it. An alternative would be to support `minLevel` and `maxLevel` on `TiledImageOverlay`. In that case, we might want to clamp `level` between `minLevel` and `maxLevel` in `calculateLevel`, even though after testing it I noticed that the rest of the codebase is already clamping `level` everywhere I checked.

### DIFF
--- a/src/three/plugins/images/ImageOverlayPlugin.d.ts
+++ b/src/three/plugins/images/ImageOverlayPlugin.d.ts
@@ -1,5 +1,6 @@
 import { Color, Matrix4, WebGLRenderer } from 'three';
 import { WMTSTileMatrix } from '../loaders/WMTSCapabilitiesLoader.js';
+import { Tile } from '../../../core/renderer/tiles/Tile.js';
 
 export class ImageOverlayPlugin {
 
@@ -37,6 +38,7 @@ export class TiledImageOverlay extends ImageOverlay {
 	readonly projection: any;
 	readonly aspectRatio: number;
 	fetchOptions: any;
+	mapZoomLevelToTile: ( ( level: number, tile: Tile ) => number | null ) | null;
 
 }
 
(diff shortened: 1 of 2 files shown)


xawill:
Thinking about it a bit more, there might be an issue with returning false from `hasContent` like this. It will prevent the tile from being refined in case the camera moves closer to the tile and `hasContent` should now return true, but it is not re-evaluated, is it?

Anyway it looks like a bit hacky to be able to return null from `calculateLevel`. Do you have any idea on how to solve the use case when the user wants to display the overlay only for zoom level >16 for instance?

gkjohnson:
Thanks for looking into this -

> In this proposal, calculateLevel can now return null.

I think it's okay for us to limit the return value of "mapZoomLevelToTile" (and therefore "calculateLevel") to numbers. Then a user can just return -1 to indicate "no content".

 > Thinking about it a bit more, there might be an issue with returning false from hasContent like this. It will prevent the tile from being refined in case the camera moves closer to the tile and hasContent should now return true, but it is not re-evaluated, is it?

I don't think I'm quite following this. "hasContent" should just indicate whether there are any image tiles in the given overlay that cover the range associated with a tile. If it returns "false" then no render target is created for that layer (since there will be nothing to write to it) but this won't impact the refinement of the 3d tile geometry.

> Anyway it looks like a bit hacky to be able to return null from calculateLevel. Do you have any idea on how to solve the use case when the user wants to display the overlay only for zoom level >16 for instance?

Returning false should be fine for a case like this but if we limit the return values to numbers then it would just look like this:

```js
layer.mapZoomLevelToTile = level => {

  return level >= 16 ? level : - 1;

};
```

--

Thinking ahead a little bit, as well - one improvement I would like to make at some point is scaling the resolution of the texture based on world-scale size of the geometry. I don't think that impacts this for now, though?

PR #1567, opened 2026-04-19
Export ImageOverlay and TiledImageOverlay

xawill:
This is a proposal to export `ImageOverlay` and `TiledImageOverlay` classes so that library users can override them.

### DIFF
--- a/src/three/plugins/images/ImageOverlayPlugin.js
+++ b/src/three/plugins/images/ImageOverlayPlugin.js
@@ -1276,7 +1276,7 @@ export class ImageOverlayPlugin {
  * @param {boolean} [options.alphaInvert=false] If true, inverts the alpha channel before
  * applying the mask or blend.
  */
-class ImageOverlay {
+export class ImageOverlay {
 
 	get isPlanarProjection() {
 
@@ -1389,7 +1389,7 @@ class ImageOverlay {
  * multiple source tiles into a single texture per 3D tile region.
  * @extends ImageOverlay
  */
-class TiledImageOverlay extends ImageOverlay {
+export class TiledImageOverlay extends ImageOverlay {
 
 	get tiling() {
 


gkjohnson:
How are you planning to use these? I'm okay just exporting them but these classes have been changing recently so I can't promise there won't be breaking changes in the future. The "TiledImageOverlay" class also requires TilingScheme, ProjectionScheme, and an "ImageSource" to function - none of which are exported nor documented.

xawill:
I created a custom `ImageOverlay` to overlay a PNG image on top of quantized mesh terrain. Each pixel of the image encodes the snow depth at defined location. I do some processing on the snow depth data depending on terrain zoom level to smooth a bit the harsh transitions in the snow depth image, since its resolution is pretty low. This allows to have snow depth regions that are a bit interpolated instead of fully-squared.

I can easily imagine other use cases such as implementing another format than the ones currently supported (WMS, WMTS, GeoJSON, etc).

gkjohnson:
> I created a custom ImageOverlay to overlay a PNG image on top of quantized mesh terrain. 

Is this just a single, non tiled PNG image?

That aside I do agree it would be useful to expose these classes at some point - they've just changed a bit recently and I expect they will further with the time I'm going to be putting into MVT support coming up. If you need these classes exported for your use case we can do that but you'll have to be prepared for the APIs to change in patch releases until everything is more properly documented. Otherwise I think just creating a custom "ducktyped" class definition should work for now, which it sounds like you may already be doing?

xawill:
Thanks a lot for the merge!

> Is this just a single, non tiled PNG image?

Correct

> you'll have to be prepared for the APIs to change in patch releases until everything is more properly documented

No worries at all for me.

> Otherwise I think just creating a custom "ducktyped" class definition should work for now, which it sounds like you may already be doing?

I actually currently have my own fork of your library, but they are now converging quickly and I'll soon drop my fork. Thank you for implementing my PR(s) 👍🏻

Issue #943, opened 2025-01-23
Plugins: Add support for other tiled formats


gkjohnson:
### 2D Formats
- [ ] **Protomaps**
  - https://protomaps.com/
  - 2d vectorized data with text / metadata information
- [ ] **Overture Maps** Data
  - https://overturemaps.org/
- [ ] **XYZ MVT** Data
  - OpenStreetMap also provides [vector tiles](https://wiki.openstreetmap.org/wiki/Vector_tiles) in the [MVT format](https://github.com/mapbox/vector-tile-spec/tree/master/2.1).
  - Another vector format? [here](https://openmaptiles.org/schema/)
- [ ] **Cloud Optimized GeoTiff**
  - Tiled TIFF file with different levels of detail embedded
  - https://developers.planet.com/docs/planetschool/an-introduction-to-cloud-optimized-geotiffs-cogs-part-1-overview/
- [ ] **3D OpenStreetMap Data**
  - https://wiki.openstreetmap.org/wiki/3D
- [ ] **Mapbox Tilejson Format**
  - https://github.com/mapbox/tilejson-spec
- [ ] **LERC**
  - Definition: https://github.com/Esri/lerc (including JS parser)
  - Files: https://services.arcgisonline.com/arcgis/rest/services/WorldElevation3D
    - structured as WMTS
- [ ] **KML / KMZ**
    - https://en.wikipedia.org/wiki/Keyhole_Markup_Language
- [ ] **Other**
  - ~Ellipsoid projection~
  - Repeat tiles on X and Y
- [x] **WMS**
    - #1299 
    - #1302 
- [x] **GeoJSON**
    - #1305 
- [x] **WMTS Capabilities**
  - Support the WMTS Capabilities format to infer tile mtrices, etc without explicit user specification or url templating. See sample [here](https://github.com/openlayers/openlayers/blob/main/examples/data/WMTSCapabilities.xml) and GIBS capabilities [here](https://gibs.earthdata.nasa.gov/wmts/epsg4326/best/wmts.cgi?SERVICE=WMTS&request=GetCapabilities).
  - Add a parser to enumerate all the layer information and options / dimensions.
  - Allow for passing the list directly into the WMTSLayer for construction _or_ passing the capabilities url in directly and specifying the keys to use.
  - https://services.arcgisonline.com/arcgis/rest/services
- [x] **WMTS** tiles for rendering 2d map data
  - Can support planar or ellipsoid projection
[...]

PR #1575, opened 2026-05-14
ImageOverlayPlugins: Add MVT Support

gkjohnson:
Based on #1469

Add support for MVT tiles as an image overlay with the goal of supporting MVT and PMTiles formats. MVT is assumed to be served in an XYZ slippy-maps style format.

**Future**
- TerrainRGB (somehow enable PMTiles to be a data source instead of fetch?)
- Hovering, mouse effect
- Consider performance improvements for drawing
  - "bake" paths into a Path2D, preparse all features
  - redraw only visible tiles
  - webworker-based drawing
  - parsed to 3d assets, drawn (ala MapLibre, Overture Maps, Mapbox)
  - skip redrawing tiles that do not include changed styles
- Scale lines, points based on projected world size
- Adjust texture resolution based on world size (possibly fixes above)
- Register MVT as a "texture type" or "drawable type" and PMTiles as a "fetch protocol" of sorts.
- Mark associated tiled formats in #943 

<details>
<summary>Style approach</summary>

```js
styles: {
  rules: [
    { layer: 'water',    type: 'fill', paint: { 'fill-color': '#a0c8f0' } },
    { layer: 'building', type: 'fill', paint: { 'fill-color': '#e0e0e0', 'fill-opacity': 0.8 } },
    { layer: 'road',     type: 'line', paint: (f, l, z) => ({ 'line-width': z > 10 ? 2 : 1, 'line-color': '#888' }) },
    { type: 'fill',   paint: { 'fill-color': '#ffffff', 'fill-opacity': 0.3 } }, // fallback fill
    { type: 'line',   paint: { 'line-color': '#ffffff', 'line-width': 1 } },     // fallback line
    { type: 'circle', paint: { 'circle-color': '#ffffff', 'circle-radius': 3 } } // fallback circle
  ]
}
```
</details>

--

Triangulated geometry for flat or ellipsoidal surfaces will remain for future work since they are not applicable to overlays.

<details>
<summary>Vectorized shape notes</summary>

- Vectorized shapes can be triangulated with earcut
- If projected onto a sphere, the shape can be cut into a grid and projected
- Edges / points are mitered and scaled to screenspace in the vertex shader with offsets
- Once supported we could redraw these textures every frame and scale draped features, lines, etc
</details>


gkjohnson:
Some boundary issues - possibly need rounding to integers:

image

PR #1641, opened 2026-06-16
Docs: fix documented default value of loadAncestors

jledentu:
The `loadAncestors` property (introduced by #1533)  was documented as defaulting to `false` in both the API doc and the JSDoc comment, but the actual implementation initializes it to `true`.

This PR fixes corrects the documented default value to match the code.


gkjohnson:
Great thanks for the catch!

PR #1643, opened 2026-06-17
Fix loadAncestors on tiles with no renderable content

### LINKED ISSUE #1642: loadAncestors fails on tiles without renderable content, causing gaps on camera movement
### Description

On some tilesets, I noticed that the option `loadAncestors` (enabled by default) was causing gaps (empty tiles) on camera movement.

It seems to appear on grouping tiles referencing other tilesets (`content.uri` is a `.json`) when renderable children are large files and take some time to be downloaded.

When cache is limited, the bug may be even more critical since gaps may not disappear.

The issue does not occur in Cesium.

### Reproduction steps

1. Go to https://nasa-ammos.github.io/3DTilesRendererJS/three/cesiumCompare?url=https://tiles.arcgis.com/tiles/z2tnIkrLQ2BRzr6P/arcgis/rest/services/Boston__Integrated_Mesh_3D_Tiles_WGS84/3DTilesServer/tileset.json
2. Move the camera
3. Observe the behavior in left side

### Live example

https://nasa-ammos.github.io/3DTilesRendererJS/three/cesiumCompare?url=https://tiles.arcgis.com/tiles/z2tnIkrLQ2BRzr6P/arcgis/rest/services/Boston__Integrated_Mesh_3D_Tiles_WGS84/3DTilesServer/

### Screenshots & Tile Set

Image

### Library Version

v0.4.28

### Three.js Version

r182

jledentu:
This PR checks that child tile has no renderable content when determining if its content is ready. This prevents gaps when  `loadAncestors` is `true` on tiles referencing `.json` files.

Closes #1642

### DIFF
--- a/src/core/renderer/tiles/traverseFunctions.js
+++ b/src/core/renderer/tiles/traverseFunctions.js
@@ -324,7 +324,7 @@ function markUsedSetLeaves( tile, renderer ) {
 			if ( isUsedThisFrame( c, frameCount ) ) {
 
 				const childCanDisplay = ! canUnconditionallyRefine( c );
-				const childContentReady = ! c.internal.hasContent || isDownloadFinished( c.internal.loadingState );
+				const childContentReady = c.internal.hasUnrenderableContent || isDownloadFinished( c.internal.loadingState );
 				const childIsReady = ( childCanDisplay && childContentReady ) || c.traversal.allChildrenLoaded;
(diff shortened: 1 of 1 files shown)


gkjohnson:
Thanks for the contribution but this fix isn't quite correct - I've fixed the issue in #1644.

PR #1645, opened 2026-06-18
fix: vitest: include .tsx files in test execution

MatrixNeoKozak:
The vitest configuration incorrectly excluded all .tsx files from the test suite, causing the existing test/r3f/typeCheck.test.tsx to be ignored. This change removes the broad exclusion of .tsx files, ensuring all relevant test files in the project are correctly discovered and executed by the test runner.


vitest.config.js
```diff
@@ -8,10 +8,9 @@ export default defineConfig( {
...
 		exclude: [
 			'**/node_modules/**',
-			'**/*.tsx',
 		],
 	},
 	resolve: {
 		alias: packageAliases,
 	},
-} );
+} );
```
gkjohnson: Thanks for the catch! Once this lint issue is fixed we can get this merged.

MatrixNeoKozak:
I have pushed updates to address the feedback and ensure vitest runs smoothly:

1. **Fixed trailing newline:** Added the missing trailing newline to `vitest.config.js` to satisfy the `eol-last` lint rule.
2. **Wrapped type check in a vitest suite:** Since `test/r3f/typeCheck.test.tsx` was now included in the vitest runs, it was causing a failure because it had no actual vitest tests (only TSX declarations for compile-time check). I wrapped the JSX elements in a standard vitest `test` block so that it parses and runs successfully in the test suite while keeping the TypeScript compiler check active.

gkjohnson:
Great, thanks!

PR #1660, opened 2026-07-15
TilesRendererBase: Export default shared cache and queues

WilliamLiu-1997:
Related to https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1661

## Description

`DEFAULT_DOWNLOAD_QUEUE` is already exported, but the other shared default resources are private.

This PR additionally exports:

- `DEFAULT_LRU_CACHE`
- `DEFAULT_PARSE_QUEUE`
- `DEFAULT_NODE_QUEUE`

This allows users to configure global defaults such as cache limits and queue concurrency.

No default values or runtime behavior are changed.


gkjohnson:
We can export these but generally I won't be considering them "public" until they're documented, which I'd like to wait a bit on until this has been proven out a bit more. So feel free to use them but just a heads up that they may change.

PR #1676, opened 2026-08-05
Fix EXT_structural_metadata property tables when buffer views use EXT_meshopt_compression

### LINKED ISSUE #1675: EXT_structural_metadata property tables come back empty when buffer views are meshopt-compressed
`getRelevantBuffers()` in `GLTFStructuralMetadataExtension.js` loads property-table buffer views with `parser.loadBufferView( index )`, which skips the plugin hooks — so `EXT_meshopt_compression` buffer views never get decompressed. `parser.getDependency( 'bufferView', index )` would go through the hooks like three's own extensions do.

The result is that every STRING property decodes as `''` and numeric ones as no-data, without any error. Geometry renders fine since the mesh attributes go through the normal accessor path, so it's easy to miss.

I ran into this with the Dutch Kadaster/PDOK 3D tiles, which compress the metadata buffer views (they're generated with 3DGI's tyler). 

### Reproduction

```js
import { TilesRenderer } from '3d-tiles-renderer';
import { GLTFExtensionsPlugin } from '3d-tiles-renderer/plugins';
import { MeshoptDecoder } from 'three/examples/jsm/libs/meshopt_decoder.module.js';

const tiles = new TilesRenderer(
	'https://api.pdok.nl/kadaster/3d-basisvoorziening/ogc/v1/collections/gebouwen/3dtiles'
);
tiles.registerPlugin( new GLTFExtensionsPlugin( { meshoptDecoder: MeshoptDecoder } ) );

tiles.addEventListener( 'load-model', ( { scene } ) => {

	const sm = scene.userData.structuralMetadata;
	console.log( sm.getPropertyTableData( [ 0 ], [ 0 ] ) );
	// actual:   { identificatie: '', oorspronkelijkBouwjaar: <no-data>, ... }
	// expected: { identificatie: 'NL.IMBAG.Pand.0331100000251700', oorspronkelijkBouwjaar: 1959, ... }

} );
```

Seen on 0.5.0, but the code on master looks the same. I'll open a PR.

Nahuel-M:
Fixes #1675.

`getRelevantBuffers()` loaded property-table buffer views with `parser.loadBufferView( index )`, which bypasses the GLTF loader plugin hooks — so buffer views compressed with `EXT_meshopt_compression` were never decompressed and property values silently decoded as empty/no-data. Going through `parser.getDependency( 'bufferView', index )` runs the hooks, the same way three's own extensions load buffer views.

Verified against the PDOK 3D Tiles from the issue: property values (strings and numerics) now decode correctly, and behavior for uncompressed files is unchanged.

### DIFF
--- a/src/three/plugins/gltf/GLTFStructuralMetadataExtension.js
+++ b/src/three/plugins/gltf/GLTFStructuralMetadataExtension.js
@@ -41,19 +41,19 @@ function getRelevantBuffers( parser, propertyTables = [] ) {
 			const { values, arrayOffsets, stringOffsets } = properties[ key ];
 			if ( result[ values ] === null ) {
 
-				result[ values ] = parser.loadBufferView( values );
+				result[ values ] = parser.getDependency( 'bufferView', values );
 
 			}
 
 			if ( result[ arrayOffsets ] === null ) {
 
-				result[ arrayOffsets ] = parser.loadBufferView( arrayOffsets );
+				result[ arrayOffsets ] = parser.getDependency( 'bufferView', arrayOffsets );
 
 			}
 
 			if ( result[ stringOffsets ] === null ) {
 
-				result[ stringOffsets ] = parser.loadBufferView( stringOffsets );
+				result[ stringOffsets ] = parser.getDependency( 'bufferView', stringOffsets );
 
 			}
 


gkjohnson:
Great, thank you!

PR #1680, opened 2026-08-07
Fix Orthographic camera zoomSpeed calculation

### LINKED ISSUE #1679: EnvironmentControls: Changing orthographic camera zoomSpeed causes incorrect zoom values
### Description

Changing the orthographic camera zoomSpeed in EnvironmentControls to any number that isn't 1 causes the camera zoom value to be incorrect on change.

It looks like moving the zoomSpeed calculation to[ the normalizedDelta calculation](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/3dc8b0a007d9596e45623abb7c65bb0b1230a831/src/three/renderer/controls/EnvironmentControls.js#L1285) as follows may fix the issue:

```js
const normalizedDelta = Math.pow( 0.95, Math.abs( scale * 0.05 ) * zoomSpeed );
let scaleFactor = scale > 0 ? 1 / Math.abs( normalizedDelta ) : normalizedDelta;
// scaleFactor *= zoomSpeed;
```

### in three
1. In example/three/deepZoom.js add controls.zoomSpeed = 10
2. Zoom in/out

### In r3f
1. In example/r3f/basic.jsx change the camera to orthographic
2. Change camera start position to [0, 10, 0]
3. Change zoomSpeed on <EnvironmentControls> to 10
4. Zoom in/out

### Library Version

v0.5.1

### Three.js Version

v0.185.0

heycharlieabbott:
Fixes https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/1679

Apply orthographic zoomSpeed in normalizedDelta calculation to prevent flipped/runaway zoom values

### DIFF
--- a/src/three/renderer/controls/EnvironmentControls.js
+++ b/src/three/renderer/controls/EnvironmentControls.js
@@ -1283,8 +1283,8 @@ export class EnvironmentControls extends EventDispatcher {
 			_mouseBefore.unproject( camera );
 
 			// zoom the camera
-			const normalizedDelta = Math.pow( 0.95, Math.abs( scale * 0.05 ) * zoomSpeed );
-			let scaleFactor = scale > 0 ? 1 / Math.abs( normalizedDelta ) : normalizedDelta;
+			let scaleFactor = Math.pow( 0.95, - zoomSpeed * scale * 0.05 );
+
 			if ( scaleFactor > 1 ) {
 
 				if ( maxZoom < camera.zoom * scaleFactor ) {
--- a/src/three/renderer/controls/EnvironmentControls.js
+++ b/src/three/renderer/controls/EnvironmentControls.js
@@ -1283,10 +1283,8 @@ export class EnvironmentControls extends EventDispatcher {
 			_mouseBefore.unproject( camera );
 
 			// zoom the camera
-			const normalizedDelta = Math.pow( 0.95, Math.abs( scale * 0.05 ) );
+			const normalizedDelta = Math.pow( 0.95, Math.abs( scale * 0.05 ) * zoomSpeed );
 			let scaleFactor = scale > 0 ? 1 / Math.abs( normalizedDelta ) : normalizedDelta;
-			scaleFactor *= zoomSpeed;
-
 			if ( scaleFactor > 1 ) {
 
 				if ( maxZoom < camera.zoom * scaleFactor ) {
--- a/src/three/renderer/controls/GlobeControls.js
+++ b/src/three/renderer/controls/GlobeControls.js
@@ -678,11 +678,10 @@ export class GlobeControls extends EnvironmentControls {
 			this._alignCameraUpToNorth( MathUtils.lerp( 0, 0.2, distanceAlpha * deltaAlpha ) );
 
 			const scale = this.zoomDelta;
-			const normalizedDelta = Math.pow( 0.95, Math.abs( scale * 0.05 ) );
-			const scaleFactor = scale > 0 ? 1 / Math.abs( normalizedDelta ) : normalizedDelta;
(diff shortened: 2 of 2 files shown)


gkjohnson:
Hey! Charlie Abbott! How's it going -

I just tested this and I agree the current behavior isn't right. I don't think we should embed the scaling factor inside the normalized delta exponent, though. The "zoomSpeed" value is allowed to be negative and should generally be a value that linearly scales the apparent "zoom" change. 

I'm thinking this would be better - the "normalizedDelta" value is instead scaled by the zoomSpeed magnitude. And the zoom direction is also adjusted by the zoomSpeed sign:

```js
const normalizedDelta = Math.pow( 0.95, Math.abs( scale * 0.05 ) ) * Math.abs( zoomSpeed );
let scaleFactor = scale * zoomSpeed > 0 ? 1 / Math.abs( normalizedDelta ) : normalizedDelta;
```

heycharlieabbott:
Hey hey! Going well, hope all's well with you too!

Thanks for checking and for the feedback! I've just tested it, and unless I've missed something, it looks like setting the zoomSpeed to 0 from outside of the normalized delta exponent sets the scaleFactor to 0, then when [the camera zoom is multiplied by the scaleFactor](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/3dc8b0a007d9596e45623abb7c65bb0b1230a831/src/three/renderer/controls/EnvironmentControls.js#L1308), the orthographic camera zoom is [by default set to 0](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/3dc8b0a007d9596e45623abb7c65bb0b1230a831/src/three/renderer/controls/EnvironmentControls.js#L159) and the scene becomes invisible unless the minZoom is set to a positive number. 

Also, unless I've misunderstood something with the above change it looks like setting the zoomSpeed to somewhere between 0 and 1 or 0 and -1 can cause an overall increase in the zoomSpeed, which feels a bit counterintuitive to me.

If we keep the zoomSpeed in the normalized delta exponent, when set to 0 normalizedDelta will be evaluated to 1, which leads to the scaleFactor being unchanged and the orthographic camera being unable to zoom, also any number between 0 and 1 or 0 and -1 will slow the camera zoomSpeed and any number larger than 1 or -1 will increase the zoomSpeed which feels a bit more intuitive to me. 

Flipping the zoom direction by zoomSpeed sign totally makes sense to me though! What do you think about the following?

```js
	const normalizedDelta = Math.pow( 0.95, Math.abs( scale * 0.05 ) * Math.abs( zoomSpeed ) );
	let scaleFactor = scale * zoomSpeed > 0 ? 1 / Math.abs( normalizedDelta ) : normalizedDelta;
```

In local testing, considering a zoomSpeed of 1 as default, setting as 0.5 did appear to be about half speed, setting as 2 appeared about double speed, 0 was stopped, and negative flipped the zoom direction.

gkjohnson:
You're right - I was mixing this "normalizedDelta" up with the scroll wheel value normalization elsewhere.

Your change looks good but it occurs to me at this can be simplified to the following so the exponent sign just implicitly handles the fraction inversion:

```js
let scaleFactor = Math.pow( 0.95, - zoomSpeed * scale * 0.05 );
```

We'll also need to adjust the GlobeControls case here, as well:

https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/3dc8b0a007d9596e45623abb7c65bb0b1230a831/src/three/renderer/controls/GlobeControls.js#L680-L682

heycharlieabbott:
oh nice! I've updated EnvironmentControls to the simplified version and updated GlobeControls too. Thank you!

gkjohnson:
Great, thanks!

PR #1716, opened 2026-08-28
TilesRenderer: Preserve group transform during tile fade out

### LINKED ISSUE #1715: TilesFadePlugin: fading-out tiles lose the TilesRenderer.group transform in v0.5.1
### Description

In `3d-tiles-renderer` v0.5.1, a tile using `TilesFadePlugin` can jump back to its original/local position when fade-out starts if `TilesRenderer.group` has a non-identity transform.

This appears to be a regression introduced by #1670.

### Versions

- `3d-tiles-renderer`: 0.5.1
- `three`: 0.185.1
- `TilesFadePlugin` enabled

### Steps to reproduce

1. Create a `TilesRenderer`.
2. Register `TilesFadePlugin`.
3. Apply a translation or another non-identity transform to `tiles.group`.
4. Display a tile normally.
5. Move the camera so that the tile starts fading out.

### Actual behavior

As soon as fade-out starts, the tile loses the `TilesRenderer.group` transform and fades out at its original/local position.

After the transition completes or the tile becomes visible again, its position is correct again.

### Expected behavior

The tile should retain the `TilesRenderer.group` transform for the entire fade-out animation.

### Cause

Traversal calls `setTileActive(tile, false)` before `setTileVisible(tile, false)`.

Since v0.5.1, `TilesRenderer.setTileActive(false)` immediately executes:

```js
scene.parent = null;

WilliamLiu-1997:
## Summary

- keep visible inactive tile scenes parented to `TilesRenderer.group` during fade out
- detach the scene only after it is both inactive and hidden
- add a regression test covering the active-to-fading visibility transition
- document the fix in the changelog

This preserves the tileset group transform while `TilesFadePlugin` delays the final visibility change.

Fixes #1715

## Tests

- `npm test` (619 tests passed)
- `npm run lint` (passes with 4 pre-existing warnings)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - MVTAnnotationsPlugin: Add a `useIdleCallback` option.
 
 ### Fixed
+- TilesRenderer, TilesFadePlugin: Keep fading-out tiles parented to the tiles renderer group until they are hidden.
 - EnvironmentControls, GlobeControls: Fix orthographic camera zoom when adjusting zoomSpeed.
 - GeoJSONOverlay: Fix content bounds check treating the normalized tile range as radians, causing every tile to report content.
 - MVTAnnotationsPlugin: Adjust plugin to prefetch vector tile content.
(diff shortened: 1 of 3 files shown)


gkjohnson:
Great, thanks for the catch!

PR #1723, opened 2026-09-04
GlobeControls: Fix trackball drift when rotating the globe from afar

### LINKED ISSUE #808: GlobeControls: Remove "trackball drift" when rotating globe from afar
Ie when moving the cursor back to the same place the globe orientation should match where it started.

clement-igonet:
Fixes #808.

Each drag frame applied the minimal-arc rotation (`setFromUnitVectors`) between the current cursor direction and the pivot direction, composed incrementally onto the camera. Minimal-arc rotations on a sphere are path dependent, so dragging the cursor in a loop and returning to the starting point accumulated a net roll about the pivot axis and the globe did not return to its starting orientation.

The drag is now absolute, arcball style: the camera pose is captured as a baseline on the first drag frame, the cursor ray is cast from that baseline pose so the cursor-to-globe mapping stays stable for the whole drag, and the single total rotation from the current cursor direction to the grab direction is applied to the baseline. 

Same closed-loop drag on both versions, cursor back at its starting point:

| | before drag | after drag |
|---|---|---|
| master | master before | master after |
| this PR | fix before | fix after |

https://github.com/user-attachments/assets/5b846e69-8579-46dc-9ee4-8ae115ee0815

### DIFF
--- a/src/three/renderer/controls/GlobeControls.js
+++ b/src/three/renderer/controls/GlobeControls.js
@@ -22,6 +22,7 @@ const _forward = /* @__PURE__ */ new Vector3();
 const _targetRight = /* @__PURE__ */ new Vector3();
 const _globalUp = /* @__PURE__ */ new Vector3();
 const _quaternion = /* @__PURE__ */ new Quaternion();
+const _quaternion2 = /* @__PURE__ */ new Quaternion();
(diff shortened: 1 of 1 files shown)


gkjohnson:
Looks great! Thank you! This is a nice improvement

PR #1730, opened 2026-09-06
ImageOverlayPlugin: Fix tile splitting stalling over tiles with low geometric error

### LINKED ISSUE #1636: ImageOverlayPlugin: tile splitting stalls over flat quantized-mesh terrain because split children inherit the terrain tile's geometricError
### Description

With `enableTileSplitting` (default on), `ImageOverlayPlugin` extends overlay refinement past the geometry leaves by creating virtual split children. Those children are created with:

https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/v0.4.28/src/three/plugins/images/ImageOverlayPlugin.js#L761-L769

i.e. `geometricError: tile.geometricError * 0.5`. This couples overlay refinement to the *terrain mesh* error, which has no relationship to the overlay's remaining image resolution.

Over flat ground, quantized-mesh terrain (Cesium World Terrain) reports honest, tiny mesh errors — at the repro location below the deepest real tiles (depth 14) have `geometricError` ≈ **0.21–0.6 m**. At ~100 m altitude that's at or below `errorTarget = 2`, so traversal never (or barely) elects to refine into the split children at all. The overlay's `shouldSplit()` still returns true (`calculateLevel(range) < tiling.maxLevel`, [src](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/v0.4.28/src/three/plugins/images/ImageOverlayPlugin.js#L1629-L1633)) — the splits are created, but traversal's SSE test discards them because their inherited error is centimeter-scale.

Net effect: imagery is capped at the terrain leaf's level and stays visibly blurry at street-level altitudes, even though the overlay has many more native levels available. In my measurements (details below): visible imagery capped at **level 14 of maxLevel 20**, with **0 of 80** created split tiles ever loaded.

To confirm the mechanism I registered a small plugin that only floors each terrain tile's `geometricError` to the standard level ladder (`~77067 / 2^level` m) in `processTileModel`, changing nothing else. With that single change the splits recurse correctly (nested splits, 6 levels deep) and visible imagery reaches levels 19–20 — so the split machinery itself works; it's only the inherited `geometricError` that prevents it from ever being selected.

| | stock v0.4.28 | with terrain `geometricError` floored to level ladder |
|---|---|---|
| split tiles loaded | 0 (80 created, all at chain depth 1) | recursion to chain depth 6, `virtualChildCount` 0→4 as expected |
| imagery level on visible tiles | ≤ 14 | up to 20 (= overlay `tiling.maxLevel`) |

Driving `errorTarget` down instead doesn't realistically help: reaching imagery level 19 here would need `errorTarget` ≈ 0.1 px, which explodes geometry refinement globally (related: the `errorTarget = 0.25` workaround mentioned at the end of #1334).

This is adjacent to the level-mapping discussion in #1278 but narrower and concrete: the existing tile-splitting feature is effectively non-functional over flat quantized-mesh terrain because of the inherited error value. A possible direction, in line with the texel-size ideas in #1278: floor the split child's `geometricError` at the overlay's projected texel size for the child's image level (the same world-space-pixel notion `EllipsoidProjectionTilesPlugin` uses for image-tile geometric error), e.g. `max(parentError * 0.5, texelError(level + 1))`, so splits keep being selected exactly while the overlay still has native resolution to show.

(While debugging I initially suspected the `alreadySplit` gate at [L613](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/v0.4.28/src/three/plugins/images/ImageOverlayPlugin.js#L613), since split children are created without `internal.virtualChildCount` — but `preprocessNode` pre-seeds the defaults before any content can load, so the gate behaves correctly and nested expansion works once traversal refines. Verified against v0.4.28 and current master (2d21dc5).)

### Reproduction steps

1. Load Cesium World Terrain (ion asset 1, quantized mesh) with a `CesiumIonOverlay` (e.g. Bing Aerial, ion asset 2) via `ImageOverlayPlugin`; leave `enableTileSplitting` on and `errorTarget` at 2.
2. Place the camera ~100 m above flat ground — e.g. Dubai, lat 25.118 / lon 55.133, looking down at ~45°.
3. Let streaming settle, then observe the ground imagery stays blurry (~level 14). Walking the tile tree shows the `image_overlay_tile_split` children exist but never load; the imagery level on visible tiles never approaches the overlay's `maxLevel`.
4. (Mechanism check) Floor terrain tiles' `geometricError` to the level ladder in a `processTileModel` hook → splits now recurse and imagery reaches level 19–20 in the same scene.

### Code

```js
import { TilesRenderer } from '3d-tiles-renderer';
import { CesiumIonAuthPlugin } from '3d-tiles-renderer/core/plugins';
import { ImageOverlayPlugin, CesiumIonOverlay, QuantizedMeshPlugin } from '3d-tiles-renderer/plugins';

const tiles = new TilesRenderer();
tiles.registerPlugin( new CesiumIonAuthPlugin( {
	apiToken: ION_TOKEN,
	assetId: '1', // Cesium World Terrain
	assetTypeHandler: ( type, tilesBase ) => {

		if ( type === 'TERRAIN' && tilesBase.getPluginByName( 'QUANTIZED_MESH_PLUGIN' ) === null ) {

			tilesBase.registerPlugin( new QuantizedMeshPlugin( { useRecommendedSettings: true } ) );

		}

	},
} ) );
tiles.registerPlugin( new ImageOverlayPlugin( {
	renderer,
	overlays: [ new CesiumIonOverlay( { assetId: 2, apiToken: ION_TOKEN } ) ], // Bing Aerial
} ) );
tiles.errorTarget = 2;

// camera ~100 m above flat ground (Dubai): lat 25.118, lon 55.133, tilted ~45deg

// --- diagnostics used for the numbers above (run after streaming settles) ---
const splitRe = /image_overlay_tile_split/;
const chains = {};
( function walk( tile, chain ) {

	const uri = tile.content?.uri ?? '';
	const c = splitRe.test( uri ) ? chain + 1 : 0;
	if ( c ) {

		chains[ c ] = chains[ c ] || { n: 0, loaded: 0 };
		chains[ c ].n ++;
		if ( tile.internal?.loadingState === 4 /* LOADED */ ) chains[ c ].loaded ++;

	}
	tile.children.forEach( ch => walk( ch, c ) );

} )( tiles.root, 0 );

const plugin = tiles.getPluginByName( 'IMAGE_OVERLAY_PLUGIN' );
const overlay = plugin.overlays[ 0 ];
const info = plugin.overlayInfo.get( overlay );
const levels = {};
tiles.visibleTiles.forEach( tile => {

	const ti = info.tileInfo.get( tile );
	if ( ti?.range ) {

		const lvl = overlay.calculateLevel( ti.range );
		levels[ lvl ] = ( levels[ lvl ] || 0 ) + 1;

	}

} );
console.log( { chains, levels, maxLevel: overlay.tiling.maxLevel } );

// --- the "mechanism check" plugin that makes splitting work ---
const LEVEL0_ERROR = ( 6378137 * 2 * Math.PI * 0.25 ) / ( 65 * 2 );
tiles.registerPlugin( {
	name: 'TERRAIN_ERROR_FLOOR',
	priority: - 500,
	processTileModel( scene, tile ) {

		const match = /(\d+)\/\d+\/\d+\.terrain/.exec( tile.content?.uri ?? '' );
		if ( match ) tile.geometricError = Math.max( tile.geometricError, LEVEL0_ERROR / 2 ** Number( match[ 1 ] ) );

	},
} );
```

### Live example

Not possible to share a runnable fiddle since the repro needs a Cesium ion token — the snippet above is complete apart from `ION_TOKEN` and standard three.js scene boilerplate (the globe fiddle template from the issue form works as a base).

### Screenshots & Tile Set

Measured numbers in the table above (Dubai 25.118 / 55.133, camera 100 m, 45° tilt, errorTarget 2, ~750 px viewport height):

- stock: terrain leaves at depth 14 with `geometricError` 0.21–0.6 m; imagery levels on visible tiles `{ …, 13: 3, 14: 4 }`, nothing deeper; 80 `image_overlay_tile_split` tiles in the tree, all chain depth 1, 0 loaded.
- with the geometricError floor only: split chains to depth 6; imagery levels on visible tiles include `{ 18: 3, 19: 18, 20: 8 }`.

Tile sets: Cesium ion assets 1 (Cesium World Terrain) + 2 (Bing Maps Aerial).

### Library Version

v0.4.28 (also checked the relevant source on master @ 2d21dc5 — unchanged)

### Three.js Version

r184

clement-igonet:
Fixes #1636.

Virtual split children were created under tiles whose own `geometricError` stays that of the source geometry. Over flat quantized-mesh terrain that error is centimeter-scale, so traversal never elected to enter or select the split children (children are only traversed while the parent's error exceeds the error target) and imagery stayed capped at the terrain leaf level.

When split children are created, the tile's `geometricError` is now raised to at least the world-space size of one texel of its composited overlay texture (content size divided by the plugin's `resolution`), and restored when the splits are removed. Each split child inherits half of that as before, so the floor halves with each split level, and `shouldSplit()` still stops the recursion at the overlays' max level, so no image tiles are loaded beyond the native resolution of the overlays. The net change is 22 lines.

Before/after on a keyless repro (four flat tiles with `geometricError` 0.3 m, like flat quantized-mesh leaves, under an OSM overlay via `ImageOverlayPlugin`, camera 600 m):

| master | this PR |
|---|---|
| master: imagery frozen at the level shown when splits were created | fix: splits refine to the overlay's native resolution |

Unit tests cover the new behavior: the error is raised to the texel size when splits are created, untouched when no split happens, and restored when splits are removed. Lint and the `test/three` suite pass.

### DIFF
--- a/src/three/plugins/images/ImageOverlayPlugin.js
+++ b/src/three/plugins/images/ImageOverlayPlugin.js
@@ -28,6 +28,7 @@ const _box = /* @__PURE__ */ new Box3();
 const SPLIT_TILE_DATA = Symbol( 'SPLIT_TILE_DATA' );
 const SPLIT_HASH = Symbol( 'SPLIT_HASH' );
 const ORIGINAL_REFINE = Symbol( 'ORIGINAL_REFINE' );
+const ORIGINAL_GEOMETRIC_ERROR = Symbol( 'ORIGINAL_GEOMETRIC_ERROR' );
 
 const PROCESS_QUEUE = /* @__PURE__ */ new PriorityQueue();
 PROCESS_QUEUE.maxJobs = 10;
@@ -282,6 +283,13 @@ export class ImageOverlayPlugin {
 		tile.refine = tile[ ORIGINAL_REFINE ];
 		delete tile[ ORIGINAL_REFINE ];
 		delete tile[ SPLIT_HASH ];
+
+		if ( ORIGINAL_GEOMETRIC_ERROR in tile ) {
+
+			tile.geometricError = tile[ ORIGINAL_GEOMETRIC_ERROR ];
+			delete tile[ ORIGINAL_GEOMETRIC_ERROR ];
+
+		}
 
 	}
 
@@ -655,6 +663,20 @@ export class ImageOverlayPlugin {
 			} );
 
 		} );
+
+		// The tile's overlays are composited into textures "resolution" texels across, so one
+		// texel spans roughly the content size divided by that resolution. Raise the tile's
+		// error to at least that value while split children exist so traversal keeps entering
+		// and selecting them (children are only traversed while the parent error exceeds the
+		// error target), and restore it when the splits are removed. Each split child inherits
+		// half, and "shouldSplit" stops the recursion at the overlays' max level (#1636).
+		if ( ! ( ORIGINAL_GEOMETRIC_ERROR in tile ) ) {
+
+			const texelError = _box.setFromObject( clone ).getSize( _vec ).length() / this.resolution;
+			tile[ ORIGINAL_GEOMETRIC_ERROR ] = tile.geometricError;
+			tile.geometricError = Math.max( tile.geometricError, texelError );
+
+		}
 
 		// run the clipping operations by performing every permutation of sides
 		// defined by the split directions
--- a/src/three/plugins/images/ImageOverlayPlugin.js
+++ b/src/three/plugins/images/ImageOverlayPlugin.js
@@ -670,7 +670,9 @@ export class ImageOverlayPlugin {
 		// and selecting them (children are only traversed while the parent error exceeds the
 		// error target), and restore it when the splits are removed. Each split child inherits
 		// half, and "shouldSplit" stops the recursion at the overlays' max level (#1636).
-		if ( ! ( ORIGINAL_GEOMETRIC_ERROR in tile ) ) {
+		// Additive tiles are excluded: they can have real children, and raising their error
+		// would make traversal load that child geometry far deeper than the tile set intends.
+		if ( tile.refine !== 'ADD' && ! ( ORIGINAL_GEOMETRIC_ERROR in tile ) ) {
 
 			const texelError = _box.setFromObject( clone ).getSize( _vec ).length() / this.resolution;
 			tile[ ORIGINAL_GEOMETRIC_ERROR ] = tile.geometricError;
--- a/test/three/ImageOverlayPlugin.test.js
+++ b/test/three/ImageOverlayPlugin.test.js
@@ -1,0 +1,96 @@
+import { Group, Matrix4, Mesh, MeshBasicMaterial, PlaneGeometry } from 'three';
+import { ImageOverlayPlugin } from '../../src/plugins.js';
+import { WGS84_ELLIPSOID } from '../../src/three/renderer/math/GeoConstants.js';
+
+// Builds the minimal structures needed to exercise the tile splitting logic
+// without a renderer: a plugin with one stub overlay that always wants to
+// split, and a loaded tile scene containing a single 100 x 100 m quad.
+function createSplitFixture() {
+
+	const plugin = new ImageOverlayPlugin( { resolution: 256 } );
+	plugin.tiles = {
+		ellipsoid: WGS84_ELLIPSOID,
+		processNodeQueue: { remove() {} },
+		lruCache: { remove() {} },
+	};
+
+	const tile = {
+		refine: 'REPLACE',
+		geometricError: 0.3,
+		boundingVolume: { box: [ 0, 0, 0, 50, 0, 0, 0, 50, 0, 0, 0, 1 ] },
+		children: [],
+		content: { uri: './tile.glb' },
+		internal: { virtualChildCount: 0 },
+		engineData: { transformInverse: new Matrix4() },
+	};
+
+	const overlay = { frame: null, shouldSplit: () => true };
+	plugin.overlayInfo.set( overlay, {
+		tileInfo: new Map( [[ tile, { target: {}, range: [ 0, 0, 1, 1 ] } ]] ),
+	} );
+
+	const scene = new Group();
+	const mesh = new Mesh( new PlaneGeometry( 100, 100 ), new MeshBasicMaterial() );
+	mesh.position.set( WGS84_ELLIPSOID.radius.x, 0, 0 );
+	scene.add( mesh );
+	scene.updateMatrixWorld( true );
+
+	return { plugin, tile, scene };
+
+}
+
+describe( 'ImageOverlayPlugin tile splitting', () => {
+
+	it( 'raises the tile geometric error to the overlay texel size while splits exist', async () => {
+
+		const { plugin, tile, scene } = createSplitFixture();
+		await plugin.expandVirtualChildren( scene, tile );
+
+		// one texel of the 256 px overlay texture spans contentSize / resolution,
+		// here the diagonal of a 100 x 100 quad divided by 256
+		const texelError = Math.sqrt( 2 * 100 * 100 ) / 256;
+		expect( tile.children.length ).toBeGreaterThan( 0 );
+		expect( tile.internal.virtualChildCount ).toBe( tile.children.length );
+		expect( tile.geometricError ).toBeCloseTo( texelError, 10 );
+
+		// split children inherit half the raised error, keeping the ladder halving
+		tile.children.forEach( child => {
+
+			expect( child.geometricError ).toBeCloseTo( texelError / 2, 10 );
+
+		} );
+
+	} );
+
+	it( 'does not raise the error of tiles that produce no splits', async () => {
+
+		const { plugin, tile, scene } = createSplitFixture();
+		for ( const info of plugin.overlayInfo.values() ) {
+
+			info.tileInfo.get( tile ).target = null;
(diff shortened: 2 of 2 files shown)


clement-igonet:
While building a keyless repro for this I found the first commit was not sufficient on its own: traversal only enters a tile's children when the tile's own error exceeds the error target, so flooring the child error never opened the gate at the first split level. The second commit also raises the split parent's `geometricError` to one texel at its current display level (restored when the splits are removed), which is the same thing the mechanism check in #1636 did externally.

Before/after on the repro (four flat tiles with `geometricError` 0.3 m, like flat quantized-mesh leaves, under an OSM overlay via `ImageOverlayPlugin`, camera 600 m):

| master | this PR |
|---|---|
| master: imagery frozen at the level shown when splits were created | fix: splits refine to the overlay's native resolution |

The repro is a ~100 line page plus a tiny generated tileset; happy to share it in the thread or as a fiddle if useful.

src/three/plugins/images/ImageOverlayPlugin.js
```diff
@@ -654,6 +662,20 @@ export class ImageOverlayPlugin {
...
+		// and selecting them (children are only traversed while the parent error exceeds the
+		// error target), and restore it when the splits are removed. Each split child inherits
+		// half, and "shouldSplit" stops the recursion at the overlays' max level (#1636).
+		if ( ! ( ORIGINAL_GEOMETRIC_ERROR in tile ) ) {
+
+			const texelError = _box.setFromObject( clone ).getSize( _vec ).length() / this.resolution;
+			tile[ ORIGINAL_GEOMETRIC_ERROR ] = tile.geometricError;
+			tile.geometricError = Math.max( tile.geometricError, texelError );
+
+		}
```
gkjohnson, line 677: Thanks for this - I think at first glance this is an okay approach to dealing with reaching leaf tiles that have a low error values, and it's a probably a better solution that boosting the error with the "EnforceNonZeroErrorPlugin", but a couple wrinkles that come time mind: 

- Additive tiles need to be split _and_ can have children, so blindly raising their error based on the tile overlay resolution could result in causing a ton of child geometry to be loaded (in addition to being split) at a far higher rate than the tile set is designed for, which could cause both bandwidth and performance issues.
- Even when working with just replacement leaf tiles it's possible that raising the error this way could also an explosion in split children depending on the relationship between the 3d tile geometric error and the overlay error derived from resolution.

I'm not sure if there are any solutions that come to mind here.

--

Dealing with mapping the image overlays to the geometric tiles has been a pretty significant point of difficulty due to the difference in error metrics. I believe MapLibre loads image tile LoD based on something like a "distance to ellipsoid" value, but it's not a greate solution when zooming in close to a high altitude surface. And generating a fixed texture per-geometry tile can result in the image-tile LoD "jumps" depending on the geometric error and child tile sizes between 3d tiles layers.

I've toyed around with the idea of loading image tiles based on some kind of distance-to-geometry metric and sampling the image overlay LoD hierarchy in-shader to allow for smoother transitions within a single tile but I haven't had the bandwidth to think this through too fully. This would let us get rid of the need to split tiles entirely, though.

clement-igonet: Thanks, both points are fair. Additive tiles were indeed reaching the boost through the `refine === 'ADD'` branch of the split gate, and raising their error would over-refine their real children. Pushed f336f359, which skips the boost for additive tiles (they still get split children as before, error untouched), with a test.

On the split explosion: once additive tiles are excluded, the boosted tiles are childless replacement leaves, so the raised error only drives the virtual split ladder, which is the same one that already runs on tiles with a naturally large error (halving per level, entered only while above the error target, capped by `shouldSplit` at the overlay max level). Split count ends up screen-limited, roughly `viewport pixels / resolution²` fully refined tiles.

Agreed the distance-based image LoD with in-shader hierarchy sampling is the better long-term design. Happy to keep this as the stopgap until then, or to look into that direction if it helps.

clement-igonet: Thanks for the review. On the two wrinkles:

**Additive tiles.** Agreed. Pushed a commit that leaves the geometric error of `ADD` tiles untouched (they are still split, only the raise is skipped), with a unit test. Flat additive leaves keep today's behavior, which seems the right trade-off since they can carry real children.

**Split explosion.** The raise is bounded by construction: the floor is `contentDiagonal / resolution`, i.e. the error a tile of that size would need to declare for its overlay texture to be sampled 1:1. So a tile never gets more splits than a tile of the same size that already declares that error today, and the depth is still capped by `overlay.shouldSplit` (the overlays' max level) and the SSE target, both unchanged. Measured on the repro from my earlier comment, with a control tileset where the leaves declare 50 m on master:

| | leaf error | visible tiles | splits created | split depth |
|---|---|---|---|---|
| master, 0.3 m leaves | 0.3 | 4 | 16 (0 selected) | 2 |
| this PR, 0.3 m leaves | raised to 45.5 | 28 | 336 | 7 |
| master, 50 m leaves (control) | 50 | 28 | 400 | 7 |

Same frame, same depth, slightly fewer splits than the control. The PR does not add a regime the plugin does not already have, it makes flat tiles behave like the others.

On sampling the overlay LoD in-shader from a distance-to-geometry metric: agreed that would remove splitting altogether and I would be glad to help with it, but it is a much larger change, so I would suggest landing this as the fix for #1636 in the meantime.

PR #1731, opened 2026-09-07
SUBTREELoader, ImplicitTilingPlugin: Add support for subtree json files

Om-singhaI:
Adds support for subtree files in the json format. It's the subtree json item from the list in #608.

Most of it is wiring. A subtree json has no `root` property, so the core already falls past the external tileset branch and hands it to `parseTile`. `SUBTREELoader.parse` now takes either an ArrayBuffer or an already parsed object, and the plugin sends json through to it.

I used `tile.implicitTilingData` to tell a subtree json from other json reaching the plugin. It's the only thing on the tile at that point that says it belongs to an implicit tileset, and it's close to the condition the list asks for in place of the `.subtree` extension check. Disposal keys off the same thing, so the children generated under a json subtree still get cleared. It won't catch a tileset serving its content as json as well, since then nothing separates the two. Happy to change it if you'd rather key off something else.

There's no test for either file today, so this adds some. They build the same subtree twice, once as a binary buffer and once as a plain object, then check both expand to the same tiles, covering constant availability with no buffers and bitstreams in an external buffer or in the binary chunk, with the fetch stubbed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - MVTAnnotationsPlugin: Add support for changing "resolution" after initialization.
 - GlobeControls: Adjust the globe controls to use stable orientation during rotation.
+- ImplicitTilingPlugin: Add support for subtree files in the json format.
 
 ### Fixed
 - MVTAnnotationsPlugin: Fix line features without a name or id sharing an id, causing them to be merged.
--- a/src/core/plugins/ImplicitTilingPlugin.js
+++ b/src/core/plugins/ImplicitTilingPlugin.js
@@ -52,7 +52,9 @@ export class ImplicitTilingPlugin {
 
 	parseTile( buffer, tile, extension ) {
 
-		if ( /^subtree$/i.test( extension ) ) {
+		// json content arrives already parsed rather than as a buffer
+		const isJsonSubtree = /^json$/i.test( extension ) && this.isSubtreeJson( tile );
+		if ( /^subtree$/i.test( extension ) || isJsonSubtree ) {
 
 			const loader = new SUBTREELoader( tile );
 			loader.workingPath = tile.internal.basePath;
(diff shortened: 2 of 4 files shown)


gkjohnson:
Thanks for the contribution. How have you tested this? Do you have a demo data set to try?

Om-singhaI:
The tests in the PR build the same subtree twice, once as a binary `.subtree` buffer and once as a plain object, then check both expand to the same tiles. That way the json path is measured against the binary one rather than against what I expected. They cover constant availability with no buffers, and a bitstream held in an external buffer or in the binary chunk, with the fetch stubbed.

Those were all made up though, so on your second question I went and used `ImplicitTilesetWithJsonSubtree` from Cesium's spec data, the one linked in the checklist. Its subtrees are `subtrees/{level}.{x}.{y}.json` and it expands to the five `.b3dm` tiles the fixture describes. I've just pushed that as another test with the tileset and subtree copied in, so it runs without a network fetch, and it fails on master like the others.

I haven't tried the two ion assets you mentioned, since I don't have a key. Happy to add an example page if you'd rather see it render.

gkjohnson:
> I haven't tried the two ion assets you mentioned, since I don't have a key.

Getting a key or test assets is the bare minimum effort I'd expect here, and understanding and testing the code you submit is a basic prerequisite for a PR - on this project or any other. Unit tests that only compare the code against itself aren't a substitute for loading a real dataset and confirming it renders.

If the code isn't being validated or understood before it's submitted then nothing is really being offered. I can ask for unverified generated code myself. I have to ask again that some effort be made to understand why this is being added and to make sure it's actually serving the intent of the fix. Other projects have started banning users for low-effort AI PRs, and I'd really rather not go there, but as I've mentioned the burden of contributions like this isn't worth my time when I have to confirm whether even the most basic expectations have been met. If you're genuinely interested in contributing to the project consistently please make sure you understand what the code is doing and how it's used in context.

To move this forward: please load a json-subtree dataset (the ion assets or the Cesium sample) in an example, confirm it renders, and post a screenshot.

Om-singhaI:
Loaded Cesium's ImplicitTilesetWithJsonSubtree sample in example/three/index.html with the tileset url in the hash.

On master, nothing loads past the root
---
On this branch:

https://github.com/user-attachments/assets/3f2faf08-784d-4b56-86e4-890f0ded0ed9


The root's implicitTiling.subtrees.uri is subtrees/{level}.{x}.{y}.json, so the first subtree comes back as parsed json rather than an ArrayBuffer. Master only hands content to SUBTREELoader when the extension is subtree, so that file is fetched and then nothing claims it, and no tiles are created. With the change the loader takes the parsed object, the subtree's tileAvailability of constant 1 over its two levels expands to the root plus four children, and each pulls content/{level}/{x}/{y}.b3dm. Refine is ADD, so all five draw together, which is the Visible: 5 in the header.

I should have rendered real data before replying, my bad.

PR #1734, opened 2026-09-08
Fix BatchedTilesPlugin misplacing tiles when  has a transform

xawill:
### Bug

Since 0.5.0, `TilesRenderer.setTileActive()` parents an active tile scene to `TilesRenderer.group`, and traversal calls it before `setTileVisible()`. So `mesh.matrixWorld` in `addSceneToBatchedMesh()` now includes `group.matrixWorld`, where it used to be group-local. It goes straight into `BatchedMesh.setMatrixAt()`, and the batched mesh is itself a child of that group — the group transform is applied twice.

Repro: register `BatchedTilesPlugin` on a tileset whose group has a non-identity transform, e.g. the Google Photorealistic example's `tiles.group.rotation.x = -Math.PI / 2`. Tiles render on the wrong side of the globe.

`ImageOverlayPlugin` already guards against this with `scene.parent` / `group.matrixWorldInverse`; `BatchedTilesPlugin` was missed.

### Fix

Convert the mesh matrix to group space before `setMatrixAt()`, with the same guard so an already-detached scene is left alone.


gkjohnson:
Thanks for the catch!

PR #1743, opened 2026-09-10
Add PolygonClippingPlugin

hh-hang:
This PR adds polygon clipping support for 3D Tiles.

The initial use case was to clip specific regions of a 3D Tiles dataset so that BIM models could be integrated without overlapping the existing geometry.

https://github.com/user-attachments/assets/cbeaa2fa-25c3-4f8c-8775-802b241e4809

The implementation supports arbitrary polygon shapes, including concave polygons and holes.

b80707145b399929968b62501b9398d7


gkjohnson:
Hello! Thanks for the contribution. Is there a reason that the ImageOverlayPlugin doesn't work for your use case? It has the ability to render geojson or other custom shapes as "masks" to clip out tile geometry from tile sets. The demo had dropped a necessary material flag which I fixed in 30b0236d8afaa1247df3b47c66ef24b59703c57a but you can see it [here](https://nasa-ammos.github.io/3DTilesRendererJS/three/geojson.html).

image

hh-hang:
Thanks for pointing this out. I hadn't noticed the ImageOverlayPlugin before. It looks great and should meet my needs very well.

PR #1745, opened 2026-09-14
Add Babylon.js tile fade plugin and example

CedricGuillemet:
Adds Babylon.js tile fading with shared Three.js timing, material/lifecycle integration, public exports, and a demo using Babylon.js 9.26.1.

image


gkjohnson:
Hi Cedric - thanks for the contribution! Am I right to assume you are part of the Babylon team?

Regarding the PR - it looks like there are some unnecessary changes here outside the scope of the plugin, at least around dependencies. Do you mind taking a look through the changes again and making sure they're all intentional?

Also, I'm thinking it would make more sense to refactor the three.js plugin to a "TilesFadePluginBase" class that exposes something like a "setFadeState" hook that each implementation can then implement. This should help reduce a lot of code duplication.

CedricGuillemet:
Hi Garrett, yes, I'm part of Babylon team.

Sorry for the dependencies left overs. removed now.

Makes sense. Let me know what you think of this refactor.

package.json
```diff
@@ -89,8 +93,8 @@
     "@babel/preset-modules": "^0.1.6",
     "@babel/preset-react": "^7.26.3",
     "@babel/preset-typescript": "^7.26.0",
-    "@babylonjs/core": "^8.47.2",
-    "@babylonjs/loaders": "^8.47.2",
+    "@babylonjs/core": "9.26.1",
+    "@babylonjs/loaders": "9.26.1",
```
gkjohnson, line 97: Can we use a `^` version specifier here.

package.json
```diff
@@ -119,8 +123,8 @@
     "vitest": "^4.0.15"
   },
   "peerDependencies": {
-    "@babylonjs/core": ">=8.0.0",
-    "@babylonjs/loaders": ">=8.0.0",
+    "@babylonjs/core": ">=9.26.1",
+    "@babylonjs/loaders": ">=9.26.1",
```
gkjohnson, line 127: Is it necessary to bump these versions? Or is it possible to have the plugin fail gracefully while keeping the 8.x versions? Eg log a warning if the feature isn't available. I generally try to avoid breaking version bumps frequently especially when the dependency requirement is so new.

src/babylonjs/plugins/fade/FadeMaterialManager.js
```diff
@@ -0,0 +1,220 @@
...
+
+		}
+
+		if ( material.alpha < 1 || material.needAlphaBlendingForMesh( mesh ) ) {
+
+			throw new Error( `TilesFadePlugin: Mesh "${ mesh.name }" uses alpha blending, but tile fading currently supports opaque materials only.` );
+
+		}
+
+	}
```
gkjohnson, line 53: Would it be better to just log a warning or something here? Having a hard failure for basic rendering features that can show up in loaded meshes feels excessive. If dither blending can't be supported with something like transparent meshes, for example, it seems like it would be better to just display them as-is and note it in the docs.

src/babylonjs/plugins/fade/FadeMaterialManager.js
```diff
@@ -0,0 +1,220 @@
...
+
+			}
+
+			validateMesh( mesh, materials );
+			const owner = MESH_OWNERS.get( mesh );
+			if ( owner && owner !== this ) {
+
+				throw new Error( `TilesFadePlugin: Mesh "${ mesh.name }" is already controlled by another fade manager.` );
+
+			}
```
gkjohnson, line 95: When can this situation happen? I'm just trying to wrap my head around some of the babylon life cycle differences but a single mesh should only be only be owned by a single tiles renderer and therefore one fade plugin.

src/babylonjs/plugins/fade/TilesFadePlugin.js
```diff
@@ -0,0 +1,108 @@
...
+	previous.decompose( _scale, _fromRotation, _fromPosition );
+	current.decompose( _scale, _toRotation, _toPosition );
+
+	const dot = Math.min( 1, Math.abs( Quaternion.Dot( _fromRotation, _toRotation ) ) );
+	const angle = 2 * Math.acos( dot );
+	return angle > 0.25 + 1e-6 || Vector3.Distance( _fromPosition, _toPosition ) > 0.1 + 1e-6;
+
+}
+
+export class TilesFadePlugin extends TilesFadePluginBase {
```
gkjohnson, line 41: Can we add JSDoc comments for this? I generate the API.md from the JSDoc comments - I just haven't added that script publicly, yet.

gkjohnson (commented):
Added some comments on some issues that stuck out to me. I will review any impacts to the three.js plugin along with any changes to the base classes made in the "core/plugins" migration next week. Thanks for adding this!

src/three/plugins/fade/FadeManager.js
```diff
@@ -1,232 +1 @@
...
-				this.completeFade( object );
-
-			}
-
-		} );
-
-	}
-
-}
+export { FadeManager } from '../../../core/plugins/fade/FadeManager.js';
```
gkjohnson, line 1: It doesn't look like this file is used anymore.

CedricGuillemet: fixed

src/babylonjs/plugins/fade/TilesFadePlugin.js
CedricGuillemet: yes, added version in the comment to get the fade with Babylon.

src/babylonjs/plugins/fade/FadeMaterialManager.js
CedricGuillemet: Totally! Message is logged if Babylon feature is not found.

package.json
CedricGuillemet: fixed.

package.json
CedricGuillemet: fxed

src/babylonjs/plugins/fade/FadeMaterialManager.js
CedricGuillemet: Too defensive programming. I've removed it. Sorry for the noise.

Issue #1722, opened 2026-09-04
Cache saturation stalls refinement with no observable signal: stats.queued stays 0 and loadProgress stays 1


seb3D974:
### Context

Follow-up to #1689, which was answered with "working as designed". We are not asking for
`maxBytesSize` to become a soft limit — we agree it should be hard. We are asking for **one
observable signal**, so that an application can tell apart:

* "the cache is full and the traversal is happy with what it has", from
* "the cache is full and the traversal wants tiles it will never be allowed to request".

Today those two states are byte-for-byte identical from the outside.

Versions: `3d-tiles-renderer@0.5.2`, `three@0.185.1`, Chromium 1[2]x, desktop GPU.

### What we measured

Real photogrammetry tileset (4 007 tiles, 3D Tiles 1.1, served over a CDN), `lruCache.maxBytesSize`
deliberately reduced to 1 MiB to reproduce a low-memory device. Camera stopped, sample taken 12 s
and again 24 s after the last camera movement — the state never changes:

| field | value |
|---|---|
| `lruCache.isFull()` | **true** |
| `lruCache.cachedBytes` / `maxBytesSize` | 10 549 754 / 1 048 576 |
| `stats.downloading` / `stats.parsing` / `stats.queued` | 0 / 0 / 0 |
| `queuedTiles.length` | 0 |
| `loadProgress` | **1** |
| tiles total / `used` / `inFrustum` / `active` / `visible` | 4007 / 12 / 7 / 6 / 2 |
| `used` tiles whose `loadingState !== LOADED` | **0** |
| time spent in that state | 24.2 s, then 25.5 s in a second run (it never leaves) |

The renderer is stuck at a coarse LOD: the tile that is actually rendered at the frontier has a
screen-space error of **38.4 px** against `errorTarget = 16`, and it has children with content that
will never be requested. Nothing is thrown, nothing is logged, and `tiles-load-end` has already
fired.

### Why no existing field can see it

`queueTileForDownload()` returns **before** anything is counted:

```js
queueTileForDownload( tile ) {
	if ( tile.internal.loadingState !== UNLOADED || this.lruCache.isFull() ) {
		return;                      // ← the tile is dropped here, silently
	}
	this.queuedTiles.push( tile );
}
```

`stats.queued` is only incremented in `requestTileContents()`, which is never reached. Since
`loadProgress = 1 - (queued + downloading + parsing) / total`, **`loadProgress` is 1 for the whole
[...]

PR #1755, opened 2026-09-16
AdaptiveErrorTargetPlugin: Add a plugin that adjusts the error target when the cache is full

seb3D974:
Follow up to #1722 and #1746, per your suggestion there: a plugin that iteratively adjusts the
error target when the cache is determined to be full.

## Motivation

`stats.refused` (#1746) makes the stalled state observable, but reacting to it is still left to
the application, and the reaction is always the same: ask the traversal for fewer, coarser tiles
until they fit. When the cache is full and the camera stops moving, nothing is downloaded, nothing
is unloaded, and the scene stays under refined for as long as the view is held — `loadProgress` is
1, the queue is empty, and `tiles-load-end` has already fired. This plugin turns `stats.refused`
into that reaction, bounded in both directions.

## Behavior

On `update-after`, and only for updates that actually ran a traversal (`frameCount` changed, so an
update skipped by `UpdateOnChangePlugin` does not count as a fresh sample):

- `stats.refused > 0 && lruCache.isFull()` held continuously for `holdTime` ⇒ `errorTarget *= factor`,
  capped at `maxErrorTarget`.
- `stats.refused === 0 && ! lruCache.isFull()` held continuously for `holdTime` ⇒ `errorTarget /= factor`,
  never below the value the application assigned.
- Anything in between (full but satisfied, or refused by a cache that has since been unloaded) holds
  the current target instead of stepping either way.
- Each step restarts the window, so the next step needs a full `holdTime` of its own.
- An `errorTarget` assigned from outside the plugin is adopted as the new base value, so the plugin
  never refines past what the application asked for.
- Every change dispatches `needs-update`, otherwise a raised target would not be read until the
  camera moved again. While a hold window is running the plugin also schedules one `needs-update`
  for the end of that window, so an on-demand renderer gets the chance to re-sample.
- `dispose()` restores the base error target.

## Options

| option | default | |
|---|---|---|
| `holdTime` | `2500` | ms the cache must stay full and refusing before each step |
| `factor` | `1.5` | multiplier up, divisor down |
| `maxErrorTarget` | `null` | ceiling; `null` means 4× the error target the plugin started from |
| `onChange` | `null` | called with `{ errorTarget, previousErrorTarget, reason }` |
| `logging` | `false` | `console.warn` on every change |

It is placed in `core/plugins` since it only touches `errorTarget`, `lruCache`, `stats` and events.

## Example

`example/three/index.js` gains a `cacheMegabytes` slider and an `adaptiveErrorTarget` toggle, and
displays the current error target next to the `Refused` counter added in #1746. Lowering the cache
until `Refused` stops reporting 0 reproduces the stall; toggling the plugin on recovers a displayed
scene. `errorTarget` is now assigned from the GUI `onChange` rather than every frame, otherwise the
example would overwrite whatever the plugin sets.

Measured on that example (local `data/tileset.json`, `errorTarget` 6, `holdTime` shortened to 1 s
for the trace, camera still the whole time):

| cache | error target over time | `stats.refused` | visible tiles |
|---|---|---|---|
| 400 MB (default) | 6, never changed | 0 | 91 |
| 12 MB | 6 → 9 → 13.5 → 20.25 → 24, then held | 85 → 73 → 41 → 24 | 1 |
| 12 MB, raised to 400 MB at t=7s | 24 → 16 → 10.67 → 7.11 → 6, then held | 0 | 1 → 91 |

The first row is the control: a cache that is never full produces no change at all.

## Limits

- Raising the error target does not load more tiles, it asks for fewer and coarser ones. The scene
  is displayed below the requested quality, which is why every change is reported.
- The ceiling is a real stop: in the 12 MB row above, `stats.refused` falls from 85 to 24 but never
  reaches 0, because even the coarse set does not fit. The plugin bounds the degradation rather than
  chasing an error target that would empty the scene.
- Stepping back down needs the cache to actually free up. At rest with a full cache nothing is
  evicted, so the target is held until the camera moves or the budget is raised — the third row
  above is the budget being raised.
- The name is the part I am least attached to; `CacheFullErrorTargetPlugin` or
  `ErrorTargetOnCacheFullPlugin` would read fine too, happy to rename.

`npm run lint`, `npm test` (640 tests, 11 new) and `npm run build-lib` are green. `API.md` is left
alone since it is regenerated separately; the JSDoc is written for it.


gkjohnson:
Can you please simplify the write up? It would be good to read through the description first and ask yourself if you'd like to read through and understand it before submitting. These AI PR descriptions have become impossible to quickly parse and make dealing with issues and other contributions a slog.

I'm more specifically interested in how you came to this solution, whether it's something you're applying with success on your existing project (as in how the approach has been tested), and how.

